PR #295, opened 2021-12-23
fix adsp lint warning and accessiblity

athena-chen-chen:


libs/react-components/src/experimental/icons/icon.component.tsx
```diff
@@ -448,14 +448,14 @@ export type IconStyle = 'outline' | 'filled' | 'sharp';
 interface Props {
   type: GoAIconType,
   size?: IconSize;
-  style?: IconStyle;
+  format?: IconStyle;
```
chrisolsen, line 451: Do you guys use different icon styles? Based on the design guidelines you shouldn't be and therefore setting the `style` should not be needed. The name does clash with the native style attribute, but only if you set it. In the next ui component version that will be out soon, this attribute has been renamed to `variant`

libs/react-components/src/experimental/textarea/textarea.component.tsx
```diff
@@ -20,6 +20,7 @@ export const GoATextArea: FC<Props> = ({ id, name, value, placeholder, onChange
       placeholder={placeholder}
       onChange={(e) => onChange(name, e.target.value)}
       value={value}
+      {...other}
```
chrisolsen, line 23: Is this causing lint issues? I am not sure if opening up the interface like this is a good thing, but I don't think having it open breaks anything, where as there is the possibility that removing this could cause issues.

libs/react-components/src/lib/dropdown/dropdown.tsx
```diff
@@ -136,7 +136,8 @@ export const GoADropdown: FC<Props> = ({ selectedValues = [], ...props }) => {
               readonly={true}
               onChange={null}
               placeholder={props.placeholder}
-              value={getSelectedLabel()} />
+              value={getSelectedLabel()}
+              aria-label="search" />
```
chrisolsen, line 140: According to the aria docs I am not sure this is a `search`.
```
The search role is used to identify the search functionality; the section of the page used to search the page, site, or collection of sites.
```

libs/react-components/src/experimental/icons/icon.component.tsx
athena-chen-chen: variant is better name, thanks

libs/react-components/src/experimental/textarea/textarea.component.tsx
athena-chen-chen: It is accessibility issue, I try to set element aria-label, which is a good way to provide meaningful accessible names for screen reader users, whilst presenting a shorter label for the button to sighted people.

libs/react-components/src/lib/dropdown/dropdown.tsx
athena-chen-chen: How about I just give name as dropdown?

libs/react-components/src/experimental/icons/icon.component.tsx
athena-chen-chen: Revert to style.

libs/react-components/src/experimental/textarea/textarea.component.tsx
chrisolsen: I think this is a needed attribute, but opening things up like this gives a little too much control over things. The other thing is that a new revamped version of the react components will be available soon, so I am not sure if making minor changes like this are worth the effort at the current time.

PR #388, opened 2022-02-10
Web component tests, storybook for button

kevingauld:


libs/web-components/src/components/card/Card.svelte
```diff
@@ -4,10 +4,14 @@
 <script lang="ts">
   export let elevation: number = 0;
   export let width: number = 320;
+
+  //optional
+  export let testId: string = "";
```
chrisolsen, line 9: camelCase variable names will work when testing the svelte component, but will not work when setting it via the web-component since all html properties must be lowercase.

libs/web-components/src/components/card/Card.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+import GoACard from './Card.svelte'
+
+describe('GoACardComponent', () => {
+
+  it('should render - success with elevation', async () => {
+    const baseElement = render(GoACard, { testId: "card-test", width: 500, elevation:1 });
+    const card = await baseElement.findByTestId('card-test');
+
+    expect(card).toBeTruthy();
+    expect(card.style.getPropertyValue("--width")).toEqual("500px");
```
chrisolsen, line 12: I think we need to decide on what needs to be tested for a web component, since testing things like this makes the tests brittle.

libs/web-components/src/components/button/Button.svelte
```diff
@@ -11,6 +11,9 @@
   export let title: string;
   export let disabled: string;
 
+  //optional
+  export let testId: string = "";
```
chrisolsen, line 15: See other comment regarding camelCase

libs/web-components/src/components/button/Button.spec.ts
```diff
@@ -1,16 +1,86 @@
...
+    const { getByTestId } = render(GoAButton, {
+      props: { testId: "button-test", title: 'button',  disabled:"false"}
+    });
+
+    const button = getByTestId('button-test');
+
+    expect(button).toBeTruthy();
+
+    fireEvent.click(button);
+    //expect(mockOnClick).toHaveBeenCalled();
```
chrisolsen, line 32: Missing the verification that the click was made.

libs/web-components/src/components/app-header/AppHeader.svelte
```diff
@@ -4,11 +4,14 @@
 <script lang="ts">
   export let url: string;
   export let title: string;
+
+  //optional
+  export let testId: string = "";
```
chrisolsen, line 9: See other comments regarding camelCase

libs/web-components/src/components/card/Card.spec.ts
kevingauld: I think this was to test the value was actually getting set since it was in input param

chrisolsen:
Kevin's changes were moved into [new PR](https://github.com/GovAlta/ui-components/pull/399) that includes some additional changes

PR #410, opened 2022-02-22
Beta release

chrisolsen:


tarynlaramie:
review comments:
- need a commit with the text "Breaking Change:" so that a major version will be released to npm
- why is this coming from chrisolsen:beta-release and not the alpha branch?

PR #471, opened 2022-03-14
feat: textarea

chrisolsen:


libs/docs/src/components/common/TextArea.stories.mdx
```diff
@@ -0,0 +1,145 @@
...
+import { GoATextArea } from '@abgov/react-components'
+
+<Meta title="Components/TextArea" />
+
+### Component
+# TextArea
+
+[Figma design documentation](https://www.figma.com/file/3pb2IK8s2QUqWieH79KdN7/Components---DIO-DS?node-id=458%3A16042)
+
+## Basic
```
bluekurinji, line 12: To be consistent with other component story page, please add `<Story name="Basic">`

bluekurinji (changes requested):

libs/web-components/src/components/text-area/TextArea.svelte
```diff
@@ -8,21 +8,24 @@
...
   export let disabled: string;
 
   $: isDisabled = toBoolean(disabled);
 
   function onChange(e: Event) {
-    const target = e.target as HTMLInputElement;
+    const target = e.target as HTMLTextAreaElement;
     const value = target.value;
 
+    if (disabled) return;
```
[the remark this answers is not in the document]
chrisolsen: It is :) It was unexpected, but the event was firing when I was testing it, which is why it was added.

libs/web-components/src/components/text-area/TextArea.svelte
```diff
@@ -31,12 +39,16 @@
 
 <!-- HTML -->
 <textarea
+  id={id || name}
```
bluekurinji, line 42: Can we make it consistent as done in checkbox. id as "id-" + name ? just to be consistent?

bluekurinji (approved):

PR #480, opened 2022-03-18
Add style and class properties to skeleton

cjam:
- Add `style` and `class` properties to the `goa-skeleton-element` to allow for more easy customization within applications.

### DIFF
--- a/libs/angular-components/src/lib/skeleton/skeleton-element.component.html
+++ b/libs/angular-components/src/lib/skeleton/skeleton-element.component.html
@@ -1,1 +1,1 @@
-<div [ngClass]="elementClass"></div>
+<div [ngClass]="['skeleton', type, class]" [style]="style"></div>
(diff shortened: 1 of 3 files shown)


cjam:
@chrisolsen Thanks for looking at this.  I'll check out the use of size and see if it fits.  I think the challenge in this particular case is just that the goal of the skeleton pattern is to match the existing UI as closely as possible to avoid layout shifts upon loading.  I've just been finding it challenging to create a skeleton that matches my UI, however, I didn't see the size attribute so I'll give that a try and see if it works.  Does `size` have no impact on the `text` type?

cjam:
@chrisolsen Ah I see, the size is only available in the web components.  How is the web component support in angular?

chrisolsen:
From what we have seen the web-components work well with Angular, but ideally input from other Angular developers would be great.

As for the *mold*, one of the issues with allowing for injected styles is that many web components are more than a wrapper for a single native element and therefore the styles couldn't be easily added to the desired element(s). And if there was a way to do it, it would limit us in being able to make internal changes without the fear of breaking things for others.

chrisolsen:
Maybe for now the best route is to let us know exactly what you would currently want to modify. There is a good chance that if you need something that is missing, that others will feel the same. We are always open to suggestions.
________________________________

Issue #278, opened 2021-12-09
(Alpha) UI component - Checkbox


chrisolsen:
**As a:** Full stack developer
**I want:** Use the "Checkbox" from the UI library
**so that:** I can spend more time creating features or capabilities that will provide value to my users and stakeholders

**Note:** This story does not include the "indeterminate" checkbox state

### Acceptance criteria
- [x] Create web component
- [x] Create React wrapper
- [x] Passing unit tests
- [x] Linting
- [x] update storybook
- [x] Push to Alpha
- [x] Manual design review

### Presentable outcomes
- [x] I can show the ui-component in storybook and provide the link to service teams so that they can view the component for themselves

### Sprint Ready Checklist 

- Acceptance criteria defined 
- eam has defined steps to satisfy acceptance criteria 
- Acceptance criteria is verifiable / testable 
- External / 3rd Party dependencies identified

### Related issues:
GitHub issue: Blocked by #355 

### Resources:
**Mockups:** [Checkbox design](https://www.figma.com/file/3pb2IK8s2QUqWieH79KdN7/DIO-DS---Components?node-id=183%3A219)
**Testing URL:** *[If applicable a URL to the testing branch]*

### Design review:

- [x] Change description text (under main heading) from p to h3
- [x] Adjust the vertical alignment of the checkbox label to be more vertically aligned. right now it seems to be sitting a bit "low" (happy to chat about this and talk about potential solutions)
- [x] Remove "usage" section
- [x] Change the headings for all component states from h3 to h4 (e.g. "basic", "checked", "disabled" etc.)
<img width="453" alt="Screen Shot 2022-03-22 at 9 49 44 AM" src="https://user-images.githubusercontent.com/51723535/159532597-4b51a0e0-61cb-4049-9612-ec67e99a0d69.png">

- [x] Remove "States" heading and the following paragraph: "Checkboxes can be checked or unchecked. They have default, focus, hover, disabled and error states. It’s that in-between state that we might consider a “Mix” between “Selected” and “Unselected” options. Indeterminate state is not supported as of now."

- [x] Remove "Usage" heading and the following paragraph: "Use checkboxes when the user can select one or more options.
The fieldset width is responsive to 100% of the layout container while the arrangement is set vertically for ease of scanning and comparing. The fieldset title should be clear, concise, and relevant to the group of options.Add ‘Optional’ only to questions which can be left unanswered. Start all checkboxes with a capital letter. Don’t include punctuations after checkbox label."

[...]

PR #491, opened 2022-03-22
#278): Implement the missed design system and examples

bluekurinji:


libs/docs/src/components/common/Checkbox.stories.mdx
```diff
@@ -0,0 +1,222 @@
+import {
```
chrisolsen, line 1: I think you need to rebase. This file already exists

chrisolsen (approved):

libs/docs/src/components/common/Checkbox.stories.mdx
```diff
@@ -147,6 +147,25 @@ export const disabledCheckboxTemplate = () => <GoACheckbox name="desserts" text=
   </Tab>
 </Tabs>
 
+
+### Error
+
+export const errorCheckboxTemplate = () => <GoACheckbox name="desserts" text="Ice Cream" error></GoACheckbox>
```
chrisolsen, line 153: This checkbox doesn't change when clicked.

chrisolsen (changes requested):
See error checkbox comment, other than that it looks good.

chrisolsen (approved):

Issue #676, opened 2022-05-05
UI component - button update


Spark450:
**As a:** Full stack developer
**I want:** To see the changes in the design (Figma) of the button component reflected in the ui-component (code)
**so that:** The buttons used in my service are in alignment with the design system standard

### Acceptance criteria
- update tertiary button design
- font weight changed across the whole button set
- Letter spacing
- Compact button name
- type size has increased to maintain accessibility with regular font weight
- disabled states have been updated, now an opacity of 0.5 rather than the grey background used previously, this is in line with the update from CPE
- simplified the variants of the component to follow closer with use
    - sizes of primary, secondary, tertiary
        - Default, compact
    - new type of button — start button: Use a start button for the main call to action on your service’s start page.

- [ ] put on sandbox page
- [ ] documentation updated to reference "start" button instead of "large" button
- [ ] Passing unit tests
- [ ] Linting
- [ ] update storybook
- [ ] Link pull requests for all new tasks
- [ ] Push to Alpha
- [ ] Manual design review

### Presentable outcomes
- [ ] I can show the ui-component in Storybook and provide the link to service teams so that they can view the component for themselves

### Sprint Ready Checklist 

- Acceptance criteria defined 
- Team has defined steps to satisfy the acceptance criteria 
- Acceptance criteria is verifiable/testable 
- External / 3rd Party dependencies identified

### Resources:
**Mockups:** [Button](https://www.figma.com/file/3pb2IK8s2QUqWieH79KdN7/%E2%9D%96-Components-%7C-DIO-DS?node-id=622%3A13874)
**Testing URL:** *[If applicable a URL to the testing branch]*

### Design review

- [x] Arial (and not Acumin Pro Semi condensed) the font that is rendering for the Primary, Secondary, and Tertiary buttons
- [x] Letter spacing needs to be applied Figma does this differently but I think .5px or .03125rem would do it.
- [x] The button "type" for the get started would better match the design if it was "Start"
- [x] The "Start" button has the same size as the rest of the buttons "Normal" should this not be "large" or equivalent?
- [x] Text for Primary button should be "--goa-color-text-light"
- [x] Text for Start button should be "--goa-color-text-light"
- [ ] The button envelope for the tertiary button is smaller the the primary and secondary buttons, it should be the same size
- [x] --gap-size (space between buttons in the button group) should be 1.25rem
- [ ] Do we have leading and trailing icon properties included? If so can we include examples of these variants on the page?
[...]

PR #692, opened 2022-05-10
feat(#676): update Button for new design specs

chrisolsen:
BREAKING CHANGE: The previous `small`, `medium` and `large` sizes have been replaced with `compact` and `normal`.

### DIFF
--- a/libs/styles/src/lib/vars.css
+++ b/libs/styles/src/lib/vars.css
@@ -44,6 +44,8 @@
   --input-font-size: var(--fs-base);
 
   --button-height: 2.625rem;  /* 42px */
+  --button-height-compact: 2rem;  /* 32px */
+  --button-height-tall: 3.25rem;  /* 52px */
 
   /* colors */
   --color-white: #fff;
--- a/libs/web-components/src/components/button/Button.spec.ts
+++ b/libs/web-components/src/components/button/Button.spec.ts
@@ -27,68 +27,27 @@ describe('GoAButtonComponent', () => {
     });
   })
 
-  describe("type", () => {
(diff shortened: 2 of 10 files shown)


libs/web-components/src/components/button/Button.spec.ts
```diff
@@ -66,32 +66,6 @@ describe('GoAButtonComponent', () => {
     });
   });
 
-  describe("size", () => {
-    it('should render small size', async () => {
```
bluekurinji, line 70: Tests are removed for changed properties. Do you have test for new properties?

libs/styles/src/lib/vars.css
```diff
@@ -44,6 +44,8 @@
   --input-font-size: var(--fs-base);
 
   --button-height: 2.625rem;  /* 42px */
+  --button-height-compact: 2rem;  /* 32px */
+  --button-height-tall: 2rem;  /* 52px */
```
bluekurinji, line 48: is the value 2rem incorrect or the comment /* 52px */ incorrect?

libs/web-components/src/components/button/Button.spec.ts
chrisolsen: I will re-add them, but the reason they were removed was that there is no restriction on the size value, so it can by `size="foobar"` and if you test for `foobar` it will pass the test. However the tests do validate that the attribute is used, which is why it should be re-added.

libs/styles/src/lib/vars.css
chrisolsen: Good catch. It should be 3.25rem.

ravi-dio (approved):

PR #697, opened 2022-05-12
fix optional chaining op

chrisolsen:


libs/web-components/src/components/microsite-header/MicrositeHeader.svelte
```diff
@@ -41,7 +41,7 @@
   {/if}
 
   {#if level !== "live"}
-    <div data-testid="level" class="service-level service-level--{level?.toLowerCase()}">
+    <div data-testid="level" class="service-level service-level--{level.toLowerCase()}">
```
bluekurinji, line 44: The change is fine. You could also use css for this.   text-transform: capitalize;

bluekurinji (approved):

libs/web-components/src/components/microsite-header/MicrositeHeader.svelte
chrisolsen: For html text content you can, but not for css class names.

ravi-dio (approved):

ravi-dio (approved):

PR #701, opened 2022-05-17
fix: allow for better control of input, dropdown & textarea widths

chrisolsen:


ravi-dio (approved):

libs/react-components/src/lib/input/input.tsx
```diff
@@ -16,6 +16,7 @@ interface WCProps {
   readonly?: boolean;
   focused?: boolean;
   handletrailingiconclick: boolean;
+  width: string;
```
ravi-dio, line 19: @chrisolsen  Does this need to be made optional to prevent the build error?
TS2741: Property 'width' is missing in type

ravi-dio (approved):

PR #713, opened 2022-05-20
chip style updates

chrisolsen:


libs/react-components/src/lib/chip/chip.tsx
```diff
@@ -20,29 +20,30 @@ declare global {
 
 
 export interface Props {
-  onDeleteIconClick?: () => void;
+  onClick?: () => void;
```
bluekurinji, line 23: Is it by design spec to click on the chip to delete it? It feels little odd for me. If that is th design, I am fine with it. If not, we need to chat. The naming of the event makes sense, if the design spec wants the delete to happen on click. if not we need to relook at it

libs/web-components/src/components/chip/Chip.spec.ts
```diff
@@ -5,26 +5,32 @@ import GoAChip from './Chip.svelte'
...
 
   it("should render", async () => {
-    const { container } = render(GoAChip, { content: "Some Badge"});
+    const { container } = render(GoAChip, { content: "Some Badge" });
     expect(container.innerHTML).toContain("Some Badge");
   })
 
   it("should show the leading icon", async () => {
-    const { container } = render(GoAChip, { content: "Some Badge", leadingicon: "arrow-right"});
+    const { container } = render(GoAChip, { content: "Some Badge", leadingicon: "arrow-right" });
```
bluekurinji, line 13: It would be helpful to assert about the icon name also. It would be even more solid

libs/web-components/src/components/chip/Chip.spec.ts
```diff
@@ -5,26 +5,32 @@ import GoAChip from './Chip.svelte'
 describe('GoAChip', () => {
 
   it("should render", async () => {
-    const { container } = render(GoAChip, { content: "Some Badge"});
+    const { container } = render(GoAChip, { content: "Some Badge" });
```
bluekurinji, line 8: It would be nice to verify the default values of leading icon and error classes not being present. It complements the other tests that assert their presence

bluekurinji (approved):

libs/react-components/src/lib/chip/chip.tsx
chrisolsen: I agree it is odd, but it is the design I was given :)

libs/web-components/src/components/chip/Chip.spec.ts
chrisolsen: I agree this is something that could be looked into. Since the icon are used in many places a new task would have to be created to address this.

bluekurinji (approved):

PR #728, opened 2022-06-02
sam's cypress additions

chrisolsen:


bluekurinji (approved):

chrisolsen:
Sam was able to get things working on his machine with the version of node we are using, so there is no longer a need to keep the cypress tests separate

PR #729, opened 2022-06-02
fix: missed design specifications

bluekurinji:


libs/web-components/src/components/app-footer/AppFooter.svelte
```diff
@@ -317,7 +322,7 @@
   .goa-copyright {
     margin-top: 0;
     margin-bottom: 0;
-    font-size: var(--fs-base);
+    font-size: var(--goa-color-text-secondary);
```
chrisolsen, line 325: you are setting the font size to a color value.

chrisolsen (approved):

PR #731, opened 2022-06-03
fix: show the user set title for the goa-icon and goa-icon-button

chrisolsen:
There are a couple filed bugs on the ionic icons github for this issue.
To make things work on our side the pointer events are disabled on the
`ion-icon` component to allow our title to override it.

https://goa-dio.atlassian.net/browse/DDIDS-594

### DIFF
--- a/libs/web-components/src/components/icon/Icon.svelte
+++ b/libs/web-components/src/components/icon/Icon.svelte
@@ -426,13 +426,17 @@
 <script lang="ts">
   import { toBoolean } from "../../common/utils";
 
+  // required
   export let type: GoAIconType;
+
+  // optional
   export let size: IconSize = "medium";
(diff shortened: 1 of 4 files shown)


libs/web-components/src/components/icon/Icon.svelte
```diff
@@ -455,8 +460,7 @@
   `}
 >
   {#if type}
-    <ion-icon style="width: 100%; height: 100%;" name={theme === "filled" ? type : `${type}-${theme}`}
-    />
+    <ion-icon style="pointer-events: none; width: 100%; height: 100%;" name={theme === "filled" ? type : `${type}-${theme}`} />
```
bluekurinji, line 463: Lets move the inline style into a CSS class

bluekurinji (approved):

PR #737, opened 2022-06-09
Updated Cypress configuration to work with NX mono repo

SamKothari1:


apps/angular-demo-e2e/tsconfig.json
```diff
@@ -1,19 +1,15 @@
...
+    "forceConsistentCasingInFileNames": true
   },
-  "include": ["src/**/*.ts", "src/**/*.js"],
-  "angularCompilerOptions": {
-    "strictInjectionParameters": true,
-    "strictInputAccessModifiers": true,
-    "strictTemplates": true
-  }
+  "include": ["**/*.ts"]
 }
```
chrisolsen, line 15: Instead of removing the `extends` and overwriting everything can you just add in what changes to this file are required.

For example 
- `target`, `module`, `lib` are all defined in the tsconfig.base.json file. 
- Is there a reason that the Angular compile options are being removed?
- Changing the module to `commonjs`

chrisolsen (approved):

PR #741, opened 2022-06-20
feat(DDIDS-312): trap the focus inside the modal when tab or shift+ta…

bluekurinji:
feat(DDIDS-312): trap the focus inside the modal when tab or shift+tab used

Build Trap Focus as separate component

### DIFF
--- a/libs/web-components/src/components/focus-trap/FocusTrap.svelte
+++ b/libs/web-components/src/components/focus-trap/FocusTrap.svelte
@@ -1,0 +1,190 @@
+<svelte:options tag="goa-focus-trap" />
+
+<!-- ======================================================================= -->
+<script lang="ts">
+  import { toBoolean } from "../../common/utils";
+
+  let ignoreFocusChanges: boolean = false;
+  let lastFocus: Element;
+  let element: Element;
+
+  export let open: string;
+  $: isOpen = toBoolean(open);
+
+  $: {
+    if (isOpen) {
+      addListeners();
+    }
+    else {
+      removeListeners();
+    }
+  }
+
+  function removeListeners() {
+    document.removeEventListener('focus', trapFocus, true);
+  }
+
+  function addListeners() {
+    document.addEventListener('focus', trapFocus, true);
+  }
+
+  function attemptFocusOnFirstDescendant(descendants: NodeList | Node[]) {
+
+    if (!descendants) return false;
+
+    for (let i = 0; i < descendants.length; i++) {
+
+      let descendant = descendants[i];
+
+      if (attemptFocus(descendant) ||
+          attemptFocusOnFirstDescendant(descendant.childNodes) ||
+          attemptFocusOnFirstDescendantOfSlotElement(descendant) ||
+          attemptFocusOnFirstDescendantOfShadowDOM(descendant)) {
+        return true;
+      }
+
+    }
+
+    return false;
+  };
+
+  function attemptFocusOnFirstDescendantOfSlotElement(node: Node) {
+
+    if (node instanceof HTMLSlotElement) {
+
+      let assingedNodesOfSlotElement = (node as HTMLSlotElement)?.assignedNodes();
+
+      if (attemptFocusOnFirstDescendant(assingedNodesOfSlotElement)) {
+        return true;
+      }
+    }
+  }
+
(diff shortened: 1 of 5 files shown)


libs/web-components/src/components/modal/Modal.svelte
```diff
@@ -48,7 +49,7 @@
 <!-- ======================================================================= -->
 <!-- Html -->
 <!-- ======================================================================= -->
-
+<goa-focus-trap {open}>
```
chrisolsen, line 52: I think this should be moved inside the `if` block. It also shouldn't be bound to the `open` param since that is as string variable type.

libs/web-components/src/components/focus-trap/FocusTrap.svelte
```diff
@@ -0,0 +1,151 @@
...
+
+  function focusFirstDescendant(nodes: NodeList | Node[]) {
+
+    if (!nodes) return false;
+
+    for (var i = 0; i < nodes.length; i++) {
+
+      var child = nodes[i];
+
+      if (attemptFocus(child) || focusFirstDescendant(child.childNodes)) {
```
chrisolsen, line 39: For readability I think these conditions should be moved into a variable before being used in the `if` statement.
```
const isFocusable = attemptFocus(child) || focusFirstDescendant(child.childNodes)
if (isFocusable) {
   // ...
}
```

This is true for most of the `if` statements in this file.

libs/web-components/src/components/focus-trap/FocusTrap.svelte
```diff
@@ -0,0 +1,151 @@
...
+
+  function trapFocus(event) {
+
+    if (ignoreUtilFocusChanges) {
+      return;
+    }
+
+    const slotElements = (element.firstChild as HTMLSlotElement)?.assignedElements();
+
+    if (!slotElements || !slotElements.length) return;
```
chrisolsen, line 107: This could be changed to 
```
if (!slotElements?.length) return;
```

libs/web-components/src/components/focus-trap/FocusTrap.svelte
```diff
@@ -0,0 +1,151 @@
...
+  };
+
+
+  function attemptFocus(element) {
+
+    if (!isFocusable(element)) {
+      return false;
+    }
+
+    ignoreUtilFocusChanges = true;
```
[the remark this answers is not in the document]
bluekurinji: trapfocus() responsiblity is to process elements focus by user or external code.  When attemptFocus() function calls element.focus(), which will trigger trapfocus()  if ignoreFocusChanges is not used. 

`  function attemptFocus(element) {

    if (!isFocusable(element)) {
      return false;
    }

    ignoreFocusChanges = true;

    try {
      element.focus();
    }
    catch (e) {
    }

    ignoreFocusChanges = false;
    return (document.activeElement === element);
  };

  function trapFocus(event) {

    if (ignoreFocusChanges) {
      return;
    }`

libs/web-components/src/components/focus-trap/FocusTrap.svelte
bluekurinji: done

libs/web-components/src/components/focus-trap/FocusTrap.svelte
```diff
@@ -0,0 +1,151 @@
...
+
+    if (element.disabled) {
+      return false;
+    }
+
+    switch (element.nodeName) {
+      case 'A':
+        return !!element.href && element.rel != 'ignore';
+      case 'INPUT':
+        return element.type != 'hidden' && element.type != 'file';
```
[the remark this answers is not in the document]
bluekurinji: Done

libs/web-components/src/components/modal/Modal.svelte
bluekurinji: done

libs/web-components/src/components/focus-trap/FocusTrap.svelte
bluekurinji: done

chrisolsen (approved):

chrisolsen (approved):

Issue #708, opened 2022-05-19
Clean out old angular libraries, so that we can upgrade node, so that we can get Cypress installed


Spark450:
### Background:
Our team is running into a blocker that is preventing us from getting Cypress installed. 

### Acceptance criteria:
- [ ] Clean out (remove?) old angular libraries
- [ ] Upgrade our version of Node to v16.13.0
- [ ] Work with Sam Kothari to install Cypress

PR #744, opened 2022-06-22
fix(#708): initial values not being visible in dropdown

chrisolsen:


libs/web-components/src/components/dropdown/Dropdown.svelte
```diff
@@ -57,7 +72,8 @@
       switch (data?.type) {
         case BIND: {
           const _data = data as BindMessage
-          const selected = value === _data.value;
+          const selected = _values.includes(_data.value);
```
bluekurinji, line 75: we have data and _data variables. Lets keep the naming consistent. Avoid using this sort of a pair as we dont have any defined meaning to variables starting with _

libs/web-components/src/components/dropdown/Dropdown.svelte
```diff
@@ -57,7 +72,8 @@
...
+
```
bluekurinji, line 76: We need to talk with designers for the need for keeping width as per the max item width. If they say yes, we may need to move this width calculating util somewhere for consistent usage in other components

libs/react-components/src/lib/dropdown/dropdown.tsx
```diff
@@ -6,14 +6,16 @@ export * from './dropdown-option';
...
-  values: string;
+  value: string;
   leadingicon?: string;
-  maxheight?: number;
+  maxheight?: string;
   placeholder?: string;
   filterable?: boolean;
   disabled?: boolean;
   error?: boolean;
   multiselect?: boolean;
```
bluekurinji, line 16: Need to hide multiselect for now

bluekurinji (changes requested):

bluekurinji:
The story book does not have some then exposed properties in the code samples like disabled, leadingicon and placeholder

libs/web-components/src/components/dropdown/Dropdown.spec.ts
```diff
@@ -119,6 +129,8 @@ describe('GoADropdown', () => {
 
     const dropdown = result.queryByTestId("test-id");
     const inputField = dropdown.querySelector('goa-input');
+
+    // show menu
```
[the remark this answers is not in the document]
bluekurinji: Right now, the presence of error is asserted. For better coverage, we also need to assert absence of the error as well. The placement of absence of error assertion could be done in 'should render' test as it is an excellent place for asserting all the default state of an component

libs/web-components/src/components/dropdown/Dropdown.svelte
```diff
@@ -6,24 +6,27 @@
...
 
-  const MAX_HEIGHT = 300;
+  const MAX_HEIGHT = "300px";
 
   // Props
 
   export let name: string;
   export let value: string;
   export let leadingicon: GoAIconType;
-  export let maxheight: number = MAX_HEIGHT;
```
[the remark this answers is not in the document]
chrisolsen: This height is for the option menu and a width is required.

libs/web-components/src/components/dropdown/Dropdown.svelte
```diff
@@ -6,24 +6,27 @@
...
   export let placeholder: string = "";
-  export let disabled: boolean;
-  export let error: boolean;
-  export let testid: string;
-  export let width: string;
+  export let disabled: boolean = false;
+  export let error: boolean = false;
+  export let testid: string = "";
+  export let width: string = "";
+  export let multiselect: boolean = false;
```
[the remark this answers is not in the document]
chrisolsen: It isn't in the docs so it is ok.

libs/web-components/src/components/dropdown/Dropdown.svelte
```diff
@@ -6,24 +6,27 @@
...
 
   // TODO: remove this once goa-input has the toBoolean method removed
   $: isError = error ? "true" : "false";
+  // $: isMultiSelect = toBoolean(multiselect) 
 
   // Private
+  let _values: string[] = [];
```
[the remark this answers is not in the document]
chrisolsen: This is a private var

libs/web-components/src/components/dropdown/Dropdown.svelte
bluekurinji: Looks like, design team to come up with very specific details about how width can be defined. It is out of the scope of this ticket

libs/web-components/src/components/dropdown/Dropdown.svelte
chrisolsen: because `data` is already declared in the current scope.

libs/web-components/src/components/dropdown/Dropdown.svelte
chrisolsen: This is the only component that needs to compute the max item width, if they choose to do so.

libs/react-components/src/lib/dropdown/dropdown.tsx
chrisolsen: It isn't in storybook so I think that is sufficient for now

bluekurinji:
Storybook Example for basic should change into

export class MyComponent {
  colors: string[] = ['red', 'green', 'blue'];
  selectedColor: string[] = ['red'];
  selectColor(event: any) {
    this.selectedColor = event.detail.value;
  }
}

bluekurinji:
As per the react code green should be displayed as selected item. As per angular code red should be the selected item. That needs a correct

bluekurinji:
When same name is applied for more than one dropdown in the same page accidently,  like below 

```typescript 
<h3>Dropdown</h3>
<goa-dropdown
  name="colors"
  placeholder="Select a user"
  [value]="selectedColor"
  (_change)="selectColor($event)"
>
  <goa-dropdown-item
    name="colors"
    *ngFor="let color of colors"
    [value]="color"
    [label]="color"
  >
  </goa-dropdown-item>
</goa-dropdown>

<h3>Error</h3>
<goa-dropdown name="colors" placeholder="Select a user" error="true">
  <goa-dropdown-item value="red" name="colors" label="Red"></goa-dropdown-item>
  <goa-dropdown-item
    value="green"
    name="colors"
    label="Green"
  ></goa-dropdown-item>
  <goa-dropdown-item
    value="blue"
    name="colors"
    label="Blue"
  ></goa-dropdown-item>
</goa-dropdown>
```

The console error is not helpful. 
1. The re should be helpful error suggestion
2. Also, leave a note in storybook for developers

image

bluekurinji:
In storybook, the angular example does not match react example. example placeholder is used in one and another does not

libs/web-components/src/components/dropdown/Dropdown.svelte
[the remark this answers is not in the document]
bluekurinji: Lets pass it for now

PR #747, opened 2022-06-24
DDI DS - 700 Cypress Test support : Button, Checkbox , Radio, Dropdown, Input ,TextArea and Modal code written to support cypress testing

bluekurinji:
…ea code written to support cypress testing

### DIFF
--- a/apps/angular-demo/angular.json
+++ b/apps/angular-demo/angular.json
@@ -1,0 +1,107 @@
+{
+  "$schema": "./node_modules/@angular/cli/lib/config/schema.json",
+  "version": 1,
+  "newProjectRoot": "projects",
+  "projects": {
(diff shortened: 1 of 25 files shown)


package.json
```diff
@@ -54,6 +54,8 @@
   },
   "private": true,
   "dependencies": {
+    "@abgov/styles": "^1.0.0-alpha.12",
+    "@abgov/web-components": "^1.0.0-alpha.70",
```
chrisolsen, line 58: These changes may prevent Nx from being able to link the internal libs properly. So when developing on the React side we might always then be a version behind.

bluekurinji: done

chrisolsen (approved):

PR #748, opened 2022-06-27
fix: dropdown width not being propery set

chrisolsen:


libs/react-components/src/lib/dropdown/dropdown.spec.tsx
```diff
@@ -36,9 +36,9 @@ describe('GoADropdown', () => {
       testId="foo"
       width="200px"
       onChange={() => { }}>
-        <GoADropdownOption label="Red" value="red" />
-        <GoADropdownOption label="Blue" value="blue" />
-        <GoADropdownOption label="Yellow" value="yellow" />
+        <GoADropdownOption name="favColor" label="Red" value="red" />
```
bluekurinji, line 39: As per the intent of this test, other attributes need to be asserted.

libs/react-components/src/lib/dropdown/dropdown.tsx
```diff
@@ -30,8 +30,8 @@ declare global {
 
 interface Props {
   name: string;
-  value: string[]
-  onChange: (name: string, values: string[]) => void;
+  value: string[] | string;
```
bluekurinji, line 33: For readability,  we need to find right name. That takes a single value or araay of values.

libs/web-components/src/components/dropdown/Dropdown.svelte
```diff
@@ -70,12 +71,12 @@
     ctx.subscribe(data => {
       switch (data?.type) {
         case BIND: {
-          const _data = data as BindMessage
+          const _data = data as BindMessage;
```
bluekurinji, line 74: Lets give _ for variables that are props strings to bool.  In this place, _data is actually a bindMessage. Could you replace _data with bindMessage?

libs/web-components/src/components/dropdown/Dropdown.svelte
```diff
@@ -15,14 +16,14 @@
...
+  export let error: string = "false";
+  export let multiselect: string = "false";
 
-  // TODO: remove this once goa-input has the toBoolean method removed
-  $: isError = error ? "true" : "false";
+  $: _disabled = toBoolean(disabled);
+  $: _multiselect = toBoolean(multiselect);
 
   // Private
   let _values: string[] = [];
```
bluekurinji, line 29: How about values instead of _values.   Lets give _ for variables that are props strings to bool.

bluekurinji (changes requested):

PR #767, opened 2022-07-11
fixes to issues found when creating react sample app

chrisolsen:


bluekurinji:
Some issues to be discussed

1. Style is being passed around. Do we want to go down this route?   
<goa-page-block width="704px">
   <router-outlet></router-outlet>
</goa-page-block>

2. button click sensitivity is falling outside of button for the whole container width in angular

3. flexrow and flexcolumn, is duplicate work of css? should we leave it to CS?

4. page components, footer was not sticking to the bottom of the page, when the content is just a text

5. <goa-button-group alignment="end" slot="actions">. This internal deatils could provide backdoor for hacky stuff.

6. Input type="number" is used in sample app. is it supported in release?

bluekurinji (approved):

PR #768, opened 2022-07-12
Component testing

GovAlberta:


apps/angular-demo-e2e/src/fixtures/properties.json
```diff
@@ -78,5 +80,16 @@
...
+  "model-content-margin-bottom": "1rem",
+  "model-content-margin": "0",
+  "input-border-radius-0px" : "0px",
+  "footer-max-width": "60rem",
+  "footer-navigator-column-count": "3",
+  "footer-navigator-column-gap": "1.75rem",
+  "footer-multi-section-padding-top": "1.75rem",
+  "footer-navigator-multi-flex-grow": "1",
+  "footer-navigator-multi-allservice-column-flex-grow": "2",
+  "footer-navigator-multi-social-column-flex-grow": "2"
```
chrisolsen, line 94: FYI: with the changes to the footer the tests using these values will break

chrisolsen:
As was discussed in the git meeting, we will want to squash all these commits into one before merging

apps/angular-demo-e2e/src/fixtures/properties.json
GovAlberta: can this be a placeholder for now until the footer changes are pushed?

PR #779, opened 2022-07-21
chore(DDI-767): React Component test cases for Cypress Test

bluekurinji:


apps/react-demo/src/routes/radio.tsx
```diff
@@ -0,0 +1,49 @@
...
+      </GoARadioGroup >
+
+      <h3>Orientation</h3>
+      <GoARadioGroup name="color" value="orange" orientation="horizontal" onChange={onChange}>
+        <GoARadioItem name="color" value="red"></GoARadioItem>
+        <GoARadioItem name="color" value="blue"></GoARadioItem>
+        <GoARadioItem name="color" value="orange"></GoARadioItem>
+      </GoARadioGroup >
+
+    </>
```
chrisolsen, line 47: These are all using the same `name`, which will cause binding problems.

bluekurinji: fixed

chrisolsen (approved):

PR #787, opened 2022-07-29
fix focus trap

chrisolsen:


chrisolsen:
Initially the focus trap was not working at all on my machine, so this PR was an attempt at a different approach to handling things. After later getting the existing node to work and minor refactor was done instead.

PR #790, opened 2022-08-02
fix(DDI-888): Handle callout heading value being optional

bluekurinji:


ArakTaiRoth (dismissed):
Everything looks good.

ArakTaiRoth (approved):

PR #807, opened 2022-08-17
fix(DDI-829): add aria-labels to form components

chrisolsen:


apps/angular-demo/src/app/dropdown/dropdown.component.html
```diff
@@ -15,7 +15,7 @@ <h3>Dropdown</h3>
 </goa-dropdown>
 
 <h3>Error</h3>
-<goa-dropdown name="colors2" placeholder="Select a user" error="true">
+<goa-dropdown arialabel="Select user" name="colors2" placeholder="Select a user" error="true">
```
ArakTaiRoth, line 18: Is `arialabel` accurate? I know the attribute should be `aria-label` everywhere I've seen. But even if it is accurate, you didn't uppercase the L.

apps/angular-demo/src/app/app.component.html
```diff
@@ -1,83 +1,33 @@
-<goa-microsite-header level="alpha" version="UAT"></goa-microsite-header>
```
ArakTaiRoth, line 1: What is all of this? This doesn't appear to have anything to do with the story.

apps/angular-demo/src/app/form-item/form-item.component.html
```diff
@@ -1,33 +1,34 @@
 <h3>Basic</h3>
-<goa-form-item id="fname" label="First name">
-  <goa-input type="text"></goa-input>
+<goa-form-item label="First name">
```
ArakTaiRoth, line 2: Is all of this actually necessary? We're adding an aria label, to an input, that is already surrounded by a label. Have we tested this with an actual screen reader to hear the result? I'm worried this will end with the screen reader repeating the label twice.

libs/docs/src/components/common/Dropdown.stories.mdx
```diff
@@ -60,6 +60,11 @@ import { useState } from 'react';
     type="boolean"
     description="Sets the dropdown in an error state"
   />
+  <Prop 
+    name="arialabel" 
+    type="string" 
+    description="Defines how the selected value will be translated for the screen reader. If not specified it will fall back to the name"
```
ArakTaiRoth, line 66: Not sure where we're doing it. But I'm commenting on it here, as this is where I first saw mention of it. I don't think we should have a fallback if arialabel isn't specified. aria-label isn't always wanted, nor required.

apps/angular-demo/src/app/form-item/form-item.component.html
chrisolsen: It's not surrounded by a label

apps/angular-demo/src/app/dropdown/dropdown.component.html
chrisolsen: arialabel is our property name, aria-label is the native property name

libs/docs/src/components/common/Dropdown.stories.mdx
chrisolsen: For the form controls it is always wanted. The fallback always ensures some value is present, which in many cases is the value you want.

apps/angular-demo/src/app/app.component.html
chrisolsen: It was done to remove the large number of accessibility errors that existed in the demo app, which made it hard to validate these changes.

apps/angular-demo/src/app/form-item/form-item.component.html
ArakTaiRoth: @chrisolsen What's this then, `<goa-form-item label="First name">`?

libs/docs/src/components/common/Dropdown.stories.mdx
ArakTaiRoth: That isn't true, if an input element has an explicit label associated with it, then aria-label isn't necessary, nor is it wanted. All screen readers will read the label element.

ArakTaiRoth (approved):

PR #815, opened 2022-08-19
chore(DDI-927):change half url to full url

Radha-Alberta:


ArakTaiRoth:
@Radha-Alberta Could we possibly edit the description of this prop as well? To indicate that a full URL is required.

ArakTaiRoth (approved):

PR #818, opened 2022-08-20
chore(DDI-948): warning to important

Radha-Alberta:


ArakTaiRoth:
@Radha-Alberta I believe you also need to modify the `Badge,svelte` file to actually allow `important` as an option that will work. @chrisolsen can correct me if I'm wrong.

ArakTaiRoth (approved):

PR #820, opened 2022-08-20
chore(DDI-949): remove inactive from type

Radha-Alberta:


ArakTaiRoth:
@Radha-Alberta Again, I believe you need to modify `Badge.svelte` to remove `inactive` as an option. This only removes it from the storybook, but not from the actual code.

ArakTaiRoth (approved):

PR #821, opened 2022-08-20
chore(DDI-952): danger to destructive button

Radha-Alberta:


ArakTaiRoth:
@Radha-Alberta I believe you need to modify the `Button.svelte` file to change the options and css styling to `destructive` as well.

ArakTaiRoth (approved):

PR #822, opened 2022-08-20
chore(DDI-957): changes to variant in container

Radha-Alberta:


ArakTaiRoth:
@Radha-Alberta You also need to modify the `Container.svelte` file to allow these changes in code, this currently just modifies the storybook, but not the actual code of the Container.

chrisolsen (approved):

PR #824, opened 2022-08-20
chore(DDI-967): add missing properties and defualt states in radio co…

Radha-Alberta:
…mponent


libs/docs/src/components/common/Radio.stories.mdx
```diff
@@ -36,6 +36,8 @@ import { useState } from 'react';
     type="CustomEvent<{detail: {name: string, value: string}}>"
     description="onChange event"
   />
+  <Prop name="error" type="boolean" defaultValue="false" />
```
ArakTaiRoth, line 39: @Radha-Alberta You're missing a description for both of these props.

ArakTaiRoth (approved):

PR #836, opened 2022-08-26
791 container story updates

chrisolsen:


libs/docs/src/components/common/Container.stories.mdx
```diff
@@ -143,31 +125,23 @@ export const HeadingTemplate = (args) => {
   </Tab>
 </Tabs>
 
-#### Small accent bar
+#### Non-interactive Container
 
-Use the small accent bar container for "secondary" content.
+Non-interactive containers indicate that the content is only for the user to review. Use a grey accent to indicate that there is content in the container that the content is not interactive.
```
ArakTaiRoth, line 130: I know this is nitpicky and I'm sorry. But can you add `and` in between `the container` and `that the content`. Without it, that sentence looks weird.

libs/docs/src/components/common/Container.stories.mdx
```diff
@@ -187,161 +161,156 @@ Use the small accent bar container for "secondary" content.
   </Tab>
 </Tabs>
 
-#### Accent bar with text
+#### Heading bar variants (large, small, with text)
 
-Use the accent bar with text for groupings of content where status or actions need to be applied to the whole container.
+There are three main types within the interactive and non-interactive container types: a large accent bar, a small accent bar, and a accent bar with text.
```
ArakTaiRoth, line 166: Should be `and AN accent bar with text`

libs/docs/src/components/common/Container.stories.mdx
```diff
@@ -187,161 +161,156 @@ Use the small accent bar container for "secondary" content.
...
-      <h2>Heading</h2>
-      Content
-    </GoAContainer>
-  );
-};
+#### Use the accent bar with text for groupings of content where status or actions need to be applied to the whole container.
 
-<Story name="Compact Padding">{CompactPaddingTemplate.bind({})}</Story>
+<Story name="Heading bar with text">
+  <>
```
ArakTaiRoth, line 290: What's this?

libs/docs/src/components/common/Container.stories.mdx
```diff
@@ -352,56 +321,65 @@ export const CompactPaddingTemplate = () => {
...
+            Content
+          </GoAContainer>
+        </GoAFlexRow>
     `}
     />
   </Tab>
 </Tabs>
 
-#### Relaxed Padding
+#### Alternate color for default container
```
ArakTaiRoth, line 362: Couple things, should be `colour`, not `color`.
Also, just a question about this alternate colour option. It looks like it's used with the `colored` option (should also be `coloured`). But that doesn't indicate what colour it's using or anything. Would it not be better to have an option where a user could supply their own alternative colour in a hex format to be used maybe?

libs/react-components/src/lib/container/container.tsx
```diff
@@ -1,14 +1,14 @@
...
+type ContainerVariant = 'interactive' | 'non-interactive' | 'info' | 'error' | 'success' | 'warning';
 type HeadingSize = 'large' | 'small' | 'none';
 type ContainerPadding = "relaxed" | "compact";
 
 interface WCProps {
-  variant: ContainerVariant;
-  headingsize: HeadingSize;
-  colored: boolean;
-  padding: ContainerPadding;
+  variant?: ContainerVariant;
```
ArakTaiRoth, line 8: Another question. I'm not seeing where the defaults are set for these values that aren't required?

ArakTaiRoth (changes requested):

libs/docs/src/components/common/Container.stories.mdx
chrisolsen: We don't give the flexibility to allow them to choose colors, as that would allow them to deviate from the GoA branding.

libs/react-components/src/lib/container/container.tsx
chrisolsen: image

libs/docs/src/components/common/Container.stories.mdx
chrisolsen: The doc header should probably be converted to the English spelling, but all code properties follow the US spelling

libs/docs/src/components/common/Container.stories.mdx
chrisolsen: :+1:  This was all copy and pasted from the UI docs

ArakTaiRoth (approved):

PR #841, opened 2022-08-29
fix(DDI-837): provide 'submit' button type

chrisolsen:


libs/docs/src/components/common/Button.stories.mdx
```diff
@@ -58,6 +58,9 @@ There are three main button types: Primary, Secondary, and Tertiary. If there is
     <GoAButton type="primary" onClick={() => console.log('clicked')}>
       Primary
     </GoAButton>
+    <GoAButton type="submit" onClick={() => console.log('clicked')}>
```
ArakTaiRoth, line 61: This is for above, on line 54. We say there are three main button types, and then list them off. I think we'll want to edit that now that we have a 4th.

ArakTaiRoth (changes requested):

ArakTaiRoth (dismissed):

ArakTaiRoth (approved):

PR #845, opened 2022-08-30
951 badge icon

Radha-Alberta:


chrisolsen (dismissed):
Can you squash these commits?

libs/docs/src/components/common/Badge.stories.mdx
```diff
@@ -166,7 +166,7 @@ which display minimal amounts of information, system feedback, or states.
     <CodeSnippet
       lang="html"
       code={`
-        <goa-badge type="information" content="Information" icon />
+        <goa-badge type="information" content="Information" icon `/>
```
chrisolsen, line 169: Looks like a backtick was accidentally added to this line

chrisolsen:
These changes seem to be doing the opposite of what is specified in the ticket, which is to convert `icon` to `icon="true"`. Now, I believe both are valid on the Angular side, but I would want to verify that just `icon` converts properly to the web component

https://goa-dio.atlassian.net/jira/software/c/projects/DDIDS/boards/185?modal=detail&selectedIssue=DDIDS-951

ArakTaiRoth:
> Can you squash these commits?

@chrisolsen I can squash these when I merge them in, not a big deal.

ArakTaiRoth:
> These changes seem to be doing the opposite of what is specified in the ticket, which is to convert `icon` to `icon="true"`. Now, I believe both are valid on the Angular side, but I would want to verify that just `icon` converts properly to the web component
> 
> https://goa-dio.atlassian.net/jira/software/c/projects/DDIDS/boards/185?modal=detail&selectedIssue=DDIDS-951

I told her to go this way with reducing it to just `icon`. I thought that was a simpler approach, and the whole intent of the story was to make things the same throughout. But that's fair, we should probably test to ensure that this way works in both React and Angular.

ArakTaiRoth (approved):

PR #846, opened 2022-08-30
fix(DDI-924): allow for dynamic data-bindinging to radio and dropdown…

chrisolsen:
… components

### DIFF
--- a/libs/web-components/src/components/dropdown/Dropdown.spec.ts
+++ b/libs/web-components/src/components/dropdown/Dropdown.spec.ts
@@ -33,7 +33,7 @@ describe('GoADropdown', () => {
     });
   });
 
-  it("raise an error if name is not supplied", () => {
+  it("raise an error if name is not supplied", async () => {
     jest.spyOn(console, "error");
     const items = ["red", "blue", "orange"];
     render(GoADropdown, {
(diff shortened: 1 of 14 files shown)


libs/web-components/src/components/dropdown/Dropdown.spec.ts
```diff
@@ -13,7 +13,7 @@ describe('GoADropdown', () => {
     const items = ["red", "blue", "orange"];
     const result = render(GoADropdown, {
       testid: 'test-id',
-      name: 'favcolor',
+      name: 'favcolor2',
```
ArakTaiRoth, line 16: It doesn't look like this test cares about the `name` attribute, so just wondering why did we change this?

libs/web-components/src/components/dropdown/Dropdown.svelte
```diff
@@ -41,15 +41,43 @@
...
+        parseValues();
+        bindContext();
+
+        clearInterval(fn);
+      }
+      if (attempts > maxAttempts) {
+        console.error("goa-dropdown: missing the required `name` attribute. It must match the children's name attribute.")
+        clearInterval(fn);
+      }
+    }, 10);
```
ArakTaiRoth, line 60: What's the time between intervals?

ArakTaiRoth (changes requested):

libs/web-components/src/components/dropdown/Dropdown.svelte
chrisolsen: 10ms

ArakTaiRoth: Of course, I'm such a dunce, sorry.

ArakTaiRoth (approved):

PR #847, opened 2022-08-31
937 default boolean values

Radha-Alberta:


libs/docs/src/components/common/Checkbox.stories.mdx
```diff
@@ -43,11 +43,11 @@ export const checkboxTemplate = (args) => {
 
 <Props>
   <Prop name="name" type="string" required="true" />
-  <Prop name="checked" type="boolean" required="true" />
+  <Prop name="checked" type="boolean" defaultValue="false" required="true" />
```
chrisolsen, line 46: Since this is required, because it is bound within the app, it will never fall back to the default value.

ArakTaiRoth (approved):

PR #851, opened 2022-09-01
DDI-940: add two column layout component

chrisolsen:


apps/angular-demo/src/app/two-column-layout/two-column-layout.component.html
```diff
@@ -0,0 +1,65 @@
+<goa-two-column-layout>
```
ArakTaiRoth, line 1: You've got 3 examples of the two column layout on this page. But I can't see any difference between any of the examples, perhaps it's just super subtle?

libs/docs/src/_stories.mdx.template
```diff
@@ -1,5 +1,5 @@
 import { Meta, Story } from '@storybook/addon-docs';
```
ArakTaiRoth, line 1: What's the point of this file?

apps/angular-demo/src/app/two-column-layout/two-column-layout.component.html
ArakTaiRoth: After looking at the storybook, I can see the differences, but they're just really subtle. Could we maybe make the differences more obvious by using fairly dramatic property values?

libs/docs/src/components/common/TwoColumnLayout.stories.mdx
```diff
@@ -0,0 +1,248 @@
...
+
+<Props showTabs={true}>
+  <Prop name="navcolumnwidth" type="string" description="% | px | rem | ch | etc" defaultValue="256px" />
+  <Prop name="maxcontentwidth" type="string" description="% | px | rem | ch | etc" defaultValue="904px" />
+
+  <Prop name="header" lang="angular" type="slot" required="true" description="How to add GoAAppHeader component" />
+  <Prop name="footer" lang="angular" type="slot" required="true" description="How to add GoAAppFooter component" />
+  <Prop name="nav" lang="angular" type="slot" required="true" description="Component containing nav links for application" />
+
+  <Prop name="header" lang="react" type="ReactNode" required="true" description="GoAAppHeader component" />
```
ArakTaiRoth, line 25: Just wondering. If this is required, and it can only be one thing. Why don't we just create it with that one thing instead of making it a property that the user has to type out?

libs/react-components/src/lib/app-header/app-header.tsx
```diff
@@ -3,6 +3,7 @@ import React, { FC } from 'react';
 interface WCProps {
   heading?: string;
   url?: string;
+  maxcontentwidth?: string;
```
ArakTaiRoth, line 6: I don't think we updated the AppHeader properties in the storybook with this new property.

libs/react-components/src/lib/footer/footer.tsx
```diff
@@ -1,22 +1,27 @@
...
     interface IntrinsicElements {
-      'goa-app-footer': React.HTMLAttributes<HTMLElement>
+      'goa-app-footer': WCProps & React.HTMLAttributes<HTMLElement>
     }
   }
 }
 
 /* eslint-disable-next-line */
 export interface FooterProps {
+  maxContentWidth?: string;
```
ArakTaiRoth, line 18: I don't think we updated the AppFooter properties in the storybook with this new property.

libs/docs/src/_stories.mdx.template
chrisolsen: it's a template file

libs/docs/src/components/common/TwoColumnLayout.stories.mdx
chrisolsen: That is like in OOP having an object that exposes a number of properties of an inner object to allow it to be created internally, instead of just accepting an instance of that object.

apps/angular-demo/src/app/two-column-layout/two-column-layout.component.html
chrisolsen: This is something that Mark would have to make the decision on. Having non-subtle examples also means ones that would not normally, and shouldn't, be used; so I am not sure that is the best thing to do since now the examples shouldn't be used as is.

ArakTaiRoth (approved):

PR #853, opened 2022-09-02
fix(837): Updated the demos to include the new submit type

ArakTaiRoth:
Also updated the package.json to include a command for running the react demo.
And updated the react-components to actually have an option for submit.
I have no idea how it was working in the Storybook prior to this.


libs/react-components/src/lib/button/button.stories.mdx
```diff
@@ -17,7 +17,7 @@ import { DoDont } from '@abgov/shared/storybook-common';
     type: {
```
chrisolsen, line 17: This is an outdated file that will be removed in the current cleanup

apps/angular-demo/src/styles.scss
```diff
@@ -6,4 +6,4 @@
 //@import '../../../dist/libs/styles/styles.esm.css';
 //@import '@abgov/styles/styles.esm.css';
 /****/
-//  @import '../../../dist/libs/styles/styles.esm.css'
+@import '../../../dist/libs/styles/styles.esm.css'
```
chrisolsen, line 9: Adding this line will break the build. This is something that needs to be looked at in the future.

chrisolsen (approved):

PR #862, opened 2022-09-07
fix(DDI-969): Adding more examples of datetime input in stories

ArakTaiRoth:
Also updated the angular demo to include values in the input boxes. And updated the react demo to have seconds in one of the time inputs


chrisolsen:
Looks good. The only thing that need to be tweaked is your commit Jira prefix. It need to have `DDI-` before the number to properly link to the issue.

chrisolsen (approved):

PR #866, opened 2022-09-08
fix(DDI-973): Updated react and angular demos to show prefix and suff…

ArakTaiRoth:
…ix input values

### DIFF
--- a/apps/angular-demo/src/app/input-component/input-component.component.html
+++ b/apps/angular-demo/src/app/input-component/input-component.component.html
@@ -44,7 +44,7 @@
 <h3>Character count with limit</h3>
 <goa-input name="firstname" showcounter="true" maxcharcount="20"></goa-input>
 
-<h3>Prefix and Suffix</h3>
+<h2>Prefix and Suffix</h2>
 <goa-input name="input" prefix="$"></goa-input>
 <goa-input name="input" suffix="items"></goa-input>
 <goa-input name="input" prefix="$" suffix="per item"></goa-input>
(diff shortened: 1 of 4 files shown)


libs/react-components/src/lib/input/input.tsx
```diff
@@ -86,10 +86,10 @@ interface Props {
   step?: number;
 };
 
-export const GoAInput: FC<Props & { type: GoAInputType }> = ({
+export const GoAInput: FC<Props & { type?: GoAInputType }> = ({
   id,
   name,
-  type,
+  type = 'text',
```
chrisolsen, line 92: The default value is already defined in the web component, so this line should be removed.
image

ArakTaiRoth: @chrisolsen I'm not sure that was working. There was a problem until I added that default of 'text' to it.

chrisolsen: That was because previously it was a required value. Now that it is optional line 92 shouldn't be needed anymore.

apps/angular-demo/src/app/input-component/input-component.component.html
```diff
@@ -43,3 +43,8 @@ <h3>Character count</h3>
 
 <h3>Character count with limit</h3>
 <goa-input name="firstname" showcounter="true" maxcharcount="20"></goa-input>
+
+<h3>Prefix and Suffix</h3>
```
chrisolsen, line 47: Can you change this to an `h2` to jive with my large PR?

apps/react-demo/src/routes/input.tsx
```diff
@@ -149,6 +150,11 @@ export default function Input() {
         showCounter={true}
         maxCharCount={20}
       />
+
+      <h3>Prefix and Suffix</h3>
```
chrisolsen, line 154: h2 here too

chrisolsen (approved):

chrisolsen (approved):

PR #877, opened 2022-09-14
feat(DDI-871): add autocapitalize property to GoAInput component

chrisolsen:


ArakTaiRoth:
@chrisolsen Forgot to update the angular demo with the examples. I don't really care about the react demo, as we're already using react in the storybook.

ArakTaiRoth (approved):

PR #879, opened 2022-09-14
fix(DDI-885): rename GoAContainer `variant` property to `type`

chrisolsen:


ArakTaiRoth:
@chrisolsen I think you're missing the react demo as well?

chrisolsen:
@ArakTaiRoth there is no code for this in the react demo

PR #907, opened 2022-10-04
feat(DDI-1002): Adding new property variant

Radha-Alberta:
Added a new property variant in the Chip component


libs/web-components/src/components/chip/Chip.svelte
```diff
@@ -34,6 +35,7 @@
   class="chip"
   class:deletable={_deletable}
   class:error={_error}
+  class:variant={variant}
```
ArakTaiRoth, line 38: Actually, looking over this again, considering this variant doesn't do anything different at the moment, I don't think we need this line.

ArakTaiRoth (approved):

PR #908, opened 2022-10-04
fix(DDI-991): fix date formatting causing date error offsets

chrisolsen:


libs/react-components/src/lib/input/input.tsx
```diff
@@ -176,29 +193,40 @@ export const GoAInputPassword: FC<InputProps> = (props) => {
...
-export const GoAInputDate: FC<Omit<InputProps, "value"> & { value: Date | string}> = ({value, min, max, ...props}) => {
-  const _value: Date = typeof value === "string" ? new Date(value) : value;
-  const _min = min ? format(new Date(min), "yyyy-MM-dd") : "";
-  const _max = max ? format(new Date(max), "yyyy-MM-dd") : "";
-  return <GoAInput {...props} min={_min} max={_max} value={format(_value, "yyyy-MM-dd")} type="date" />;
+export const GoAInputDate: FC<DateInputProps> = ({value, min = "", max = "", ...props}) => {
+  const _format = (value: Date): string => {
+    return format(value, "yyyy-MM-dd");
+  }
+  const _value = _format(typeof value === "string" ? parseISO(value) : value);
```
ArakTaiRoth, line 200: I think this means we can modify `value` in the storybook props to say it can be a date as well. And another question, with this change, could we not also pass a number (ie. a Unix timestamp) for value, min, and max?

chrisolsen: I update the storybook docs with the new value types. As for numeric timestamp values let's just hold off on that until we have requests for that type.

ArakTaiRoth (approved):

PR #909, opened 2022-10-05
fix: Modify padding for AppHeader

ArakTaiRoth:
Modified both the AppHeader and AppFooter stories to include the actual measurements for a property. Increased the max width in the CSS that's used for both AppHeader and AppFooter. Modified width of AppFooter to 100% of the container, instead of 100% of viewport. Removed a console onMount message for Modals.

### DIFF
--- a/libs/docs/src/components/goa/AppFooter.stories.mdx
+++ b/libs/docs/src/components/goa/AppFooter.stories.mdx
@@ -38,7 +38,7 @@ The footer will not position itself on the bottom of the page, but must instead 
   <Prop
     name="maxContentWidth"
     type="string"
-    defaultValue="--layout-max-content-width defined within the styles lib"
+    defaultValue="1280px"
     description="The maximum width of the main content area"
   />
 </Props>
--- a/libs/docs/src/components/goa/AppFooter.stories.mdx
+++ b/libs/docs/src/components/goa/AppFooter.stories.mdx
@@ -38,7 +38,7 @@ The footer will not position itself on the bottom of the page, but must instead 
   <Prop
     name="maxContentWidth"
     type="string"
-    defaultValue="1280px"
+    defaultValue="100%"
     description="The maximum width of the main content area"
   />
 </Props>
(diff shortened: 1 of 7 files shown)


libs/web-components/src/components/footer/Footer.svelte
```diff
@@ -36,7 +36,7 @@
   .content {
     padding: 2rem 1rem;
     margin: 0 auto;
-    width: min(var(--max-content-width), 100vw);
+    width: min(var(--max-content-width), 100%);
```
chrisolsen, line 39: Why is this being changed?

libs/styles/src/lib/vars.css
```diff
@@ -3,7 +3,7 @@
   --font-family: acumin-pro-semi-condensed, sans-serif;
 
   --layout-nav-column-width: 256px;
-  --layout-max-content-width: 904px;
+  --layout-max-content-width: 1280px;
```
chrisolsen, line 6: This default needs to be a small value that results in short line widths for easy reading. After talking to Tom, the value that should be used is 720px

libs/docs/src/components/goa/AppFooter.stories.mdx
```diff
@@ -38,7 +38,7 @@ The footer will not position itself on the bottom of the page, but must instead
   <Prop
     name="maxContentWidth"
     type="string"
-    defaultValue="--layout-max-content-width defined within the styles lib"
+    defaultValue="1280px"
```
chrisolsen, line 41: This should also be 100%

libs/docs/src/components/goa/app-header.stories.mdx
```diff
@@ -36,7 +36,7 @@ import { GoAAppHeader } from '@abgov/react-components';
   <Prop
     name="maxContentWidth"
     type="string"
-    defaultValue="--layout-max-content-width defined within the styles lib"
+    defaultValue="1280px"
```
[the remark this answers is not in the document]
ArakTaiRoth: Done

libs/styles/src/lib/vars.css
ArakTaiRoth: This variable has been removed, and everywhere that used it changed to be 100%

libs/docs/src/components/goa/AppFooter.stories.mdx
ArakTaiRoth: Done

chrisolsen (approved):
When you merge, since the first commit was overwritten by the second, make sure to squash your changes.

PR #911, opened 2022-10-06
Araktairoth/vertical padding

ArakTaiRoth:


libs/web-components/src/components/app-header/AppHeader.svelte
```diff
@@ -81,7 +81,7 @@
   }
   @media (min-width: 640px) {
     .content {
-      padding: 2rem 1.5rem;
+      padding: 1.5rem 1.1rem;
```
chrisolsen, line 84: Does 1.1 (~ 17.6px) matches the design specs? I also think the issue is more related the margin `auto` setting rather than the padding.

ArakTaiRoth: The issue being fixed here is literally the padding, it is too much. It's supposed to be 24px vertical, and 18px horizontal.

chrisolsen: The below image is with your current fix. As I mentioned there is also an issue with the `margin: xx auto`
image

And the vertical padding is 18px, then use `1.125rem`

chrisolsen (approved):

PR #917, opened 2022-10-11
fix(DDI-991): allow date inputs to be cleared

chrisolsen:


apps/react-demo/src/routes/input.tsx
```diff
@@ -1,106 +1,57 @@
 import { GoAInput, GoAInputDate, GoAInputDateTime, GoAInputNumber, GoAInputTime } from '@abgov/react-components';
-import { format } from 'date-fns';
 import * as React from 'react';
 import { useState } from 'react';
 
+import type { GoADate } from "@abgov/react-components";
```
ArakTaiRoth, line 5: Why not just add `GoADate` to the line 1 import, instead of making its own line?

ArakTaiRoth (approved):

ArakTaiRoth (approved):

PR #925, opened 2022-10-17
dropdown menu fix

chrisolsen:


libs/web-components/src/components/dropdown/Dropdown.svelte
```diff
@@ -170,14 +168,50 @@
   }
 
   const onInputKeyDown = (e: KeyboardEvent) => {
+    console.log(e.key, e.keyCode, e.charCode, e.code)
```
ArakTaiRoth, line 171: We should remove this, or at the very least comment it out. We don't want to push a console log to our actual deployed code.

libs/web-components/src/components/dropdown/Dropdown.svelte
```diff
@@ -17,7 +17,7 @@
   export let leadingicon: GoAIconType = null;
   export let maxheight: string = MAX_HEIGHT;
   export let placeholder: string = "";
-  export let testid: string = "";
+  // export let testid: string = "";
```
ArakTaiRoth, line 20: If we aren't using testid anymore, can we just delete this?

ArakTaiRoth (approved):

PR #926, opened 2022-10-17
chore(DDI-985): Added requirement prop

Radha-Alberta:
Renamed existing "optional" prop to "requirement". 
Added code to display optional and required labels.

### DIFF
--- a/libs/react-components/src/lib/form/form-item.tsx
+++ b/libs/react-components/src/lib/form/form-item.tsx
@@ -1,10 +1,10 @@
 import React, { FC } from 'react';
 
-type RequirementText = 'optional' | 'required'
+type RequirementType = 'optional' | 'required'
 
 interface WCProps {
   label: string;
-  requirement?: RequirementText;
+  requirement?: RequirementType;
   error?: string;
   helptext?: string;
 }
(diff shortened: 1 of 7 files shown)


chrisolsen:
I think using `RequirementType` rather than `RequirementText` would be a better choice.

libs/react-components/src/lib/form/form-item.tsx
```diff
@@ -1,8 +1,10 @@
 import React, { FC } from 'react';
 
+type RequirementText = 'optional' | 'required'
+
 interface WCProps {
   label: string;
-  optional?: string;
+  requirement?: RequirementText;
```
chrisolsen, line 7: I think using `RequirementType` makes more sense

chrisolsen, line 7: I think using RequirementType makes more sense

chrisolsen (changes requested):

libs/web-components/src/components/form-item/FormItem.spec.ts
```diff
@@ -24,16 +24,16 @@ describe("GoA FormItem", () => {
   });
 
   it("should not show optional text for a required (default) field", async () => {
```
chrisolsen, line 26: I think this test needs to be updated to make sense with how the `requirement` prop now works. The test name also needs to be changed as there `optional` no longer exists and required is no longer the default value.

We also need an additional test for when when the `requirement="required' works.

chrisolsen (changes requested):

chrisolsen (approved):

PR #928, opened 2022-10-17
fix(DDI-1015): Added a wait command to the start-build

ArakTaiRoth:
It looks like the deployment is happening before the actual build finishes. This should fix that.


chrisolsen (approved):
If this is the solution, It is odd that this was not needed before.

PR #934, opened 2022-10-20
remove warnings

chrisolsen:
DDI-1042


ArakTaiRoth:
@chrisolsen There's so more errors to look into, specifically when running `npm run build:all` or more specifically `nx run react-components:build-lib`. Specifically `The 'this' keyword is equivalent to 'undefined' at the top level of an ES module, and has been rewritten` and `"default" is imported from external module "react" but never used in`.

Also in that first command, or `nx run angular-demo:build-production`, there is `Warning: initial exceeded maximum budget. Budget 500.00 kB was not met by 74.80 kB with a total of 574.80 kB.`

Also more warnings when running `nx run angular-demo-e22:lint`, but I'm not sure if we really care about these ones.

There's errors when linting web-components.

And errors when linting angular-demo and react-demo.

Also, not sure why it's not failing here, but when I run tests on web-components, it's failing. Specifically with `src/components/dropdown/Dropdown.spec.ts`

We could also reduce the output of the unit tests with something like this, https://eloquentcode.com/reduce-verbose-output-and-show-only-failing-tests-in-jest

ArakTaiRoth (approved):

PR #935, opened 2022-10-21
feat(DDI-1023): Added Notification Banner Storybook doc

Radha-Alberta:
Added Storyboook page for Notification Banner documentation


libs/web-components/src/components/icon-button/IconButton.svelte
```diff
@@ -18,12 +18,13 @@
   export let testId: string = "";
   export let disabled: string;
   export let inverted: string;
+  export let noeffects: string;
```
chrisolsen, line 21: What is the prop for? If we need a button without effects it would probably be a variant rather than an extra prop. This changes is also outside the scope of the ticket, so it should be removed.

libs/web-components/src/components/notification/Notification.spec.ts
```diff
@@ -0,0 +1,15 @@
+import '@testing-library/jest-dom';
+import { render, waitFor } from '@testing-library/svelte';
+import GoANotification from './Notification.svelte'
+
+describe('GoANotificationComponent', () => {
+  
+  ["important", "information", "emergency", "event"].forEach(type => {
```
chrisolsen, line 7: We should also be testing to ensure non-valid values show a console error.

libs/web-components/src/components/notification/Notification.svelte
```diff
@@ -4,13 +4,14 @@
 <script lang="ts">
   import { fade } from "svelte/transition";
 
-  export let type: "emergency" | "caution" | "information" | "event";
-
+  export let type: "emergency" | "important" | "information" | "event";
+  export let testid: string = "";
```
chrisolsen, line 8: I want to eliminate the testids being public

chrisolsen (changes requested):

libs/web-components/src/components/notification/Notification.svelte
```diff
@@ -75,7 +94,15 @@
...
   .content {
     flex: 1 1 auto;
+    align-self: flex-start;
+  }
+
+  ::slotted(a) {
+    color: unset !important;
+    outline-color: unset !important;
+    outline-style: unset !important;
   }
```
chrisolsen, line 104: What is this for?

libs/web-components/src/components/icon-button/IconButton.svelte
```diff
@@ -68,30 +67,27 @@
...
     fill: var(--goa-color-interactive);
     cursor: pointer;
     transition: background-color 100ms ease-in, transform 100ms ease-in;
   }
 
-  .nocolor {
-    border-radius: 0.5rem;
-    padding: calc(var(--size) / 4);
+  .darken:not(.inverted){
+    color: unset;
```
chrisolsen, line 83: I am still not so sure about this.

libs/web-components/src/components/notification/Notification.spec.ts
Radha-Alberta: Done

libs/web-components/src/components/notification/Notification.svelte
Radha-Alberta: Removed the testid

libs/web-components/src/components/icon-button/IconButton.svelte
```diff
@@ -23,7 +23,6 @@
   $: css = `${variant} ${isInverted ? "inverted" : ""}`;
   $: isDisabled = toBoolean(disabled);
   $: isInverted = toBoolean(inverted);
-
   $: _size = {
     small: "1rem",
     medium: "1rem",
```
chrisolsen, line 28: Since the calculated size value is always 1rem, let's remove this value and hardcode the 1rem value.

libs/web-components/src/components/icon-button/IconButton.svelte
```diff
@@ -68,30 +67,27 @@
     background: transparent;
     cursor: pointer;
     border: none;
-    padding: 0 0.75rem;
+    border-radius: 0.5rem;
+    padding: calc(var(--size) / 4);
```
chrisolsen, line 71: Since --size will no longer exist we can put the hardcoded value here

chrisolsen (approved):

PR #945, opened 2022-11-03
Araktairoth/update angular setup

ArakTaiRoth:


libs/docs/src/angular-setup.stories.mdx
```diff
@@ -11,6 +11,7 @@ This is the web component library and utilizes Angular's web component integrati
 Add Dependencies
 
 ```bash
+npm i @abgov/web-components@alpha
 npm i @abgov/angular-components@alpha
```
chrisolsen, line 15: Because Nx is still using an older version of Angular, this install line needs to be forced `-f`, at least on my machine it does.

chrisolsen (approved):

PR #955, opened 2022-11-09
feat(DDI-1070): Added storybook doc for Divider component.

Radha-Alberta:
Added a storybook page for Divider component.
Added a new value "none" for "spacing" prop.

### DIFF
--- a/libs/web-components/src/components/divider/Divider.svelte
+++ b/libs/web-components/src/components/divider/Divider.svelte
@@ -3,9 +3,9 @@
 <script lang="ts">
   import { onMount } from "svelte";
 
-  export let spacing: string = "";
+  export let spacing: string = "small";
 
-  const SPACING = ["small", "medium", "large"];
+  const SPACING = ["small", "medium", "large", "none"];
   type Spacing = typeof SPACING[number];
 
   function isSpacing(value: string): value is Spacing {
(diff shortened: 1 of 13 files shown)


libs/web-components/src/components/divider/Divider.svelte
```diff
@@ -3,9 +3,9 @@
 <script lang="ts">
   import { onMount } from "svelte";
 
-  export let spacing: string = "";
+  export let spacing: string = "none";
```
chrisolsen, line 6: I don't think the default should be `none` since that conflicts with what is expected, based on the html `<hr>` element.

chrisolsen: Adding none is also not really required, since a blank value is already seen as being none.

ArakTaiRoth: @chrisolsen Should none even be an option? Why would someone ever add a divider element with no spacing? Is there a use case?

chrisolsen: I talked to Tom about this and he does want it, but we will likely shift to work more inline with the design tokens.

Radha-Alberta: Changed the default spacing to small.

chrisolsen (approved):

PR #956, opened 2022-11-14
1026 component margin

chrisolsen:


libs/docs/src/components/common/Block.stories.mdx
```diff
@@ -0,0 +1,230 @@
+import { Meta, Story } from "@storybook/addon-docs";
```
ArakTaiRoth, line 1: There's nothing in here showing the different gap values. I would suggest adding that.

ArakTaiRoth (approved):

ArakTaiRoth (approved):

PR #961, opened 2022-11-15
fix(DDIDS-1031): styling changes of callout modal

Radha-Alberta:
Adjusted border radius for callout variant
Adjusted padding for modals without callout
Added modal border and box-shadow properties

### DIFF
--- a/libs/web-components/src/components/modal/Modal.svelte
+++ b/libs/web-components/src/components/modal/Modal.svelte
@@ -72,7 +72,7 @@
         out:fly={{ delay: _transitionTime, duration: _transitionTime, y: -100 }}
         class="modal-pane"
       >
-        {#if calloutvariant !== ""}
+        {#if calloutvariant !== null}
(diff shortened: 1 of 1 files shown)


libs/web-components/src/components/modal/Modal.svelte
```diff
@@ -181,9 +182,10 @@
     width: 90%;
     display: flex;
     margin: 1rem;
-    box-shadow: var(--shadow-2);
+    box-shadow: 6px 6px 6px rgba(0, 0, 0, 0.16);
     border-radius: 4px;
     max-height: 90%;
+    border: 1px var(--color-gray-600);
```
chrisolsen, line 188: you are missing the pattern definition
```
border: 1px solid var(--color-gray-600)
```

Radha-Alberta: Changed.

chrisolsen (approved):

PR #962, opened 2022-11-15
fix(DDI-1031): Remove margin-bottom from last element in content area

Radha-Alberta:
Removed the margin-bottom from last element in modal-content

### DIFF
--- a/libs/web-components/src/components/modal/Modal.svelte
+++ b/libs/web-components/src/components/modal/Modal.svelte
@@ -199,6 +199,11 @@
     margin: 1.5rem 0 0;
   }
 
+  .modal-content ::slotted(:last-child) {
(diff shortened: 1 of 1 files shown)


libs/web-components/src/components/modal/Modal.svelte
```diff
@@ -199,6 +199,11 @@
     margin: 1.5rem 0 0;
   }
 
+  .modal-content ::slotted(:last-child) {
+    margin-bottom: unset !important;
```
chrisolsen, line 203: Is there a reason this is not being set to 0?

Radha-Alberta: Changed the margin-bottom to 0

chrisolsen (approved):

PR #979, opened 2022-12-04
fix(DDI-1123): Removed background-color from textarea component

ArakTaiRoth:
This was done so it would use the natural color for the HTML component, including when it was disabled.

### DIFF
--- a/libs/web-components/src/components/text-area/TextArea.svelte
+++ b/libs/web-components/src/components/text-area/TextArea.svelte
@@ -108,17 +108,14 @@
     display: block;
     width: 100%;
     box-sizing: border-box;
-
     outline: none;
     transition: box-shadow 0.1s ease-in;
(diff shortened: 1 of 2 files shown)


libs/web-components/src/components/text-area/TextArea.svelte
```diff
@@ -108,17 +108,14 @@
     display: block;
     width: 100%;
     box-sizing: border-box;
-
     outline: none;
     transition: box-shadow 0.1s ease-in;
     border: 1px solid var(--color-gray-600);
     border-radius: 3px;
-    background: var(--color-white);
```
chrisolsen, line 116: This should stay, it is the background in the disabled state that needs to be set

chrisolsen (approved):

chrisolsen (approved):

PR #985, opened 2022-12-07
feat(DDI-1077): add ability for Dropdown component to render native s…

chrisolsen:
…elect element


libs/docs/src/components/common/Dropdown.stories.mdx
```diff
@@ -76,6 +76,11 @@ import { useState } from "react";
     type="string"
     description="Defines how the selected value will be translated for the screen reader. If not specified it will fall back to the name"
   />
+  <Prop
+    name="native"
+    type="boolean"
+    description="When true will render the native <select> html element"
```
ArakTaiRoth, line 82: Add `defaultValue="false"` to align with the rest of our boolean values in the table.

libs/web-components/src/components/input/Input.svelte
```diff
@@ -246,10 +251,10 @@
     flex: 1 1 auto;
   }
 
-  input[type=text],
-  input[type=date],
+  input[type="text"],
```
ArakTaiRoth, line 254: Just a question, not really a comment. Maybe this is auto formatting, but doesn't this potentially change the meaning?

libs/docs/src/components/common/Dropdown.stories.mdx
```diff
@@ -41,18 +41,18 @@ import { useState } from "react";
...
+    description="Maximum height of the dropdown menu items popover. **Non-native only**"
   />
   <Prop
     name="placeholder"
     type="string"
-    description="The text displayed for the dropdown before a selection is made."
+    description="The text displayed for the dropdown before a selection is made. **Non-native only**"
   />
   <Prop
     name="width"
```
ArakTaiRoth, line 58: I noticed this property is not in the NativeSelect. Should we expand its description as well to say `**Non-native only**`?

libs/web-components/src/components/dropdown/Dropdown.svelte
```diff
@@ -217,82 +227,100 @@
...
+    onSelect(option.value, option.label, false);
+  }
 </script>
 
 <!-- Template -->
-
 <div
   data-testid={`${name}-dropdown`}
-  class="goa-dropdown-box"
+  class="dropdown"
```
ArakTaiRoth, line 241: Could this potentially be considered a breaking change? I'm wondering about people that might have their own javascript, or tests, that maybe target this class.

libs/web-components/src/components/native-select/NativeSelect.svelte
```diff
@@ -0,0 +1,116 @@
+<svelte:options tag="goa-select" />
```
ArakTaiRoth, line 1: Why do we have this component, if we're also just allowing Dropdown to have a `native` property?

libs/web-components/src/components/dropdown/Dropdown.svelte
chrisolsen: No, if that was the case, then any changes we make internally would be considered breaking.

libs/web-components/src/components/native-select/NativeSelect.svelte
chrisolsen: Missed deleting file

libs/docs/src/components/common/Dropdown.stories.mdx
chrisolsen: The `width` attribute is used for both the native and non-native version of the Dropdown

libs/web-components/src/components/input/Input.svelte
chrisolsen: It was autoformatting. Meaning is the same

PR #988, opened 2022-12-13
feat(DDI-1101): Added slots for leading and trailing contents

Radha-Alberta:
Added two new properties leadingContent and trailingContent in the input component.
Added warning message to show when prefix/suffix are used.
Updated tests accordingly.

### DIFF
--- a/libs/docs/src/components/common/Input.stories.mdx
+++ b/libs/docs/src/components/common/Input.stories.mdx
@@ -15,6 +15,7 @@ import {
   GoAInputRange,
   GoAFormItem,
   GoACallout,
+  GoABadge,
 } from "@abgov/react-components";
 import { useState, useEffect } from "react";
 import {
@@ -147,12 +148,20 @@ import { Meta, Story } from "@storybook/addon-docs";
   <Prop
     name="prefix"
     type="string"
-    description="Text that comes before the input"
+    description={
+      <>
+        Text that comes before the input. <strong>Deprecated</strong>
+      </>
+    }
   />
   <Prop
     name="suffix"
     type="string"
-    description="Text that comes after the input"
+    description={
+      <>
+        Text that comes after the input. <strong>Deprecated</strong>
+      </>
+    }
   />
   <Prop
     name="autoCapitalize"
@@ -165,6 +174,34 @@ import { Meta, Story } from "@storybook/addon-docs";
     lang="angular"
     type="CustomEvent"
     description="onclick function invoked when trailing icon is clicked"
+  />
+  <Prop
+    name="leadingContent"
+    type="slot"
+    lang="angular"
+    defaultValue=""
+    description="Sets the content to the left of the input field"
+  />
+  <Prop
+    name="leadingContent"
+    type="ReactNode"
+    lang="react"
+    defaultValue=""
+    description="Sets the content to the left of the input field"
+  />
+  <Prop
+    name="trailingContent"
+    type="slot"
+    lang="angular"
+    defaultValue=""
+    description="Sets the content to the right of the input field"
+  />
+  <Prop
+    name="trailingContent"
+    type="ReactNode"
+    lang="react"
+    defaultValue=""
+    description="Sets the content to the right of the input field"
   />
   <Prop
     name="onChange"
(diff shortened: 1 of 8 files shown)


libs/docs/src/components/common/Input.stories.mdx
```diff
@@ -166,6 +166,34 @@ import { Meta, Story } from "@storybook/addon-docs";
...
+    defaultValue=""
+    description="Sets the content to the right of the input field"
+  />
+  <Prop
+    name="trailingContent"
+    type="ReactNode"
+    lang="react"
+    defaultValue=""
+    description="Sets the content to the right of the input field"
+  />
```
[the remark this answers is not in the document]
ArakTaiRoth: Oh, and remove the examples using prefix and suffix from the Storybook as well

libs/react-components/src/lib/input/input.tsx
```diff
@@ -95,6 +95,8 @@ export interface InputProps extends BaseProps {
   min?: number | string;
   max?: number | string;
   step?: number;
+  leadingContent?: React.ReactNode;
+  trailingContent?: React.ReactNode;
```
chrisolsen, line 99: These should be moved into the BaseProps interface

libs/react-components/src/lib/input/input.spec.tsx
```diff
@@ -3,6 +3,7 @@ import { render } from "@testing-library/react";
 import { fireEvent, screen } from "@testing-library/dom";
 import { GoAInputText, InputProps } from "./input";
 import { GoAIconType } from "../icon/icon";
+import React, { useState } from "react";
```
chrisolsen, line 6: I don't think you are using the `useState`

libs/docs/src/components/common/Input.stories.mdx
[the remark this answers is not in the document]
chrisolsen: Should the be removed or marked as deprecated?

libs/docs/src/components/common/Input.stories.mdx
```diff
@@ -987,6 +1015,55 @@ export const AutocapitalizeTemplate = () => {
   </Tab>
 </Tabs>
 
+#### LeadingContent and TrailingContent
```
chrisolsen, line 1018: I think there should be an additional example that includes adding a one of the web components to the slot so it is clear that it isn't text only.

chrisolsen:
leadingIcon and trailingIcon also need to be included what is being moved to the new slots

chrisolsen:
You will also need to fix your commit message. Your DDI-XXX has a trailing underscore instead of closig paren

libs/react-components/src/lib/input/input.tsx
Radha-Alberta: Done.

libs/react-components/src/lib/input/input.spec.tsx
Radha-Alberta: Yes, removed useState from import

Radha-Alberta:
> You will also need to fix your commit message. Your DDI-XXX has a trailing underscore instead of closig paren

Fixed

libs/docs/src/components/common/Input.stories.mdx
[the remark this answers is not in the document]
Radha-Alberta: Done. Marked them as deprecated.

libs/docs/src/components/common/Input.stories.mdx
```diff
@@ -147,12 +147,12 @@ import { Meta, Story } from "@storybook/addon-docs";
   <Prop
     name="prefix"
     type="string"
-    description="Text that comes before the input"
+    description="Text that comes before the input. **Deprecated**"
```
ArakTaiRoth, line 150: For some reason this doesn't make the `Deprecated` bold in the storybook. This should definitely be made bold, if possible.

libs/docs/src/components/common/Input.stories.mdx
```diff
@@ -987,6 +1019,55 @@ export const AutocapitalizeTemplate = () => {
...
+#### LeadingContent and TrailingContent
+
+<Story name="LeadingContent and TrailingContent">
+  <>
+    <h5>Leading Content</h5>
+    <GoAInput name="input" value="" leadingContent="$" onChange={noop} />
+    <h5>Trailing Content</h5>
+    <GoAInput name="input" value="" trailingContent="items" onChange={noop} />
+    <h5>Leading and Trailing Content</h5>
+    <GoAInput
```
ArakTaiRoth, line 1031: Can we change one or more of these examples to showcase how different things other than just strings can be used in here? One example definitely needs to be a div with its own styled width.

Radha-Alberta: Added examples

ArakTaiRoth (approved):

PR #992, opened 2022-12-15
feat(DDI-1066): add basic table component

chrisolsen:


Radha-Alberta:
Width is set to Required True in table properties
image

ArakTaiRoth (approved):

PR #994, opened 2022-12-21
feat(DDI-1084): Adding typeValidation to components

Radha-Alberta:
Added typeValidation to all properties having multiple values

### DIFF
--- a/libs/docs/src/components/common/Modal.stories.mdx
+++ b/libs/docs/src/components/common/Modal.stories.mdx
@@ -499,7 +499,6 @@ export const RouteChangesTemplate = (args) => {
         actions={
           <GoAButtonGroup alignment="end">
             <GoAButton
-              size="medium"
               onClick={() => {
                 setShowModal(false);
               }}
@@ -583,7 +582,7 @@ export const RouteChangesTemplate = (args) => {
 
 export const CalloutVariantTemplate = (args) => {
   const [showModal, setShowModal] = useState(false);
-  const [calloutVariant, setCalloutVariant] = useState(false);
+  const [calloutVariant, setCalloutVariant] = useState("");
   function show(calloutVariant) {
     setCalloutVariant(calloutVariant);
     setShowModal(true);
--- a/libs/web-components/src/components/form-item/FormItem.svelte
+++ b/libs/web-components/src/components/form-item/FormItem.svelte
@@ -5,6 +5,16 @@
   import { onMount } from "svelte";
   import type { Spacing } from "../../common/styling";
   import { calculateMargin } from "../../common/styling";
+  import { typeValidator, toBoolean } from "../../common/utils";
+
+  // Validators
+  const [REQUIREMENT_TYPES, validateRequirementType] = typeValidator(
+    "Requirement type",
+    ["optional", "required"],
+    false,
+  );
+
+  type RequirementType = (typeof REQUIREMENT_TYPES)[number];
 
   export let testid: string = "";
 
@@ -14,14 +24,6 @@
   export let mb: Spacing = null;
   export let ml: Spacing = null;
 
-  const REQUIREMENT_TYPES = ["optional", "required", ""];
-  type RequirementType = (typeof REQUIREMENT_TYPES)[number];
-
-  // type check function
-  function isRequirementType(value: string): value is RequirementType {
-    return REQUIREMENT_TYPES.includes(value);
-  }
-
   // Optional
   export let label: string = "";
   export let helptext: string = "";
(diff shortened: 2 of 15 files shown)


libs/web-components/src/components/input/Input.svelte
```diff
@@ -5,36 +5,32 @@
...
+  );
+
+  const [AutoCapitalize, validateAutoCapitalize] = typeValidator(
+    "Input auto capitalize",
+    ["on", "off", "none", "sentences", "words", "characters"]
+  );
+
+  // Types
+  type Type = typeof Types[number];
+  type autoCapitalize = typeof AutoCapitalize[number];
```
chrisolsen, line 27: We should capitalize types

libs/docs/src/components/common/Modal.stories.mdx
```diff
@@ -583,7 +582,7 @@ export const RouteChangesTemplate = (args) => {
 
 export const CalloutVariantTemplate = (args) => {
   const [showModal, setShowModal] = useState(false);
-  const [calloutVariant, setCalloutVariant] = useState(false);
+  const [calloutVariant, setCalloutVariant] = useState();
```
ArakTaiRoth, line 585: To be consistent with the code we're telling them below to use, I think this needs to be `useState("");`

libs/web-components/src/components/input/Input.svelte
Radha-Alberta: Done

libs/docs/src/components/common/Modal.stories.mdx
Radha-Alberta: Done

libs/web-components/src/components/form-item/FormItem.svelte
```diff
@@ -5,6 +5,16 @@
   import { onMount } from "svelte";
   import type { Spacing } from "../../common/styling";
   import { calculateMargin } from "../../common/styling";
+  import { typeValidator, toBoolean } from "../../common/utils";
+
+  // Validators
+  const [REQUIREMENT_TYPES, validateRequirementType] = typeValidator(
```
ArakTaiRoth, line 11: I don't think `REQUIREMENT_TYPES` fits here. It's so different than everywhere else we're doing this. I would probably just change `REQUIREMENT_TYPES` to `Requirement`, to be consistent with everything else in the codebase.

libs/web-components/src/components/input/Input.svelte
```diff
@@ -5,36 +5,32 @@
...
-    | "time"
-    | "url"
-    | "week" = "text";
+  // Validators
+  const [Types, validateType] = typeValidator(
+    "Input type",
+    ["text", "number", "password", "email", "date", "datetime-local", "month", "range", "search", "tel", "time", "url", "week"]
+  );
+
+  const [AUTO_CAPITALIZE, validateAutoCapitalize] = typeValidator(
```
ArakTaiRoth, line 20: I would change `AUTO_CAPITALIZE` to just `AutoCapitalize`, again for consistency.

libs/web-components/src/components/notification/Notification.svelte
```diff
@@ -4,16 +4,19 @@
 <script lang="ts">
   import { fade } from "svelte/transition";
   import { onMount } from "svelte";
+  import { typeValidator, toBoolean } from "../../common/utils";
 
-  export let type: NotificationType = "";
+  // Validator
+  const [NOTIFICATION_TYPES, validateType] = typeValidator(
```
ArakTaiRoth, line 10: Also here, change `NOTIFICATION_TYPES` to just `Types` as per everywhere else we have a type property.

libs/web-components/src/components/form-item/FormItem.svelte
Radha-Alberta: Changed to Requirements

libs/web-components/src/components/input/Input.svelte
Radha-Alberta: Changed to AutoCapitalize

libs/web-components/src/components/notification/Notification.svelte
Radha-Alberta: Changed to Types

libs/web-components/src/components/form-item/FormItem.svelte
chrisolsen: `REQUIREMENT_TYPES` is what was used before Radha's changes. `REQUIREMENT_TYPES` is not a type which is why it does not use the pascal case notation; it is just the array that was passed in. Naming it `Requirements` implies it is a type, which then makes one of the lines of code confusing. It could be camelcased, but initially I went with the old-school node all uppercase style. 
```
      {#if requirement && Requirements.includes(requirement)}
        <em>({requirement})</em>
      {/if}
```

ArakTaiRoth (approved):

PR #999, opened 2023-01-04
use design token values within web-components

chrisolsen:


libs/web-components/src/variables.css
```diff
@@ -0,0 +1,11 @@
+:root {
```
ArakTaiRoth, line 1: Just wondering, would it be better if we had these variables coming in from the design docs instead, through the design tokens?

package.json
```diff
@@ -71,6 +71,7 @@
     "zone.js": "~0.11.4"
   },
   "devDependencies": {
+    "@abgov/design-tokens": "^1.0.5-dev.4",
```
ArakTaiRoth, line 74: Just wondering the order of operations we want to take for this. I was thinking once Tom approves this work, then we merge design tokens to production, and change this line to use production?

chrisolsen: I don't think the design tokens are stable to the point were it should be released, so we will have to keep it referring to the dev branch until that happens.

ArakTaiRoth (approved):

ArakTaiRoth (approved):

PR #1011, opened 2023-01-13
feat(DDI-848): add pagination component

chrisolsen:


libs/react-components/src/index.ts
```diff
@@ -1,5 +1,7 @@
 import "@abgov/web-components";
 
+export * from "./lib/pagination/pagination";
```
ArakTaiRoth, line 3: It looks like this line is repeated on line 39, could remove this maybe?

libs/web-components/src/components/input/Input.svelte
```diff
@@ -258,8 +270,16 @@
     padding: var(--input-padding);
     line-height: calc(40px - calc(var(--input-padding) * 2));
     background-color: transparent;
+
     width: 0;
     flex: 1 1 auto;
+  }
+
+  input[type="text"],
```
ArakTaiRoth, line 278: Lines 278 to 282 are going backwards. We moved the font-family to be under `input` instead of separated by type, so that every input type used the same font.

libs/react-components/src/lib/pagination/pagination.spec.tsx
```diff
@@ -0,0 +1,10 @@
+import { render } from "@testing-library/react";
+
+import Pagination from "./pagination";
+
+describe("Pagination", () => {
+  it("should render successfully", () => {
```
ArakTaiRoth, line 6: This seems a little light on tests. Should we also be testing if all props work as expected, at the very least?

libs/web-components/src/components/pagination/Pagination.svelte
```diff
@@ -0,0 +1,68 @@
...
+  $: _pageCount = Math.ceil(itemcount / perpagecount)
+
+  // private
+  let inputEl: HTMLInputElement = null;
+  let prevEl: HTMLInputElement;
+  let nextEl: HTMLInputElement;
+
+  // hooks
+  onMount(() => {
+    validateRequired("GoAPagination", { itemcount })
```
ArakTaiRoth, line 28: It looks like you're missing validating the `variant` property?

libs/web-components/src/index.ts
```diff
@@ -43,3 +43,5 @@ export * from "./layouts/FullScreenNavbarLayout.svelte";
 export * from "./layouts/FullScreenNavbarLayout.svelte";
 export * from "./layouts/one-column-layout/OneColumnLayout.svelte";
 export * from "./layouts/two-column-layout/TwoColumnLayout.svelte";
+
+export { default as Pagination } from './components/pagination/Pagination.svelte';
```
ArakTaiRoth, line 47: Just wondering why the export on Pagination is different from all the other exports?

libs/docs/src/components/common/Pagination.stories.mdx
```diff
@@ -0,0 +1,293 @@
...
+      lang="html"
+      code={`
+    `}
+    />
+  </Tab>
+  <Tab label="Angular">
+    <CodeSnippet
+      lang="ts"
+      code={`
+      interface User {
```
ArakTaiRoth, line 150: Maybe good to show the import of the `faker` library here, as the code below could be misunderstood if they don't realize that's a library being used.

libs/web-components/src/components/pagination/Pagination.svelte
chrisolsen: pagenumber, yes, variant has a default value.

libs/web-components/src/index.ts
chrisolsen: This is just an auto-generated line that uses a different format

ArakTaiRoth (approved):

PR #1015, opened 2023-01-16
feat(DDI-998): Added example for linecount prop

Radha-Alberta:
Added example for linecount prop

### DIFF
--- a/libs/web-components/src/components/skeleton/Skeleton.svelte
+++ b/libs/web-components/src/components/skeleton/Skeleton.svelte
@@ -48,7 +48,7 @@
       <svelte:self type="image" {size} />
       <div class="card-content">
         <svelte:self type="header" {size} />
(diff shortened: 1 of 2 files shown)


libs/web-components/src/components/skeleton/Skeleton.svelte
```diff
@@ -6,6 +6,7 @@
   import { calculateMargin } from "../../common/styling";
   import { onMount } from "svelte";
   import { typeValidator } from "../../common/utils";
+  import { to_number } from "svelte/internal";
```
chrisolsen, line 9: Probably shouldn't be using internal methods. Use `Number.parseInt` instead

Radha-Alberta: Done

ArakTaiRoth (approved):

PR #1017, opened 2023-01-19
feat(DDI-1134): add details component

chrisolsen:


Radha-Alberta:
1. In safari browser, the yellow border doesn't have rounded corners
image
2. The hover and focus states doesn't have grey background as in Figma.
image
image
3. The chevron icon should align with the first line of heading
image

libs/web-components/src/common/styling.ts
```diff
@@ -1,4 +1,28 @@
...
+  "5": "m",
+  "6": "l",
+  "7": "xl",
+  "8": "2xl",
+  "9": "3xl",
+  "10": "4xl",
+}
+
+/**
+ * Allow for 0-10 values to be used along side the existing Spacing values
```
ArakTaiRoth, line 18: Do we want to document anywhere that this is a thing? I'm not sold on that it matters, just wondering if we allow it.

ArakTaiRoth:
Feedback from Tom:

- [ ] icon is not changing to interactive-hover colour when closed and hovering
- [ ] background should be white when heading is open and not hovering
- [ ] text should be interactive-default when heading is open and not hovering
- [ ] Space between icon and heading should be 8px. Currently it is 2px+4px (margin on icon and heading respectively)
- [ ] There should be interior padding on the right side of the content when open (1rem)
- [ ] Total height should be 42px tall overall (aligns with buttons)

Radha-Alberta:
Few other comments after checking with Tom.
1. The line height should be larger. It is currently 24px, and it should be 28px (--goa-line-height-2)
2. The content text should be styled as --goa-typography-body-m
3. The icon is also not vertically centered when the detail is open.
4. Currently, the focus on detail heading only has yellow border. But just like the tertiary button, on focus the background should be grey as well as the yellow border.

ArakTaiRoth (approved):

apps/angular-demo/src/app/detail/detail.html
```diff
@@ -0,0 +1,30 @@
+<h2>Detail</h2>
```
vanessatran-ddi, line 1: Suggestion: Can we rename the file from detail.html to detail.component.html?

apps/angular-demo/src/app/detail/detail.ts
```diff
@@ -0,0 +1,9 @@
+import { Component } from "@angular/core";
```
vanessatran-ddi, line 1: Suggestion: Can we rename detail.ts to detail.component.ts?

libs/docs/src/components/common/Details.stories.mdx
```diff
@@ -0,0 +1,111 @@
+import { Meta, Story } from "@storybook/addon-docs";
+import {
+  Props,
+  Prop,
+  Tabs,
+  Tab,
+  CodeSnippet,
+  SupportInfo,
+  Markdown,
```
vanessatran-ddi, line 9: Picky: "Markdown" is un-used import.

vanessatran-ddi (approved):
I just have some comments. The rest is good.

PR #1020, opened 2023-01-20
feat(DDI-331): Added doc for supported browsers

Radha-Alberta:
Added storybook page for supported browsers and devices.

### DIFF
--- a/libs/docs/src/BrowserSupport.stories.mdx
+++ b/libs/docs/src/BrowserSupport.stories.mdx
@@ -1,0 +1,65 @@
+import { Meta } from "@storybook/addon-docs";
+import { GoATable } from '@abgov/react-components';
+
+<Meta title={`Supported browsers`} />
+
+# Supported browsers
+
+The design system components are tested on the following browsers and devices:
+
+<GoATable width="90%">
+  <tr>
+    <th colSpan="2">Browsers</th>
+    <th colSpan="4">Platforms</th>
+    <th>Representative Mobile OS</th>
+  </tr>
+  <tr>
+    <th>Name</th>
+    <th>Version</th>
+    <th>Windows</th>
+    <th>Mac OS X</th>
+    <th>Android</th>
+    <th>iOS</th>
+    <th></th>
+  </tr>
+  <tr>
+    <td>Google Chrome</td>
+    <td>Latest* and Latest -1**</td>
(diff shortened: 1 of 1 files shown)


libs/docs/src/BrowserSupport.stories.mdx
```diff
@@ -0,0 +1,65 @@
+import { Meta } from "@storybook/addon-docs";
+import { GoATable } from '@abgov/react-components';
+
+<Meta title={`Supported browsers and devices`} />
+
+# Supported browsers and devices
+
+The design system components are tested on the following browsers and devices:
+
+<table width="75%">
```
ArakTaiRoth, line 10: I would recommend using our GoATable, instead of the normal HTML table

libs/docs/src/BrowserSupport.stories.mdx
```diff
@@ -0,0 +1,65 @@
...
+    <td>Supported</td>
+    <td>N/A</td>
+    <td>Supported</td>
+    <td><b>Android OS:</b> 10+, <b>iOS:</b> 15+</td>
+  </tr>
+</table>
+
+*Latest: The latest stable version
+
+*Latest -1: The version prior to the latest version.
```
ArakTaiRoth, line 65: Change this to have 2 *'s. And then change in the table to have `Latest* and Latest - 1**`, and only have that in one cell. The other cells would all just have `Latest and Latest - 1`

libs/docs/src/BrowserSupport.stories.mdx
Radha-Alberta: Done

libs/docs/src/BrowserSupport.stories.mdx
```diff
@@ -0,0 +1,65 @@
+import { Meta } from "@storybook/addon-docs";
+import { GoATable } from '@abgov/react-components';
+
+<Meta title={`Supported browsers and devices`} />
```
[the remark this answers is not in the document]
Radha-Alberta: Done

libs/docs/src/BrowserSupport.stories.mdx
Radha-Alberta: Done

ArakTaiRoth (approved):

PR #1021, opened 2023-01-24
feat(DDI-1148): Update two column layout doc

Radha-Alberta:
Added the descriptions for examples in two column layout doc

### DIFF
--- a/libs/docs/src/components/common/TwoColumnLayout.stories.mdx
+++ b/libs/docs/src/components/common/TwoColumnLayout.stories.mdx
@@ -13,6 +13,7 @@ import {
   GoATwoColumnLayout,
   GoAAppHeader,
   GoAAppFooter,
+  GoACallout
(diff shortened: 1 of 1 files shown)


libs/docs/src/components/common/TwoColumnLayout.stories.mdx
```diff
@@ -62,6 +59,8 @@ import {
 
 #### Basic
 
+For situations where the default values of 256px for the nav column and 100% max-content width are desirable.
```
ArakTaiRoth, line 62: I would use a GoACallout component, type information for displaying this text, and all the other new text in this document

Radha-Alberta: Added description in the callout

chrisolsen (approved):

ArakTaiRoth (approved):

PR #1025, opened 2023-01-26
1169 table sort

chrisolsen:


chrisolsen:
As I previously mentioned Firefox does not yet provide the ability to use `:has()` and results in this unwanted gap, but this has been discussed with Tom and we are in agreement that this is ok for now. 

As for the chrome showing the same issue, this is something that does *not* exist when testing with an actual Android device.

libs/web-components/src/components/table/Table.svelte
```diff
@@ -71,30 +112,38 @@
     top: 0;
   }
   td {
-    padding: 0.75rem 1rem 0.5rem;
+    padding: 0.75rem 1rem;
```
ArakTaiRoth, line 115: Why did we change the padding on the td elements, and for the relaxed variant as well?

libs/web-components/src/components/table/TableSortHeader.spec.ts
```diff
@@ -0,0 +1,17 @@
+import { render } from '@testing-library/svelte'
```
ArakTaiRoth, line 1: I think we should expand these tests, to verify that sorting will actually work when implemented.

libs/web-components/src/components/table/TableSortHeader.svelte
```diff
@@ -0,0 +1,59 @@
...
+  {:else}
+    <img class={direction} alt="" src="data:image/svg+xml;base64,PD94bWwgdmVyc2lvbj0iMS4wIiBlbmNvZGluZz0iVVRGLTgiIHN0YW5kYWxvbmU9Im5vIj8+CjxzdmcKICAgd2lkdGg9IjE2IgogICBoZWlnaHQ9IjIyIgogICB2aWV3Qm94PSIwIDAgMTYgMjIiCiAgIGZpbGw9Im5vbmUiCiAgIHZlcnNpb249IjEuMSIKICAgaWQ9InN2ZzEwNjgiCiAgIHhtbG5zPSJodHRwOi8vd3d3LnczLm9yZy8yMDAwL3N2ZyIKICAgeG1sbnM6c3ZnPSJodHRwOi8vd3d3LnczLm9yZy8yMDAwL3N2ZyI+CiAgPGRlZnMKICAgICBpZD0iZGVmczEwNzIiIC8+CiAgPHBhdGgKICAgICBkPSJtIDAuNTk0Mzc0NzEsNy45MDkwOTkgNi41NTIxODc5OSw3LjY0NjI1IGMgMC4xMDU2MSwwLjEyMzE4IDAuMjM2NjIsMC4yMjIwNiAwLjM4NDA0LDAuMjg5ODUgMC4xNDc0MiwwLjA2Nzc5IDAuMzA3NzYsMC4xMDI4OSAwLjQ3MDAyLDAuMTAyODkgMC4xNjIyNiwwIDAuMzIyNiwtMC4wMzUxIDAuNDcwMDIsLTAuMTAyODkgMC4xNDc0MiwtMC4wNjc3OSAwLjI3ODQzLC0wLjE2NjY3IDAuMzg0MDQsLTAuMjg5ODUgbCA2LjU1MjE4MDMsLTcuNjQ2MjUgYyAwLjYyNTMsLTAuNzI5ODQgMC4xMDY5LC0xLjg1NzE4OCAtMC44NTQxLC0xLjg1NzE4OCBIIDEuNDQ2NTYyNyBjIC0wLjk2MDkzNzk5LDAgLTEuNDc5Mzc1NjksMS4xMjczNDggLTAuODUyMTg3OTksMS44NTcxODggeiIKICAgICBmaWxsPSIjMDA3MGM0IgogICAgIGlkPSJwYXRoMTA2NiIgLz4KPC9zdmc+Cg==" />
```
ArakTaiRoth, line 57: `alt="sorted"`?

libs/web-components/src/components/table/Table.svelte
chrisolsen: Because the css was changed a while back resulting in the text being pushed to the bottom of the table row.

libs/web-components/src/components/table/TableSortHeader.spec.ts
chrisolsen: The table sort header doesn't perform any sorting, it just shows the user the current sort status; so there is nothing more to test.

libs/web-components/src/components/table/TableSortHeader.svelte
chrisolsen: `alt={direction}` was added

ArakTaiRoth (dismissed):

ArakTaiRoth (approved):

PR #1027, opened 2023-01-30
feat(DDI-1209): create stepper component

chrisolsen:


Radha-Alberta:
In mac safari and iOS, Focus (yellow border) is not visible when clicking on step.

https://user-images.githubusercontent.com/107069489/219506257-ada5efe3-fcd4-4915-9322-50195071191c.mp4

Radha-Alberta:
- Clicking spacebar is not selecting the step (It is just scrolling the page)

https://user-images.githubusercontent.com/107069489/219793633-c193d44a-b322-4c16-b4f4-4e75d71cb0dc.mp4

- On windows: black border is seen on steps that are not clickable

https://user-images.githubusercontent.com/107069489/219793797-5d480880-5d49-4557-a0b2-e6d4aff94c60.mp4

Radha-Alberta:
- _change prop is missing in the Properties table.
- The line height for text should be 20px. Currently it is 24px.
- The spacing between node and text should be 14px. Currently it is 8px.

libs/react-components/src/lib/form-step/form-step.tsx
```diff
@@ -0,0 +1,37 @@
...
+  text: string;
+  subText: string;
+  status: Status;
+}
+
+export function GoAFormStep(props: FormStepProps) {
+  return (
+    <goa-form-step
+      text={props.text}
+      subtext={props.subText}
```
ArakTaiRoth, line 31: I don't see `subtext` in the Svelte component. What is this used for?

libs/web-components/src/components/form-step/FormStep.svelte
```diff
@@ -0,0 +1,173 @@
+<svelte:options tag="goa-form-step" />
```
ArakTaiRoth, line 1: We don't have `FormStep` documented in Storybook. We just use it within the context of `FormStepper`. I'd recommend maybe something like what we do for `Dropdown` and `DropdownItem`, including them both on the same page, but making it clear they have their own properties.

libs/web-components/src/components/form-step/FormStep.svelte
```diff
@@ -0,0 +1,173 @@
...
+
+<script lang="ts">
+  import { onMount } from "svelte";
+  import { toBoolean } from "../../common/utils";
+
+  type FormStepStatus = "incomplete" | "complete" | "";
```
ArakTaiRoth, line 7: Just thinking about the statuses. There are 3 statuses, `incomplete`, `complete`, or `inprogress`. It might feel weird for users to implement some statuses as just blank, I would say what you're currently using `incomplete` for would be `inprogress`, and what you're currently using blank for, would be `incomplete`.

libs/web-components/src/components/icon/Icon.svelte
```diff
@@ -537,6 +537,19 @@
   export let title: string = "";
   export let testid: string = "";
 
+  // Private
+  const _iconOverrides = {
+    pencil: `<svg width="18" height="18" viewBox="0 0 18 18" fill="none" xmlns="http://www.w3.org/2000/svg">
```
ArakTaiRoth, line 542: With these overrides being introduced, will this affect people currently using one of these icons and change them?

libs/web-components/playground/src/pg-form-stepper.svelte
```diff
@@ -0,0 +1,70 @@
+<svelte:options tag="pg-form-stepper" />
```
ArakTaiRoth, line 1: So I get that you made a playground to directly test Svelte components, but I don't understand what this is for?

libs/web-components/src/components/pages/Pages.svelte
```diff
@@ -0,0 +1,52 @@
+<svelte:options tag="goa-pages" />
```
ArakTaiRoth, line 1: We have no documentation for this `Pages` component in Storybook. If this is only to be used with the `FormStepper`, then maybe we include the properties for it on the `FormStepper` documentation?

libs/web-components/playground/src/pg-form-stepper.svelte
chrisolsen: It set the web component tag name

libs/web-components/src/components/icon/Icon.svelte
chrisolsen: Yes it would

libs/web-components/src/components/form-step/FormStep.svelte
chrisolsen: incomplete and inprogress seem like the same thing to me. The blank value would be equiv to `notstarted`, but the users will never set anything to `not started` since that, or blank, is the initial value. Don't think of `""` as a status, it is just a null status.

libs/web-components/src/components/pages/Pages.svelte
chrisolsen: Can we create a new ticket for this?

libs/react-components/src/lib/form-step/form-step.tsx
chrisolsen: This is a stale prop and needs to be removed

libs/web-components/src/components/skeleton/Skeleton.svelte
```diff
@@ -8,8 +8,20 @@
...
+    "title", 
+    "text-small", 
+    "avatar", 
+    "header", 
+    "paragraph", 
+    "thumbnail", 
+    "card", 
+    "lines",
+    "profile", 
+    "article",
```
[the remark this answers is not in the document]
chrisolsen: This was added to allow simplify the examples shown in storybook for the FormStepper. Before it becomes official maybe Tom will need to approve/change what the article skeleton looks like.

libs/docs/src/components/common/FormStepper.stories.mdx
```diff
@@ -0,0 +1,414 @@
...
+    if (page < 1 || page > 4) return;
+    setStep(page);
+  }
+  return (
+    <>
+      <GoAFormStepper step={step} onChange={(step) => setStep(step)}>
+        <GoAFormStep text="Personal details" status={status[0]} />
+        <GoAFormStep text="Employment history" status={status[1]} />
+        <GoAFormStep text="References" status={status[2]} />
+        <GoAFormStep text="Review" status={status[3]} />
```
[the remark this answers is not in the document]
chrisolsen: Because status[3] is just a null value, but an additional status will be added.

ArakTaiRoth (approved):

ArakTaiRoth (approved):

vanessatran-ddi (approved):

libs/docs/src/components/common/FormStepper.stories.mdx
```diff
@@ -0,0 +1,435 @@
...
+            <GoAButton type="tertiary" onClick={() => setPage(step + 1)}>Next</GoAButton>
+          </GoAButtonGroup>
+        </>
+      )  
+    `}/>
+  </Tab>
+</Tabs>
+
+export const StatusTemplate = () => {
+  const [step, setStep] = useState(-1);
```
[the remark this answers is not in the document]
chrisolsen: The StatusTemplate to is to show off the status attributes and -1 (which is the default value) allows the user to click on any of the items, so things do work as expected. Since the first 2 form steps are marked as complete we could set it to `3`, but hardcoding a step value is something that would never be done and it is probably best not to show an example like that.

PR #1028, opened 2023-01-31
chore(DDIDS-1175): Add testids to react components

amar-ABgov:


libs/react-components/src/lib/textarea/textarea.tsx
```diff
@@ -88,7 +89,7 @@ export const GoATextArea: FC<Props> = ({
       maxcharcount={maxCharCount}
       width={width}
       error={error}
-      data-testid={testId}
+      testid={testId}
```
chrisolsen, line 92: Having an overlap of the previous `testid` attribute probably won't deliver the expected results. The reason for this is that the `testid` we were  using was often (not all the time) bound to an internal element, that when tested from outside of the web component, will reside within the shadow DOM, which will prevent the normal DOM queries to fail. 

So instead of binding it to `testid` bind it to `data-testid` like is done in the other components.

libs/react-components/src/lib/checkbox/checkbox.spec.tsx
```diff
@@ -32,7 +32,7 @@ describe("GoA Checkbox", () => {
     expect(checkbox.getAttribute("disabled")).toBe("false");
     expect(checkbox.getAttribute("checked")).toBe("true");
     expect(checkbox.getAttribute("error")).toBe("false");
-    expect(checkbox.getAttribute("data-testid")).toBe(testId);
```
ArakTaiRoth, line 35: Why did we make this change? I don't remember seeing this failing, and `data-testid` should be the correct property getting used.

libs/react-components/src/lib/input/input.spec.tsx
```diff
@@ -58,7 +58,7 @@ describe("Input", () => {
     expect(input.getAttribute("placeholder")).toBe("placeholder");
     expect(input.getAttribute("prefix")).toBe("foo");
     expect(input.getAttribute("suffix")).toBe("bar");
-    expect(input.getAttribute("data-testid")).toBe(testId);
```
ArakTaiRoth, line 61: Same comment as above, shouldn't need to make this change, so long as we're not changing `data-testid` to `testid` in the component.

ArakTaiRoth (approved):

chrisolsen (approved):

PR #1039, opened 2023-02-21
fix: Just to force a new release for angular components

ArakTaiRoth:
This is being done mainly to force NPM to update to a new angular components release so that we can get updated documentation on NPM.

This has a side effect that I don't believe the `-f` is actually necessary


vanessatran-ddi (approved):
👍

PR #1040, opened 2023-02-23
add modal scroll shadow

chrisolsen:


libs/react-components/src/lib/modal/modal.tsx
```diff
@@ -10,7 +10,7 @@ export type CalloutVariant =
 
 interface WCProps {
   ref: React.RefObject<HTMLElement>;
-  heading?: string;
+  heading?: React.ReactNode;
```
ArakTaiRoth, line 13: This is a breaking change. No breaking change commit was made. Also, I don't believe this is necessary for this story, so I don't want to introduce a breaking change just for this. If we want this, then I would recommend we add this to a list of changes we want to make that are breaking changes, and make all of them at the same time.

ArakTaiRoth: To elaborate further, @Radha-Alberta pointed out that in the story it mentions the need for *bold* text in the header, and I see that's likely why we did this. I spoke with Tom about this and this was not listed under Acceptance Criteria because it wasn't needed, it was in the story erroneously.

libs/web-components/src/common/no-scroll.ts
```diff
@@ -6,13 +6,10 @@ export default function(_node: HTMLElement, opts: { enable: boolean }) {
     if (!isScrollable()) {
       return;
     }
-
     const scrollbarWidth = calculateScrollbarWidth();
```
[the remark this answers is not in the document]
chrisolsen: No, it is still used on line 12.

libs/react-components/src/lib/modal/modal.tsx
chrisolsen: It isn't a breaking change. ReactNode can be a string

libs/web-components/src/components/focus-trap/FocusTrap.svelte
```diff
@@ -104,14 +104,13 @@
...
   }
 
-
   function focus(element: HTMLElement): boolean {
     if (!isFocusable(element)) return false;
 
     try {
       ignoreFocusChanges = true;
       element.focus();
-    } catch(e) {}
```
chrisolsen, line 114: This shouldn't have been removed...my bad

libs/react-components/src/lib/modal/modal.tsx
ArakTaiRoth: @chrisolsen Yes, the slot can be a string, but no one is using a slot right now. Anyone currently using the Modal component, is using a string heading property, not a slot. And that doesn't work anymore with this change, thus it will break everyone's current implementation of the Modal component. And yea, as stated before, this was in the story erroneously, so should be removed even just for that.

chrisolsen: @ArakTaiRoth I didn't realize that the that the header was removed in the web component as well :+1:

chrisolsen: @ArakTaiRoth The ticket does ask for the ability to mix bold and normal text
> some designs want to be able to style the heading to something like “John Doe born on 1930-10-10 exists more than once” with a mix of bold/regular font but this is not possible as the heading just takes a string. 

This doesn't have to be a breaking change, but the react component will require the change that was made to allow for this.

ArakTaiRoth (approved):

PR #1059, opened 2023-03-13
chore (DDI-1271): Update container docs

amar-ABgov:


libs/docs/src/components/common/Container.stories.mdx
```diff
@@ -357,19 +357,35 @@ export const HeadingTemplate = (args) => {
...
     <GoAGrid minChildWidth="200px">
-      <GoAContainer accent="thick" title="Heading">
+      <GoAContainer accent="thick" title={<div>Group Heading</div>}>
         Content
       </GoAContainer>
       <GoAContainer
         type="non-interactive"
         accent="thick"
-        title="Heading"
+        title={<div>Group Heading</div>}
```
chrisolsen, line 372: The `title` attribute will accept a string, so it just a string so this example will be misleading.

chrisolsen: However, an example should be included that contains a more complex header that will utilize the ReactNode/slot change that was made.

Radha-Alberta:
- The highlighted example (from storybook doc) is giving error when run in angular demo app.

image
image

- The title "The container title" is missing 
image
image

chrisolsen (dismissed):

ArakTaiRoth (dismissed):

ArakTaiRoth (approved):

chrisolsen (approved):

PR #1063, opened 2023-03-14
fix(DDI-1269): remove margin-bottom of last <p> tag

Radha-Alberta:
Set margin-bottom to 0 of the last <p> tag from Accordion, Container, and Details components.

### DIFF
--- a/libs/web-components/src/components/container/Container.svelte
+++ b/libs/web-components/src/components/container/Container.svelte
@@ -98,6 +98,10 @@
     border-bottom-right-radius: var(--goa-border-radius-m);
   }
 
+  .content ::slotted(p:last-child) {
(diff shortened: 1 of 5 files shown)


libs/web-components/src/components/container/Container.svelte
```diff
@@ -98,6 +98,10 @@
     border-bottom-right-radius: var(--goa-border-radius-m);
   }
 
+    .content ::slotted(p:last-child) {
```
chrisolsen, line 101: Can you fix this indentation?

vanessatran-ddi (approved):
I tested it locally and it worked for me. 
Although I am not a fan of "!important" I guess this is needed for this case.

image

ArakTaiRoth (approved):

PR #1077, opened 2023-03-28
fix(DDI-1274): Added timeout before validating props

Radha-Alberta:
Console error is seen when variables are used with the following properties

- Badge - type prop
- Callout - type prop
- Microsite Header - type prop
- Notification - type prop
- Skeleton - type prop

### DIFF
--- a/libs/web-components/src/components/badge/Badge.svelte
+++ b/libs/web-components/src/components/badge/Badge.svelte
@@ -43,7 +43,7 @@
   }[type];
 
   onMount(() => {
-    validateType(type);
+    setTimeout(() => validateType(type), 1);
     if (!showIcon && !content) {
       console.warn("GoABadge must have either then content or icon property set");
     }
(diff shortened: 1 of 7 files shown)


libs/web-components/src/components/badge/Badge.svelte
```diff
@@ -43,7 +43,7 @@
   }[type];
 
   onMount(() => {
-    validateType(type);
+    setTimeout(() => validateType(type), 1);
```
chrisolsen, line 46: Wrapping all the functions with a timeout is a doesn't feel great. Have you tried other, cleaner, alternatives?
- adding a `await tick()` to the first line within the `onMount` function
- or adding the await into the validateType method
- or adding the setTimeout into the validateType method

ArakTaiRoth: @chrisolsen I'm not sure adding this to the `validateType` method (your last 2 options) makes sense, there might be instances we don't want to do that. But your first option is something I think we're already doing in other places, so definitely makes the most sense to me.

chrisolsen: @ArakTaiRoth This is just validating parameters, so it will never matter when it's done in different scenarios.

Radha-Alberta: @chrisolsen As discussed, I have left setTimeout as is (as first option might slow down screen rendering and last 2 options are still giving error) and added comment above typeValidator.

chrisolsen (approved):

PR #1085, opened 2023-04-03
feat(DDI-1273): added interaction states on input

Radha-Alberta:
If leading or trailing content slots exists, then interaction states will be shown only on input box.

### DIFF
--- a/libs/web-components/src/components/input/Input.svelte
+++ b/libs/web-components/src/components/input/Input.svelte
@@ -63,6 +63,7 @@
   $: isDisabled = toBoolean(disabled);
 
   let inputEl: HTMLElement;
+  let _rootEL: HTMLElement;
   $: if (isFocused && inputEl) {
     setTimeout(() => inputEl.focus(), 1);
   }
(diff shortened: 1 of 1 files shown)


ArakTaiRoth (approved):

libs/web-components/src/components/input/Input.svelte
```diff
@@ -299,9 +319,9 @@
   .goa-input--disabled:active,
   .goa-input--disabled:focus {
     background-color: var(--goa-color-greyscale-100);
-    border-color: var(--goa-color-greyscale-200);
+    border-color: var(--goa-color-greyscale-200) !important;
```
chrisolsen, line 322: Are these `!important`s needed? Usually they can be avoided by either increasing the css selectivity or moving them further down in the css file.

Radha-Alberta: Increasing the css selectivity or moving it further down in the css is not working. So as discussed, leaving !important as is.

PR #1086, opened 2023-04-03
feat (DDI-1259): Update popover position to be dynamic

amar-ABgov:


libs/docs/src/components/common/Popover.stories.mdx
```diff
@@ -58,6 +58,40 @@ import {
   />
```
ArakTaiRoth, line 58: There's no example for dynamic positioning in the Storybook?

libs/web-components/src/components/popover/Popover.svelte
```diff
@@ -11,6 +12,15 @@
 
   let _isContentVisible = false;
   let _targetEl: HTMLElement;
+  let _popoverEl: HTMLElement;
+  let _popoverYPos: number = 0;
+
+  // margin
+  export let mt: Spacing = null;
```
ArakTaiRoth, line 19: These need to be added to the `*.html-data.json` file. Take a look at other files for an example.

ArakTaiRoth (approved):

vanessatran-ddi (approved):

PR #1089, opened 2023-04-05
fix(DDI-1302): Removed the alternative colouring for icon-overrides

ArakTaiRoth:


Radha-Alberta:
- [ ] Pencil and remove icons color in Form steeper component has changed to black. (Image 1)

#### Image 1
image

vanessatran-ddi (approved):

libs/web-components/src/components/form-step/FormStep.svelte
```diff
@@ -92,6 +92,11 @@
     width: 2.5rem;
   }
 
+  .status > * {
```
chrisolsen, line 95: What broke this?

chrisolsen (changes requested):
What broke this? Is there a reason that more is being added instead of reverting things to how they were when things worked.

PR #1096, opened 2023-04-17
fix(DDI-1279): updated form steeper doc

Radha-Alberta:
Updated form steeper documentation


libs/docs/src/components/common/FormStepper.stories.mdx
```diff
@@ -96,7 +96,15 @@ import {
...
-#### Basic
+#### Open navigation
+
+If the service requires, you can allow the user to move forward to a new step before fully completing a step. In this case, the form validation should be run when the user attempts to submit the form at the end of the form stepper process.
+
+- The user can use the stepper to navigate between steps.
+
+- The user does not have to complete all of the information to move to the next step.
+
+“-1” is the default value, so you can leave it blank if using “Open navigation”
```
ArakTaiRoth, line 107: I don't think including this line makes sense if we're not including the code as well. Maybe verify with Mark, but that's my opinion.

libs/docs/src/components/common/FormStepper.stories.mdx
```diff
@@ -179,12 +187,22 @@ export const BasicTemplate = () => {
...
+
+The user needs to partially or completely finish a step to be able to move forward to the next step. In this case:
+
+- A step that is “Not started” will not be clickable.
+
+- A user cannot use the stepper to navigate to another page.
+
+- Clicking the Active step when you are on that step will do nothing. (no page refresh).
+
+To use the controlled type you must set a step value ≥ 1
```
ArakTaiRoth, line 205: I don't think including this line makes sense if we're not including the code as well. Maybe verify with Mark, but that's my opinion.

ArakTaiRoth (approved):

vanessatran-ddi (approved):

PR #1102, opened 2023-04-19
fix(DDI-1294): adjusted compact button padding

Radha-Alberta:
Added left and right padding for compact button.


vanessatran-ddi (approved):
👍

PR #1113, opened 2023-05-02
fix(DDI-1147): Update notification banner close button - focus state …

vanessatran-ddi:
## Description
When the documentation was being updated, it was noticed that the focus states for the notification banner were not all accessible. This has been addressed in the design and now the component needs to be updated in code.

Updated:

- Update the hover and focus treatment for all notification banner types
- Close icon button for “important” notification banner
- Ensure that if a text link is present in the Notification banner that it receives focus before the close button.
- Code each banner type as an Aria Live-region. . Use polite for every banner type.

## User Story
https://goa-dio.atlassian.net/browse/DDIDS-1147

## Screenshots
image

image


Radha-Alberta:
- [ ] When focused over the link in important banner, a thin white border can be seen. (Image 1)
- [ ] On hover over the close button, the color of the close icon is changing to blue. (Image 2)

#### Image 1
image

#### Image 2
image

Radha-Alberta:
1. On focus over link text in important banner, the border has sharp corners. It should have round corners. (image 1)
2. The round box that appears on hover/focus over close button has more empty space in the bottom. In Figma, it is even on all the sides. (image 2 & 3)
3. The border that appears on focus over the link should have border-radius of 4px. (image 4)

image 1
image

image 2
image

image 3
image

image 4
image

vanessatran-ddi:
> 1. On focus over link text in important banner, the border has sharp corners. It should have round corners. (image 1)
> 
>     2. The round box that appears on hover/focus over close button has more empty space in the bottom. In Figma, it is even on all the sides. (image 2 & 3)
> 
>     3. The border that appears on focus over the link should have border-radius of 4px. (image 4)
> 
> 
> image 1 image
> 
> image 2 image
> 
> image 3 image
> 
> image 4 image

Hi @Radha-Alberta 
All above are done.

ArakTaiRoth:
Overall question about this before I review any further. In the story it says every banner should be coded as an arialive region with a value of polite. But in this code, we're giving the power to the user to determine what type of arialive region they have, which means they can change it away from polite to something else. Was this change run by Tom and Mark?

thytran142:
> Overall question about this before I review any further. In the story it says every banner should be coded as an arialive region with a value of polite. But in this code, we're giving the power to the user to determine what type of arialive region they have, which means they can change it away from polite to something else. Was this change run by Tom and Mark?

Hi @ArakTaiRoth 
Yes good question.
I checked with Thomas about it. The live regions have 3 attributes: off, polite and assertive. 
Emergency & warning should be assertive, while event  and information should be polite (That's the recommendation on the [ARIA Live Regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions)) 
Normally, only aria-live="polite" is used. Any region which receives updates that are important for the user to receive, but not so rapid as to be annoying, should receive this attribute. The screen reader will speak changes whenever the user is idle.
aria-live="assertive" should only be used for time-sensitive/critical notifications that absolutely require the user’s immediate attention. Generally, a change to an assertive live region will interrupt any announcement a screen reader is currently making. As such, it can be extremely annoying and disruptive and should only be used sparingly.

So Thomas agreed with me that we should allow developers to make a decision depending on the business logic of the alert. The default value is polite as in the figma documentation if there is no preference declared, and the storybook is written to guide ppl about it.

libs/web-components/src/components/notification/Notification.html-data.json
```diff
@@ -12,10 +12,16 @@
       "description": "Dispatched when notification banner is closed."
     },
     {
-      "name": "maxContentWidth",
+      "name": "maxcontentwidth",
```
ArakTaiRoth, line 15: In the storybook, the property is still defined as `maxContentWidth` for angular. Either that should be changed, or this should be changed.

libs/web-components/src/components/notification/Notification.svelte
```diff
@@ -54,8 +62,11 @@
       </div>
       <div class="close">
         <!-- svelte-ignore a11y-click-events-have-key-events -->
-          <goa-icon-button on:click={close} icon="close" variant="dark"
-          inverted={type === "important" ? "false" : "true"}/>
+          <button
+            class={type}
+            on:click={close}>
+            <goa-icon type="close" inverted={type === "important" ? "false" : "true"} />
```
ArakTaiRoth, line 68: I'm questioning our change here to using `button` instead of `goa-icon-button` or at the very least `goa-button`? Why did we change this to begin with? And then, why aren't we using our own `goa-button`?

vanessatran-ddi: Hi @ArakTaiRoth , 
Yes it is a good question, that is my first approach when I do this ticket, but then I revert back to the simple button element for the close button here. 
This is the code change: 
image


The effect that if we use goa-button with inverted = false for important notification banner, the button will be like below: 
image
Which we don't want. In order for me to change the border color to the below expected close button: 
image

I will need to come to css inverted of goa-icon-button to modify there. However during the feedbacks, we have more changes that need to address: 
https://github.com/GovAlta/ui-components/pull/1113#issuecomment-1532236340
Which includes: the background color of the close button (It depends on the background color of the notification), the hover effects, the focus effect (which is different from the way goa-icon-button is doing, especially for the close button with its purpose under notification banner). 

I find out if I do a lot of changes under goa-icon-button, it may affect other components that are using the goa-icon-button too. To prevent the conflicts or changes that I may not know, and the close button here is simple to implement, I do it as an alone button element, with css. In the future, it will be easier to maintain, assume if we have more notification backgrounds or we want to change the effects/styles of the close button.

libs/web-components/src/components/notification/Notification.html-data.json
vanessatran-ddi: Yes I updated our storybook. In Angular the correct property is "maxcontentwidth".

libs/react-components/src/lib/notification/notification.tsx
```diff
@@ -31,6 +35,7 @@ interface Props {
 
 export const GoANotification = ({
   type = "information",
+  ariaLive = "polite",
```
chrisolsen, line 38: Default values should not be set in the React component, but instead in the svelte file. Otherwise it is possible for the React default value to differ from the svelte.

libs/web-components/src/components/icon-button/IconButton.svelte
```diff
@@ -86,7 +86,7 @@
     background: transparent;
     cursor: pointer;
     border: none;
-    border-radius: 0.5rem;
+    border-radius: var(--goa-border-radius-m);
```
chrisolsen, line 89: The current value of `--goa-border-radius-m` is ` 0.25rem`, so is this a desired change?

libs/web-components/src/components/notification/Notification.svelte
```diff
@@ -68,40 +79,36 @@
...
     .notification {
-      padding: 1.5rem 2rem;
+      padding: var(--goa-space-l) var(--goa-space-xl);
     }
   }
 
   @media (min-width: 1024px) {
     .notification {
-      padding: 1.5rem 4.5rem;
+      padding: var(--goa-space-l) 4.5rem;
```
chrisolsen, line 94: Maybe these padding values should be relative to on another. So in this case
```
padding: var(--goa-space-l) calc(3 * var(--goa-space-l));
```
The same with the others.

chrisolsen (changes requested):

libs/web-components/src/components/icon-button/IconButton.svelte
vanessatran-ddi: @chrisolsen Yes this is a request from Thomas.

chrisolsen (approved):

ArakTaiRoth (approved):

PR #1116, opened 2023-05-08
fix(DDI-1337): Update table sortable doc

Radha-Alberta:
Updated table sortable example to include direction prop


ArakTaiRoth:
- [ ] When setting a direction, either `asc` or `desc` on creation, it doesn't actually sort properly
image
The above image shows that `asc` was picked, but the table is not sorted properly, this is what it looks like after I click it to sort it:
image

PR #1118, opened 2023-05-09
feat(DDI-1309): Update the accessibility of badge component.

vanessatran-ddi:
### Description
Update badge component so that the screen reader can parse them. 
Talk to Mark and Dustin about it. This is the  [demo site](https://vanessatran-ddi.github.io/angular-ui-component-sample/badge) that we discussed and agreed on, especially about how the screen reader should read/announce it. 

### Screenshots:
image

image

https://github.com/GovAlta/ui-components/assets/120135417/de52a0c8-d240-4f5e-824e-1038631546ac

image

### DIFF
--- a/libs/web-components/src/components/badge/Badge.svelte
+++ b/libs/web-components/src/components/badge/Badge.svelte
@@ -5,7 +5,7 @@
   import { onMount } from "svelte";
   import type { Spacing } from "../../common/styling";
   import { calculateMargin } from "../../common/styling";
-  import { typeValidator, toBoolean } from "../../common/utils";
+  import {typeValidator, toBoolean} from "../../common/utils";
 
   // Validator
   const [Types, validateType] = typeValidator(
@@ -23,6 +23,7 @@
   export let testid: string = "";
   export let content: string = "";
   export let icon: string = "false";
+  export let arialabel: string = "";
 
   // margin
   export let mt: Spacing = null;
@@ -30,7 +31,9 @@
   export let mb: Spacing = null;
   export let ml: Spacing = null;
 
+  // private
   $: showIcon = toBoolean(icon);
+  $: showIconOnly = showIcon && !content;
 
   $: iconType = {
     success: "checkmark-circle",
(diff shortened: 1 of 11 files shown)


Radha-Alberta:
- [ ] In windows the screen reader is not reading the arialabel value on mouse hover over badge. (Video 1)

#### Video 1

https://github.com/GovAlta/ui-components/assets/107069489/25e6614f-c0a7-4898-b1c3-1c0e1e456163

vanessatran-ddi:
> * [ ]  In windows the screen reader is not reading the arialabel value on mouse hover over badge. (Video 1)
> 
> #### Video 1
>  UI.Components.Angular.Template.-.Google.Chrome.2023-05-09.19-49-46.mp4

https://github.com/GovAlta/ui-components/assets/120135417/f685eb1a-70f2-4e8b-bbe2-d17bc6326afc

libs/web-components/src/components/badge/Badge.svelte
```diff
@@ -59,17 +60,29 @@
...
+        size="small"
+      />
   {:else}
-    <div style="height: 1.2rem; margin-left:-0.25rem;" />
+  <div style="height: 1.2rem; margin-left:-0.25rem;" />
   {/if}
   {#if content}
     <div class="goa-badge-content">
+      {#if arialabel}
+        <span class="sr-only">{arialabel},</span>
```
vanessatran-ddi, line 77: Note: 
1. I need to add the aria-label here because when we hover a badge, we want it to read this span text. 
https://github.com/GovAlta/ui-components/pull/1118#issuecomment-1564856408
2. I need the screen reader to have a small "pause" between aria-label and the real content. 
3. The reason why I don't use real "aria-label" is because the text is not an interactive element (this case is a badge). 
https://stackoverflow.com/questions/39223952/accessibility-sr-only-or-aria-label

chrisolsen (approved):

ArakTaiRoth (approved):

PR #1119, opened 2023-05-10
feat(DDI-1284): app header updates

chrisolsen:


Radha-Alberta:
- [ ] In Figma, all the headings (with and without menu) are in black color. In storybook, only the heading with menu is in black color. (Image 1 and Image 2)
- [ ] Same with the mobile view (Image 3). When the menu is open, all the navigation items should be in black color as per Figma (Image 4)
- [ ] The color of the "Menu" text in mobile should be black.

#### Image 1
image
#### Image 2
image
#### Image 3
image
#### Image 4
image

libs/react-components/src/lib/popover/popover.tsx
```diff
@@ -1,9 +1,12 @@
 import React, { FC, ReactNode } from "react";
 import { Margins } from "../../common/styling";
 
+type Position = "above" | "below" | "auto";
```
vanessatran-ddi, line 4: Maybe we should export Position here, because when I checked with React 18 (from our template), there is a warning about the type. 
image

libs/web-components/src/components/app-header/AppHeader.svelte
```diff
@@ -1,131 +1,418 @@
...
+    }
+
+    ::slotted(a:focus),
+    ::slotted(goa-app-header-menu:focus) {
+      outline: var(--goa-border-width-l) solid var(--goa-color-interactive-focus);
+      outline-offset: -3px;
+    }
+
+    ::slotted(a.current) {
+      border-top: 4px solid var(--goa-color-interactive-default);
```
vanessatran-ddi, line 397: Maybe we should use var(--goa-space-2xs) for 4px?

vanessatran-ddi (commented):
I left 2 comments. Everything else is good. I tested and verified on local with React 18, Angular 15 too.

libs/web-components/src/components/app-header/AppHeader.svelte
chrisolsen: I don't think we should mix/match goa-space-x with goa-broder-width-x. Since currently there is no goa-border-width of 4px I think it is best if we leave it like this for now, but we should talk to @twjeffery to see if this new value should be added.

libs/react-components/src/lib/popover/popover.tsx
vanessatran-ddi: Sounds good to me. I was thinking of a case where we dynamically change the position depending on the content (programmatically). 
But it may be not the case here.

vanessatran-ddi (approved):
👍

libs/react-components/src/lib/popover/popover.tsx
chrisolsen: Since the popover already dynamically decides on it's position I think leaving out this control might be a good thing, for now at least.

ArakTaiRoth (approved):

PR #1124, opened 2023-05-29
feat(DDI-1316): add angular goaChecked directive

chrisolsen:


libs/angular-components/src/lib/checked-directive.ts
```diff
@@ -0,0 +1,63 @@
...
+  constructor(protected renderer: Renderer2, protected elementRef: ElementRef) {
+    super(renderer, elementRef);
+  }
+
+  get value(): string {
+    return this._checked ? "checked" : "";
+  }
+
+  set value(checked: any) {
+    console.log("value", !!checked)
```
vanessatran-ddi, line 40: Remove console.log

vanessatran-ddi (commented):
I love this approach to make an angular directive. I have 1 comment to remove console.log.

ArakTaiRoth (approved):

PR #1126, opened 2023-05-31
fix(DDI-1337): Allow for initial table sorting

chrisolsen:


Radha-Alberta:
- [ ] Issue in React: Table data is not loaded on Page load for Safari browsers and Iphone

Iphone_Safari
Iphone_Chrome
Mac_Web_Safari_Desc
Mac_Web_Safari_Asc

libs/web-components/src/components/table/TableSortHeader.svelte
```diff
@@ -21,43 +21,62 @@
...
+  }
+
+  goa-icon {
+    scale: 0.8;
+  }
 
-  img.asc {
-    transform: rotate(180deg);
+  .direction--none goa-icon {
+    height: 10px;
```
vanessatran-ddi, line 59: Maybe we can change it to 'rem'

ArakTaiRoth (approved):

PR #1132, opened 2023-06-09
DDIDS-1322 fix datetime input page crash

Radha-Alberta:


libs/react-components/src/lib/input/input.tsx
```diff
@@ -293,8 +293,8 @@ export const GoAInputDateTime: FC<DateInputProps> = ({
   return (
     <GoAInput
       {...props}
-      onChange={onDateChangeHandler(props.onChange)}
-      value={toString(value, "yyyy-MM-dd'T'HH:mm")}
+      onChange={onDateChangeHandler(props.onChange, "yyyyyy-MM-dd'T'HH:mm:ss'")}
+      value={toString(value, "yyyyyy-MM-dd'T'HH:mm:ss'")}
```
chrisolsen, line 297: Is this a valid parse string?
https://date-fns.org/v2.30.0/docs/format

Is there a reason you are adding seconds to output string?

libs/react-components/src/lib/input/input.tsx
```diff
@@ -209,13 +209,13 @@ export const GoAInput: FC<InputProps & { type?: GoAInputType }> = ({
   );
 };
 
-const onDateChangeHandler = (onChange: OnDateChange) => {
+const onDateChangeHandler = (onChange: OnDateChange, dateFormat?: string) => {
```
chrisolsen, line 212: If an invalid date string is passed in I think it should just fail silently

chrisolsen (changes requested):

libs/react-components/src/lib/input/input.tsx
Radha-Alberta: yes it's a valid parse string for date-fns.
adding seconds in format to avoid Invalid time error. 
image
image

libs/react-components/src/lib/input/input.tsx
Radha-Alberta: yes, it's not accepting invalid dates. 
Example, I tried putting 31st Feb. if you check console in below image, it's not accepting 31st Feb
image

libs/react-components/src/lib/input/input.tsx
```diff
@@ -209,13 +209,18 @@ export const GoAInput: FC<InputProps & { type?: GoAInputType }> = ({
...
```
chrisolsen, line 212: dateFormat is no longer needed. As I mentioned in slack the following is all you should need.

```
const onDateChangeHandler = (onChange: OnDateChange) => {
	return (name: string, value: string) => {
		if (!value) {
		  onChange(name, new Date(0));
		  return;
		}
        // new code
		const d = parseISO(value)
		if (!isNaN(d.getDate())) {
			onChange(name, d);
		}
        // end of new code
	}
}
```

Radha-Alberta: Changes has been updated and it's working as expected.

libs/react-components/src/lib/input/input.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { FC, useEffect, useRef } from "react";
 import { GoAIconType } from "../..";
-import { format, parseISO } from "date-fns";
+import { format, parse, parseISO } from "date-fns";
```
chrisolsen, line 3: remove this `parse` method

libs/react-components/src/lib/input/input.tsx
```diff
@@ -298,7 +304,7 @@ export const GoAInputDateTime: FC<DateInputProps> = ({
     <GoAInput
       {...props}
       onChange={onDateChangeHandler(props.onChange)}
-      value={toString(value, "yyyy-MM-dd'T'HH:mm")}
+      value={toString(value, "yyyy-MM-dd'T'HH:mm:ss'")}
```
chrisolsen, line 307: Is there a reason you are adding seconds to the format string?

chrisolsen (changes requested):

libs/react-components/src/lib/input/input.tsx
```diff
@@ -298,7 +304,7 @@ export const GoAInputDateTime: FC<DateInputProps> = ({
     <GoAInput
       {...props}
       onChange={onDateChangeHandler(props.onChange)}
-      value={toString(value, "yyyy-MM-dd'T'HH:mm")}
+      value={toString(value, "yyyy-MM-dd'T'HH:mm'")}
```
chrisolsen, line 307: There is an extra `'` that needs to be removed

libs/react-components/src/lib/input/input.tsx
```diff
@@ -243,6 +246,9 @@ function toString(value: GoADate, tmpl = "yyyy-MM-dd"): string {
   if (value.toISOString() === new Date(0).toISOString()) {
     return "";
   }
+  if (value.getFullYear().toString().length > 4) {
+    return "";
+  }
```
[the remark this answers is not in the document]
Radha-Alberta: My bad, I forgot to remove that code: it basically restricts user from entering more than 4 year digits, which we don't require now.

PR #1133, opened 2023-06-09
fix(DDI-1361): if the step is marked as complete it should be always …

vanessatran-ddi:
### Description
- step = 2
- All steps are complete
- Cannot click step 3 and 4, must use a button to navigate next and previous

### Screens

https://github.com/GovAlta/ui-components/assets/120135417/8e030160-3530-4528-8c31-9315fb80ec3e

### What have been changed:
1.  Make "status" attribute in React wrapper "GoaFormStep" to be string. It is alerted as "error" in React 18 with a stricter ts type. 
image

2. Make "testId" as optional in React wrapper "GoaFormStepper"

image

3. Make isEnabled under FormStep to rely on "enabled" passed by its parent (controlled navigation) or if its status is complete (as the requirement of this ticket)


libs/web-components/src/components/form-stepper/FormStepper.svelte
```diff
@@ -8,7 +8,7 @@
   // Public
   // ======
 
-  export let step: number = -1;  // this is a 1-based index, -1 is the unset value
+  export let step: number;
```
chrisolsen, line 11: I would think that we want to keep the default value of `-1`. Changing a default value would cause existing usages of the component to behave differently without any warning.

libs/react-components/src/lib/form-step/form-step.tsx
```diff
@@ -1,10 +1,8 @@
 import React from "react";
 
-type Status = "complete" | "incomplete";
-
 interface WCProps {
   text: string;
-  status?: Status;
+  status?: string;
```
chrisolsen, line 5: Since we are loosing typechecking with this change, I would try to find a better way to solve this issue with React 18.

libs/web-components/src/components/form-stepper/FormStepper.svelte
vanessatran-ddi: @chrisolsen 
The reason I marked it as null/undefined is because in Angular 15, the way it renders makes -1 as the value it received (same issue as Table Header sorter last time). 
image

The console log above declares the issue: 
* step is -1 as default
* it set _maxAllowedStep = 4 (= length of children)
* Even if it receives the "real value" step -1 or not later, the _maxAllowedStep is maximum ==> all steps are accessible. 

I tested again to make sure the behaviour doesn't change on React <18, React 18, Angular <15 and Angular 15. @Radha-Alberta  also will do a test to make sure everything is good. 

However if it is your concern, should I add a timeout and reset maxAllowedStep? I think it may make bigger impact than this approach.

libs/react-components/src/lib/form-step/form-step.tsx
vanessatran-ddi: @chrisolsen I was thinking to export Status type under React wrapper so developers can use. However I am not sure if this approach is user-friendly. 
I can do a check on our onMount and console warning if status "incomplete", or "complete" should be 2 choices. How do you think?

libs/web-components/src/components/form-stepper/FormStepper.svelte
chrisolsen: The reason that the `step` value is `-1` is that it is being set to that, so in all other cases where it is using the default value, breaks will occur.

libs/web-components/src/components/form-stepper/FormStepper.svelte
```diff
@@ -39,8 +39,18 @@
   // than the number of steps
   $: _progress = (_maxProgressStep-1) / (_steps.length-1) * 100;
   $: setCurrentStepStatus(step);
-  $: { 
+  $: {
     if (_steps.length) {
+
+      // allow access to all steps if not step property is provided
+      if (step <= 0) {
```
vanessatran-ddi, line 46: This is to fix Angular 15 issue, to let _maxAllowedStep reset depending on the change of "step".

image

libs/react-components/src/lib/form-stepper/form-stepper.tsx
```diff
@@ -41,7 +41,7 @@ export function GoAFormStepper({
     const current = ref.current;
     const changeListener = (e: unknown) => {
       const { step } = (e as CustomEvent).detail;
-      onChange?.(step);
+      onChange?.(+step);
```
vanessatran-ddi, line 44: This is a strange issue on React (both < 17 and 18)
Angular knows "step" is a number, but React thinks it is a string. So the "Next" button doesn't work because new page becomes "21" instead of 3. 

image

chrisolsen (approved):

libs/docs/src/components/common/FormStepper.stories.mdx
```diff
@@ -415,13 +415,17 @@ The status of each step can be configured to either “complete” or “incompl
   </Tab>
   <Tab label="React">
     <CodeSnippet lang="tsx" code={`
-      const [step, setStep] = useState(-1);
+     import {useState} from "react";
```
ArakTaiRoth, line 418: I'm thinking we shouldn't have the `import` statements in the example code. It makes it so the copy code button is essentially useless, because people would then have to copy just the import statements to paste at the top, and then copy the rest of the code for use within their application. But I'm curious if other people have a different take on this, @amar-ABgov , @chrisolsen , @vanessatran-ddi, @Radha-Alberta  ?

vanessatran-ddi: @ArakTaiRoth Yes if we want to make it aligned with the rest of our storybook sample, we should remove those basic imports. 
I was planning to import "GoAFormStepStatusType" only because it removes the warning message for React18 (it has a stricter type check) if we add it under useState line 423 and also potentially some questions from the support channel like where GoAFormStepStatusType is coming if people use basic IDE and there was no auto-completion. 

But when wrote this sample code, I was reluctant to remove the rest imports too. I'm open to removing those imports if we think it is not needed.

vanessatran-ddi: @ArakTaiRoth As discussed I updated our storybook. Thank you Dustin.

chrisolsen (approved):

PR #1218, opened 2023-06-14
feat: Add Tooltip component

amar-ABgov:


libs/web-components/src/components/tooltip/Tooltip.svelte
```diff
@@ -0,0 +1,270 @@
...
+  }
+
+  //Types
+  type Position = typeof Positions[number];
+  type Alignment = typeof Positions[number];
+
+  export let content = "";
+  export let testid: string = "";
+  export let position: Position = "top";
+  export let xalign: Alignment = "center"; // default to center
```
chrisolsen, line 35: To better match with other component attribute naming I think this should be renamed to `halign` for `horizontal` ex (hSpacing, vSpacing)

libs/web-components/src/components/tooltip/Tooltip.svelte
```diff
@@ -0,0 +1,270 @@
...
+    if (position === "left" || position === "right") {
+      if (align !== "center") {
+        console.error(`[${align}] is an invalid option for position ${position}`);
+      }
+    }
+  }
+
+  //Types
+  type Position = typeof Positions[number];
+  type Alignment = typeof Positions[number];
```
chrisolsen, line 30: This should probably be `Alignment[number]`?

Radha-Alberta:
- [ ] For Android and IOS, tooltip message doesn't display full message as shows in below image.


image

image

libs/web-components/src/components/tooltip/Tooltip.svelte
```diff
@@ -0,0 +1,280 @@
...
+    if (_rootEl && _tooltipEl) {
+      _rootEl.style.setProperty(
+        "--target-width",
+        `${_rootEl.getBoundingClientRect().width / 2}px`,
+      );
+    }
+  }
+  $: {
+    _initialPosition = position;
+  }
```
chrisolsen, line 69: Can this be moved into the `onMount`? From what I can tell this is a one time thing, so it probably doesn't need to be in a reactive block.

libs/web-components/src/components/tooltip/Tooltip.svelte
```diff
@@ -0,0 +1,298 @@
...
+  export let mb: Spacing = null;
+  export let ml: Spacing = null;
+
+  let _screenSize = 0;
+  let _rootEl: HTMLElement;
+  let _tooltipEl: HTMLElement;
+  let _tooltipWidth = 0;
+  let _initialPosition: Position;
+  let _tooltipVisible = false;
+  let _tooltipInstanceId: string;
```
chrisolsen, line 48: Is this needed? Since this is a web component everything resided in the shadow dom, there there will be no conflict with other tooltips on the page (assuming that was the reason the random was used)

Radha-Alberta:
- [ ]  there should be no gap between the arrow and the tooltip box. There should also be 4px of space between the tooltip and the icon button.
image
image

libs/web-components/src/components/tooltip/Tooltip.svelte
```diff
@@ -0,0 +1,322 @@
...
+  }
+  // call checkAndAdjustPosition function when content changes
+  $: {
+    content;
+    checkAndAdjustPosition();
+  }
+
+  const showTooltip = () => {
+    console.log("showTooltip", "onFocus");
+    // _showTooltipTimeout = setTimeout(() => {
```
chrisolsen, line 91: delete these lines and line 94

chrisolsen (changes requested):

libs/web-components/src/components/tooltip/Tooltip.svelte
chrisolsen: The `console.log` looks like it is still there. When you remove that there looks like there is some indentation issues that should be fixed.

chrisolsen (dismissed):

libs/docs/src/components/common/Tooltip.stories.mdx
```diff
@@ -0,0 +1,243 @@
...
+    `}
+    />
+  </Tab>
+</Tabs>
+
+#### Position
+
+<Story name="Position">
+  <div style={{ textAlign: "center" }}>
+    <GoATooltip content="This is a bttom tooltip" position="bottom">
```
ArakTaiRoth, line 117: You spelled `bottom` wrong here and every place below this.

libs/docs/src/components/common/Tooltip.stories.mdx
```diff
@@ -0,0 +1,243 @@
...
+      position="bottom"
+      hAlign="right"
+    >
+      <GoAIconButton icon="information-circle"></GoAIconButton>
+    </GoATooltip>
+    <GoATooltip content="This is a left tooltip" hAlign="left">
+      <GoAIconButton icon="information-circle"></GoAIconButton>
+    </GoATooltip>
+    <GoATooltip
+      content="This is a left tooltip"
```
ArakTaiRoth, line 188: To be consistent with the rest of the content, this should say "This is a left aligned bottom tooltip"

libs/web-components/src/components/tooltip/Tooltip.html-data.json
```diff
@@ -0,0 +1,85 @@
+{
+  "name": "goa-tooltip",
+  "attributes": [
+    {
+      "name": "heading",
```
ArakTaiRoth, line 5: I think `heading` and `secondaryText` needs to be removed, and `content` added.

ArakTaiRoth (approved):

chrisolsen (approved):

PR #1247, opened 2023-06-21
fix #1152: Removed wrapping of filename

### LINKED ISSUE #1152: Wrap file name for uploader component
**Current component**

* Truncates text. User cannot see the full file name. 
* Figure out what happens with long file names that have no space

----

## Acceptance criteria

**Current implementation where text is truncated**
image-20230502-211710

**Updated design**
image-20230608-165807

ArakTaiRoth:
Also updated a date file for Sidebar Group, as it was failing to compile because of a JSON error.

### DIFF
--- a/libs/web-components/src/components/file-upload-card/FileUploadCard.svelte
+++ b/libs/web-components/src/components/file-upload-card/FileUploadCard.svelte
@@ -176,9 +176,7 @@
   .filename {
     grid-area: filename;
     font-size: var(--goa-font-size-4);
-    overflow: hidden;
     overflow-wrap: anywhere;
-    text-overflow: ellipsis;
   }
 
   .error-msg {


libs/web-components/src/components/file-upload-card/FileUploadCard.svelte
```diff
@@ -175,10 +175,9 @@
 
   .filename {
     grid-area: filename;
-
     font-size: var(--goa-font-size-4);
-    white-space: nowrap;
     overflow: hidden;
```
chrisolsen, line 179: Can overflow and text-overflow also be removed?

ArakTaiRoth: They can, and have been.

chrisolsen (approved):

Issue #1136, opened 2023-06-12
Tabs


ArakTaiRoth:
## Acceptance criteria

* Tabs will ultimately need to be a component released as part of the design system in Angular and React.
* Consider using the tabs included in storybook as a starting point, reach out to @chrisolsen for more information
* Connect with @twjeffery for finalized designs 
* [WiP Design](https://www.figma.com/file/AEVq6WLepJsh4NtDK52B2t/Component---Tabs?type=design&node-id=1654-115716)

PR #1249, opened 2023-06-22
feat(1136): Add Tabs component

vanessatran-ddi:
**1. What this code change do:**

- Add Tabs and TabItem under web-components
- Add demo route for Tabs in Angular and React demo
- Add a new story book page for Tabs component
- Unit tests added 

**Screenshots**
image

MOBILE VERSION:

image

DESKTOP:

image

- **2. Make sure that you've checked the boxes below before you submit PR:**

- [x] I have read and followed the [contribution guidelines](../../contributing.md)

**3. Which issue and JIRA item does this PR fix (optional)**
#1136 
**4. Additional notes for the reviewer**
- Tabs component supports for accessibility guideline at [w3](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-manual/)

## Reviewer Checklist

- [ x] Appropriate unit test coverage.
- [x] [Coding Standards](../contribution_guidelines/coding_standards.md) followed.

### DIFF
--- a/libs/docs/src/components/common/Tabs.stories.mdx
+++ b/libs/docs/src/components/common/Tabs.stories.mdx
@@ -1,0 +1,225 @@
+import { Meta, Story } from "@storybook/addon-docs";
+
+import {
+  Tabs,
+  Tab,
+  SupportInfo,
+  Props,
+  Prop,
+  CodeSnippet,
+} from "@abgov/shared/storybook-common";
+import {
+  GoATabs,
+  GoATab,
+  GoABadge,
+  GoACallout,
+  GoATable,
+} from "@abgov/react-components";
+
+<Meta title="Components/Tabs" />
+
+# Tabs
+
+<GoACallout>
+  The tabs component lets users navigate between related sections of content,
+  displaying one section at a time.
+</GoACallout>
+
+[Figma design component](https://www.figma.com/file/AEVq6WLepJsh4NtDK52B2t/Component---Tabs?type=design&node-id=1654-115716)
+
+---
+
+#### GoATab Properties
+
+<Props showTabs={true}>
+  <Prop
+    name="heading"
+    type="slot"
+    defaultValue=""
+    description="Add components to the tab heading such as badges"
+  />
+  <Prop
+    name="open"
+    type="boolean"
+    defaultValue="false"
+    description="Sets the state of the tab open or closed"
+  />
+</Props>
+
+#### Basic
+
+<Story name="Basic">
+  <GoATabs>
+    <GoATab open heading="Profile">
+      <p>
+        <b>Profile:</b> Lorem ipsum dolor sit amet, consectetur adipiscing elit,
+        sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
+      </p>
+    </GoATab>
+    <GoATab
+      heading={
+        <>
+          Review pending <GoABadge type="important" content="2" />
+        </>
+      }
+    >
+      <p>
+        <b>Review pending:</b> Lorem ipsum dolor sit amet, consectetur
+        adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore
+        magna aliqua.
+      </p>
+    </GoATab>
+    <GoATab
+      heading={
+        <>
+          Completed <GoABadge type="midtone" content="1" />
+        </>
+      }
+    >
+      <p>
+        <b>Completed:</b> Lorem ipsum dolor sit amet, consectetur adipiscing
+        elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
+      </p>
+      <p>Ed do eiusmod tempor incididunt ut labore et dolore magna aliqua</p>
+    </GoATab>
+  </GoATabs>
+</Story>
+
+<Tabs>
+  <Tab label="Angular">
+    <CodeSnippet
+      code={`
+        <goa-tabs>
+          <goa-tab open="true" heading="Profile">
+             <p>
+               <b>Profile</b> Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut
+      labore et dolore magna aliqua.
+             </p>
+          </goa-tab>
+          <goa-tab>
+             <div slot="heading">
+                Review pending <goa-badge type="important" content="2"></goa-badge>
+             </div>
+              <p>
+                <b>Review pending:</b> Lorem ipsum dolor sit amet, consectetur adipiscing elit,
+                sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
+              </p>
+          </goa-tab>
+          <goa-tab>
+            <div slot="heading">
+               Completed
+               <goa-badge type="midtone" content="1"></goa-badge>
+            </div>
+            <p>
+             <b>Completed:</b> Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut
+        labore et dolore magna aliqua.</p>
+          </goa-tab>
+        </goa-tabs>
+      `}
+      lang="html"
+    />
+  </Tab>
+  <Tab label="React">
+    <CodeSnippet
+      code={`
+        <GoATabs>
+            <GoATab open heading="Profile">
+                <p>
+                    <b>Profile:</b> Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut
+            labore et dolore magna aliqua.
+                </p>
+            </GoATab>
+            <GoATab heading={<>Review pending <GoABadge type="important" content="1"/></>}>
+                <p>
+                    <b>Review pending:</b> Lorem ipsum dolor sit amet, consectetur adipiscing elit,
+                        sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
+                </p>
+            </GoATab>
+            <GoATab heading={<>Completed <GoABadge type="midtone" content="1"/></>}>
+                <p>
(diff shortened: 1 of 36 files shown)


libs/web-components/src/components/tab-item/TabItem.svelte
```diff
@@ -0,0 +1,137 @@
...
+        childindex: childindex
+      },
+      composed: true,
+      bubbles: true
+    }));
+  }
+</script>
+
+<!--HTML-->
+<button
```
vanessatran-ddi, line 42: For Reviewers: this follows the tutorial of accessibility at https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-manual/

Radha-Alberta:
**Keyboard Functions:**
- [ ] Right Arrow: If focus is on the last tab, does not moves focus to the first tab.
- [ ] Left Arrow: If focus is on the first tab, does not moves focus to the last tab.
image



**UI**
- [ ] Profile tabs padding should be 16px on both side (only right padding is 16px).
- [ ] Gap between 2 tabs should be 32px. (current gap 0px).
image
https://www.loom.com/share/f2835ca14b224134874146f91b514e02?sid=9e26c088-e5f0-4964-8350-b2e3a90d053f

**Accessibility:**
- [ ] Chrome OS on Browser: while hovering the tabs, it does not speak for 1st tab (Profile).


https://github.com/GovAlta/ui-components/assets/107069489/4c137883-c046-4aca-85e6-68c2eb3d279d

libs/react-components/src/lib/tab-item/tab-item.tsx
```diff
@@ -0,0 +1,30 @@
...
+interface WCProps {
+  open?: boolean;
+  heading?: string | React.ReactNode;
+}
+
+declare global {
+  // eslint-disable-next-line @typescript-eslint/no-namespace
+  namespace JSX {
+    interface IntrinsicElements {
+      "goa-tab-item": WCProps & React.HTMLAttributes<HTMLElement>;
```
chrisolsen, line 12: I find that `tab-item` seems to be redundant. I think `goa-tab` works better. We should have a discussion on this.

libs/web-components/src/components/tab-item/TabItem.svelte
```diff
@@ -0,0 +1,137 @@
...
+  bind:this={_rootEl}
+  id={`tab-${childindex}`}
+  type="button"
+  role="tab"
+  aria-selected="{_isOpen}"
+  aria-controls={`tabpanel-${childindex}`}
+  on:click={notifyParent}
+  on:click
```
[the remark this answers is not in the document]
chrisolsen: I don't think this is necessary and adds some confusion to why this is being done.

libs/web-components/src/components/tab-item/TabItem.svelte
```diff
@@ -0,0 +1,137 @@
...
+>
+  <slot name="heading" class="heading">{heading}</slot>
+</button>
+
+<style>
+
+  [role="tab"] {
+    display: block;
+    width: 100%;
+    position: relative;
```
chrisolsen, line 71: What is the reason this is `display: relative` with a `z-inde x`

libs/web-components/src/components/tab-item/TabItem.svelte
```diff
@@ -0,0 +1,137 @@
...
+    outline: none;
+    overflow: hidden;
+    cursor: pointer;
+    padding-top: var(--goa-space-xs);
+    padding-left: 15px; /* 16px - 1px border */
+    padding-bottom: var(--goa-space-s);
+    border: none;
+    font: var(--goa-typography-body-m);
+    color: var(--goa-color-text-default);
+    border-left: var(--goa-border-width-s) solid var(--goa-color-greyscale-200);
```
chrisolsen, line 84: I don't like having mobile specific styles in the "base" class, for the reason that after that you have to remember to override that style in the other class definitions for different media/container sizes.

chrisolsen (changes requested):

libs/web-components/src/components/tab-item/TabItem.svelte
vanessatran-ddi: It is a good catch. In my memory, I was doing that to make sure the tab is on the top of the stack to be visible. However now when I re-test everything without z-index it looks good still. I will remove it.

libs/web-components/src/components/tab-item/TabItem.svelte
vanessatran-ddi: I was doing that with the purpose to make it mobile first. It just happened that our component has different padding sizes for mobile and desktop. 
As discussed in slack, we can sync after your touch base with Thom regarding it.

libs/web-components/src/components/tab-item/TabItem.svelte
vanessatran-ddi: I think I remember why now. 
image
I need to make z-index here just to make sure when it is hover, active or focus, the border px width is higher, which is kind of hidden by the tabpanel border width 1px below. 

Without z-index it will be: 
image

The focus outline is hidden by the tabpanel border below.

libs/web-components/src/components/tab-item/TabItem.svelte
vanessatran-ddi: I updated the CSS and separate mobile/desktop CSS, with a base CSS like we discussed.

libs/react-components/src/lib/tab-item/tab-item.tsx
vanessatran-ddi: I changed the name from "goa-tab-item" to "goa-tab".

vanessatran-ddi:
> **Keyboard Functions:**
> 
> * [x]  Right Arrow: If focus is on the last tab, does not moves focus to the first tab.
> * [x]  Left Arrow: If focus is on the first tab, does not moves focus to the last tab.
>   image
> 
> **UI**
> 
> * [x]  Profile tabs padding should be 16px on both side (only right padding is 16px).
> * [x]  Gap between 2 tabs should be 32px. (current gap 0px).
> * [x]  the bottom border for hover and active should be 4px.
> * [x]  the focus state should also include the grey bottom border like the hover state.
> * [x]  currently the position of the active and hover bottom border is correctly positioned on the first tab, and is lower on all other tabs
> * [x]  there is extra 16px padding on the left side of all tabs, making the tab content off centre
>   - it should be space-m for padding left and right within the tab
>   image
>   https://www.loom.com/share/f2835ca14b224134874146f91b514e02?sid=9e26c088-e5f0-4964-8350-b2e3a90d053f
> 
> **Accessibility:**
> 
> * [x]  Chrome OS on Browser: while hovering the tabs, it does not speak for 1st tab (Profile).
> 
>  Tabs_Accessibility_Windows_Chrome.mp4

I fixed the above. The last point "Accessibility", we agree that NVDA should speak when we focus on the tab (not hover)

Radha-Alberta:
- [ ] active focus always does have blue underline, not grey

- [ ] It looks like the tab order should be:
            - focus active tab
            - can use arrow keys left and right to change tab
            - but next tab key goes to content
            - then next tab key goes to next part of page
https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-automatic/


image
image

libs/docs/src/components/common/Tabs.stories.mdx
```diff
@@ -0,0 +1,275 @@
...
+<GoACallout>
+  The tabs component lets users navigate between related sections of content,
+  displaying one section at a time.
+</GoACallout>
+
+[Figma design component](https://www.figma.com/file/AEVq6WLepJsh4NtDK52B2t/Component---Tabs?type=design&node-id=1654-115716)
+
+---
+
+#### GoATabItem Properties
```
ArakTaiRoth, line 32: I think we changed the name to just `GoATab` instead of `GoATabItem`

libs/docs/src/components/common/Tabs.stories.mdx
```diff
@@ -0,0 +1,275 @@
...
+      </p>
+    </GoATab>
+    <GoATab
+      heading={
+        <>
+          Review pending <GoABadge type="important" content="2" />
+        </>
+      }
+    >
+      <GoATable variant="relaxed">
```
ArakTaiRoth, line 66: Is there a specific reason we decided to show a table in this tab? If not, I would suggest changing the content to just a paragraph or something, it would make the code people are copying soooo much simpler.

libs/react-components/src/lib/tab/tab.tsx
```diff
@@ -0,0 +1,29 @@
...
+  namespace JSX {
+    interface IntrinsicElements {
+      "goa-tab": WCProps & React.HTMLAttributes<HTMLElement>;
+    }
+  }
+}
+
+export interface TabItemProps {
+  open?: boolean;
+  heading?: string | React.ReactNode;
```
ArakTaiRoth, line 19: Because we're using `children` below, I'm pretty sure we need to add `children` to the properties as well, otherwise we have problems with more modern versions of React. Refer to your `tabs` component, as that does it properly.

libs/web-components/src/components/tab/Tab.html-data.json
```diff
@@ -0,0 +1,27 @@
...
+      "description": "Add components to the tab heading such as badges"
+    },
+    {
+      "name": "open",
+      "description": "Sets the state of the tab open or closed",
+      "valueSet": "boolean",
+      "default": "false"
+    },
+    {
+      "name": "content",
```
ArakTaiRoth, line 16: This isn't accurate. There is no content `slot`, you just put content in between the opening and closing tag.

ArakTaiRoth (dismissed):

chrisolsen (commented):
In this review there are a few things that I saw, but after looking at things a little more I think things could be simplified a bit.

PR #1255, opened 2023-06-28
fix #1238 accordion supports children property in react 18

### LINKED ISSUE #1238: Accordion doesn't support children property
## Issue:

If you use the `GoAAccordion` (React issue), you get an error that property `children` does not exist. This is with the latest version of React.
image-20230609-212557

## Replication Steps:

1. Use any of our existing Accordion code, but use it in a React 18 App

## Acceptance Criteria:

1. `GoAAccordion` works correctly in the latest version of React
2. Should probably test the rest of our components as well. As this has been a re-occurring problem for a few of our components.

Radha-Alberta:


ArakTaiRoth (approved):

chrisolsen:
This commit message will not close/move the ticket.

We also no longer need the fix(DDI-999) any more, seeing as that was a way to link to the Jira ticket.

So the commit message should be multi-line like:
```
fix: blah blah blah

Closes #999
```

chrisolsen (approved):

Issue #1145, opened 2023-06-12
Add onFocus and onBlur events to inputs


ArakTaiRoth:
## Acceptance criteria:

* onFocus event is added to all input style components
* onBlur event is added to all input style components

PR #1262, opened 2023-07-05
fix(#1145): added onfocus and onblur event

Radha-Alberta:


libs/web-components/src/components/input/Input.svelte
```diff
@@ -102,6 +102,13 @@
     value = input.value;
   }
 
+  function onFocus(){
+    this.dispatchEvent(new CustomEvent("_onFocus", { composed: true }));
```
chrisolsen, line 106: Following our event naming convention, custom event should be `_focus` and `_onBlur` should be `_blur`

ArakTaiRoth:
Issues:

- [ ] React - Both focus and blur events fire twice when activated
- [ ] React and Angular - There are no event details sent back when activating the event like the change event. Send back the name of the component, and the value in the component, just like the change event.

libs/web-components/src/components/input/Input.svelte
```diff
@@ -102,6 +102,34 @@
...
+    if (_debounceId != null) {
+      clearTimeout(_debounceId);
+    }
+    _debounceId = setTimeout(() => {
+      input.dispatchEvent(new CustomEvent("_blur", {
+        composed: true,
+        bubbles: false,
+        cancelable: true,
+        detail: { name, value: input.value } }));
+    }, debounce);
```
chrisolsen, line 130: Is there a reason you are using a debounce here? Debounce is something you would use when you have continuous input, but focus and blur doesn't happen as often.

chrisolsen (changes requested):

libs/web-components/src/components/input/Input.svelte
Radha-Alberta: while clean up, I missed this one

libs/docs/src/components/common/Input.stories.mdx
```diff
@@ -228,6 +228,30 @@ import { Meta, Story } from "@storybook/addon-docs";
     type="() => void"
     description="onclick function invoked when trailing icon is clicked"
   />
+  <Prop
+    name="onFocus"
+    lang="react"
+    type="() => void"
```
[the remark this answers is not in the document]
Radha-Alberta: I updated the code, However this will remain same for angular as we're just using '$event' to get event.detail [ just like onChange ]

libs/react-components/src/lib/input/input.tsx
```diff
@@ -355,6 +378,8 @@ export const GoAInputNumber: FC<NumberInputProps> = ({
       min={min?.toString()}
       max={max?.toString()}
       value={value.toString()}
+      onFocus={onNumberChange}
+      onBlur={onNumberChange}
```
chrisolsen, line 382: You will need to use the onFocus and onBlur here

chrisolsen (changes requested):

libs/react-components/src/lib/input/input.tsx
```diff
@@ -94,21 +94,28 @@ interface BaseProps extends Margins {
 }
 
 type OnChange = (name: string, value: string) => void;
+type onFocusBlurEvent = (name: string, value: string) => void;
```
chrisolsen, line 97: This event should not include two different events. Delete this and use the blur and focus event you created below.

libs/react-components/src/lib/input/input.tsx
```diff
@@ -94,21 +94,28 @@ interface BaseProps extends Margins {
...
   min?: number | string;
   max?: number | string;
   step?: number;
+  onFocus?: onFocusBlurEvent;
+  onBlur?: onFocusBlurEvent;
 }
 
 type OnNumberChange = (name: string, value: number) => void;
+type OnFocusChange = (name: string, value: number) => void;
+type OnBlurChange = (name: string, value: number) => void;
```
chrisolsen, line 110: These should just be named OnFocus and OnBlur, since nothing is changing.

libs/react-components/src/lib/input/input.tsx
```diff
@@ -348,13 +373,21 @@ export const GoAInputNumber: FC<NumberInputProps> = ({
   const onNumberChange = (name: string, value: string) => {
     props.onChange(name, parseInt(value));
   };
+  const onFocusChange = (name: string, value: string) => {
+    props.onFocus?.(name, parseInt(value));
+  };
+  const onBlurChange = (name: string, value: string) => {
+    props.onBlur?.(name, parseInt(value));
+  };
```
chrisolsen, line 381: Remove "Change"

chrisolsen (changes requested):

libs/react-components/src/lib/input/input.tsx
Radha-Alberta: I created onFocus and onBlur events, However, we can't use below event as the value type is different for below onFocusType.

libs/react-components/src/lib/input/input.tsx
```diff
@@ -167,13 +181,27 @@ export const GoAInput: FC<InputProps & { type?: GoAInputType }> = ({
       onTrailingIconClick?.();
     };
 
+    const onFocusListener = (e: unknown) => {
```
chrisolsen, line 184: For consistency within this file this should be `focusListener` and onBlurListener should be `blurListener`

chrisolsen (dismissed):

libs/docs/src/components/common/Input.stories.mdx
```diff
@@ -228,6 +228,30 @@ import { Meta, Story } from "@storybook/addon-docs";
     type="() => void"
     description="onclick function invoked when trailing icon is clicked"
   />
+  <Prop
+    name="onFocus"
+    lang="react"
+    type="(name: string, value: string | Date | number) => void"
+    description="onFocus function triggers when an element receives focus"
```
chrisolsen, line 235: This should be changed to
`Function invoked when an element receives focus`

The same is true for the onBlur prop

chrisolsen (approved):

ArakTaiRoth (approved):

PR #1263, opened 2023-07-05
use popover in dropdown

chrisolsen:


Radha-Alberta:
- [ ] Dropdown is not working as expected in storybook, angular and react

https://www.loom.com/share/0791f1512371459e8ac64c86d53f02a8

https://www.loom.com/share/c433d267846c48b798324ae1051dba34

libs/web-components/src/components/popover/Popover.svelte
```diff
@@ -123,7 +125,8 @@
 
   // Opens the popover and adds the required binding to the new slot element
   function openPopover() {  
-    (async () => {
+    if (_disabled) return;
+    ;(async () => {
```
vanessatran-ddi, line 129: Is it an extra ; here?

chrisolsen: It was on purpose

vanessatran-ddi: Thank you Chris! I just read about semicolon before() on google. Good to know!

vanessatran-ddi (approved):

ArakTaiRoth (approved):

Issue #1266, opened 2023-07-06
[Bug]: Dropdown component does not respect the [value] binding when navigating with the keyboard


mtycholaz:
### Describe the bug

Angular 13+
@abgov/angular-components: 2.2.0
@abgov/web-components: 1.11.0

When binding an object to a dropdown. If you use keyboard navigation, the label gets set as the value instead of the actual value. Also, keyboard navigation is just simply broken, ie. can't use up and down arrows to change selection.

### Provide Code
```
monthCtrl = new FormControl();

months = [
    { number: 1, name: 'January' },
    { number: 2, name: 'February' },
    { number: 3, name: 'March' },
    { number: 4, name: 'April' },
    { number: 5, name: 'May' },
    { number: 6, name: 'June' },
    { number: 7, name: 'July' },
    { number: 8, name: 'August' },
    { number: 9, name: 'September' },
    { number: 10, name: 'October' },
    { number: 11, name: 'November' },
    { number: 12, name: 'December' }
  ]

{{ monthCtrl.value | json }}
<br />
<goa-dropdown goaValue name="selectMonth" [formControl]="monthCtrl">
  <goa-dropdown-item
    *ngFor="let month of months"
    [value]="month.number"
    [label]="month.name"
  >
  </goa-dropdown-item>
</goa-dropdown>
```

### To Reproduce

1. Use the above code in the Angular demo.
2. Select the dropdown with a mouse,
3. Select an option. Notice that the displayed value is the actual `[value]` property
4. Hit Tab, it should auto select the same dropdown. Notice that the displayed value is now the `[label]` property - **First Problem**
5. Hit space to open the dropdown
6. Try to navigate using the arrow keys, you can't - **Second Problem**

PR #1273, opened 2023-07-12
fix(#1266): prevent incorrect values from being emitted by dropdown

chrisolsen:


chrisolsen:
> Seems to fix the first issue mentioned on the bug ticket The second issues where user is unable to select from the dropdown using keyboard still exists. Is that expected?

This fix is only for the first issue. They keyboard fix can be found in #1279

Issue #1244, opened 2023-06-20
[Bug]: Toggling the `native` property on a dropdown results in null errors


chrisolsen:
### Describe the bug

Toggling the `native` property on a dropdown results in null errors
```
Uncaught TypeError: _rootEL is undefined
```

### Provide Code

```
<GoADropdown name="colors" value="red" onChange={()=>{}} native={true}>
  <GoADropdownItem value="red" label="Red" />
  <GoADropdownItem value="green" label="Green" />
  <GoADropdownItem value="blue" label="Blue" />
</GoADropdown>
```

### To Reproduce

1. Use the existing deployment of the code sandbox
2. Go to the Dropdown  Component page
3. Toggle `native` on and off, you should see the error come up in console

### What browsers are you seeing the problem on?

Chrome
Firefox

ArakTaiRoth:
Complete

PR #1282, opened 2023-07-17
fix(#1244): prevent error when toggling native dropdown prop

chrisolsen:


Radha-Alberta:
- [ ] Error in console while toggling checkbox in code Playground.

image

ArakTaiRoth (approved):

Issue #1285, opened 2023-07-19
[Bug]: click events not working on table rows inside GoATable


ArakTaiRoth:
### Describe the bug

If you use `GoATable` in React. And attach an `onClick` event to a table row inside the table, the click event will not fire.

### Provide Code

```javascript
function consoleLog() {
  console.log("I'm in the table row");
}
```

```javasacript
<GoATable width="100%">
  <thead>
    <tr>
      <th>First Name</th>
      <th>Last Name</th>
      <th>Age</th>
    </tr>
  </thead>
  <tbody>
    <tr onClick={() => consoleLog()}>
      <td>Dustin</td>
      <td>Nielsen</td>
      <td>37</td>
    </tr>
  </tbody>
  <tfoot>
    <tr>
      <td>This is the footer</td>
      <td>This is the footer</td>
      <td>This is the footer</td>
    </tr>
  </tfoot>
</GoATable>
```

### To Reproduce

Use the above code, click the row, a console message should show up. It's not.

### Additional context

From @chrisolsen "It looks like the table is swallowing all the events within it, including clicks".

PR #1288, opened 2023-07-21
fix(#1285): allow non-goa html element to emit events from within goa…

chrisolsen:
…-table


Radha-Alberta:
- onClick event is working as expected.

- [ ] There's still a error message on console that needs to be resolved.

image

Issue #1252, opened 2023-06-27
Section title for Side menu


Spark450:
## Acceptance Criteria:

- "Section title" is an available "menu item" type
- Section titles are non-interactive and help to organize menu items into sections
- [Figma design](https://www.figma.com/file/k9ZHctiEnN74GEhGWpio3K/Component---Side-Menu?type=design&node-id=2238%3A9253&t=ep61tqt2q9V9pa6U-1)
- Screen reader should announce the section heading once when the user navigates a set of child links.

### Accessibility

- Section titles **could** be implemented with styled `<h>` tags eg. `<h1>`. This should provide the accessibility we need for a section header. Note: this is one possible approach to ensure proper accessibility. Other options may be explored if they provide a better experience.
<img width="489" alt="Screen Shot 2023-06-27 at 12 16 17 PM" src="https://github.com/GovAlta/ui-components/assets/51723535/78d697a2-eafa-456b-a6c6-fff836a81ae4">

PR #1290, opened 2023-07-25
feat(#1252): Section title for Side menu

amar-ABgov:


Radha-Alberta:
- [ ] padding should be 24px on the left as shown in figma design.

image
image

libs/web-components/src/components/side-menu-section-title/SideMenuGroup.html-data.json
```diff
@@ -0,0 +1,4 @@
+{
+  "name": "goa-side-menu-section-title",
+  "attributes": []
+}
```
ArakTaiRoth, line 4: You're missing the `references` section, with the name of the component and the URL to it's position.

ArakTaiRoth:
Would you be able to add usage of this section menu title to the angular and react demo's as well?

ArakTaiRoth (dismissed):

Radha-Alberta:
Tom Feedback,
- [ ] The heading is a little lower than expected. Image attached
- [ ] can we include examples without the icon or badge as well? Most of the time it would be just the heading.

image (3)

libs/web-components/src/components/side-menu-heading/SideMenuHeading.svelte
```diff
@@ -0,0 +1,45 @@
...
+          console.error('goa-side-menu-heading slot should only contain text nodes.');
+        }
+      });
+    });
+  });
+</script>
+
+<style>
+ h2 {
+  border-top: 2px solid var(--goa-color-greyscale-200, #DCDCDC);
```
thytran142, line 24: Question: Is there a reason why we want to add #dcdcdc here instead of using token only?

libs/web-components/src/components/side-menu-heading/SideMenuHeading.svelte
```diff
@@ -0,0 +1,45 @@
...
+  background: var(--goa-color-greyscale-100, #F1F1F1);
+  padding: 0.75rem 0.75rem 0.5rem 1.5rem;
```
vanessatran-ddi, line 26: Question: Should we use --goa-space-s for example, for 0.75rem?

libs/web-components/src/components/side-menu-heading/SideMenuHeading.svelte
chrisolsen: The `2px` can be replaced with `--goa-border-width-m`

libs/web-components/src/components/side-menu-heading/SideMenuHeading.svelte
```diff
@@ -0,0 +1,45 @@
...
+    slotEls.forEach(slotEl => {
+      slotEl.childNodes.forEach((node:HTMLElement) => {
+        // Exclude nodes that belong to the 'meta' slot
+        if (node.nodeType !== Node.TEXT_NODE && (!node.slot || node.slot !== 'meta')) {
+          console.error('goa-side-menu-heading slot should only contain text nodes.');
+        }
+      });
+    });
+  });
+</script>
```
chrisolsen, line 20: I don't think this is required. There should be a description in the docs on how we recommend things should be done and I think that would be sufficient.

chrisolsen (approved):

vanessatran-ddi (approved):

PR #1300, opened 2023-07-27
fix(#1285): prevent html table events from being lost when using goa-…

chrisolsen:
…table

### DIFF
--- a/libs/web-components/src/components/table/Table.svelte
+++ b/libs/web-components/src/components/table/Table.svelte
@@ -26,6 +26,7 @@
 
   // Private
   let _rootEl: HTMLElement;
+  let _isTableRoot: boolean = false;
   $: _stickyHeader = toBoolean(stickyheader);
 
   onMount(() => {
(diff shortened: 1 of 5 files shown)


ArakTaiRoth (dismissed):

libs/web-components/src/components/table/Table.svelte
```diff
@@ -90,80 +91,31 @@
...
-    <tbody><tr><td /></tr></tbody>
-    <tfoot><tr><td /></tr></tfoot>
-  </template>
-</table>
+  `}
+>
+  {#if _isTableRoot}
+    <slot />  
+  {:else}
+    <table style="width: 100%;">
```
vanessatran-ddi, line 106: It seems that the style here override the "variant" attribute of a table.

chrisolsen: Can you better explain what is being overridden? The `variant` is `relaxed` and `sticky` and neither of those classes have a `width` style attribute.

vanessatran-ddi: Yes I tried on the storybook, looks like the table width is bigger. 
image

I see the current storybook, the table seems to have class name "normal"

libs/web-components/src/components/table/Table.svelte
```diff
@@ -90,80 +91,35 @@
...
-  </template>
-</table>
+  `}
+>
+  {#if _isTableRoot}
+    <slot />
+  {:else}
+    <table
+      style={
+       width && "width: 100%;"
```
vanessatran-ddi, line 108: 👍

vanessatran-ddi (approved):

ArakTaiRoth (approved):

PR #1309, opened 2023-08-08
feat: Merge combobox and dropdown

### LINKED ISSUE #1202: Combobox
## Acceptance criteria:

* If there are characters input or a selection is made there will be an “x” present next to the chevron to clear the selection
* Text input matching will have “bold” treatment
  * Allow matches in the middle of a result, but prioritize matches at the start of a result.
* Clear an input when focus is lost if no match is selected.
* When no match is found, state “No matches found” in pop-up, not selectable
* Should use the popover component
* Implemented in accordance with [design docs](https://www.figma.com/file/3pb2IK8s2QUqWieH79KdN7/%E2%9D%96-Components-%7C-D3?node-id=8360%3A264334&t=AmJR2Xo1EKwWF5CA-1)
* [Component API](https://www.figma.com/file/3pb2IK8s2QUqWieH79KdN7/%E2%9D%96-Components-%7C-D3?node-id=10842%3A385431&t=AmJR2Xo1EKwWF5CA-1)

## Accessibility

* Consider applying a role of “combobox”
* Consider the following attributes:
  * `aria-expanded`
  * `aria-haspopup`
  * `aria-controls`
  * `aria-activedescendant`
  * `aria-autocomplete`
* If initial state of the combo box is collapsed `aria-expanded=false`
* When the combo box is expanded `aria-expanded=true`
* If the menu is a `listbox` then the `aria-haspopup` value *does not* need to be set
* When the menu is displayed ensure that `aria-controls` attribute is set to the id of the menu `listbox`
* Chevron icon that shows menu should <type> button with a `tabindex` of -1.
  * It also must not be a descendent of the element with a role of `combobox`
* Code must be written to enable the user to move focus between the `combobox` input field and the elements in the listbox.
  * `down arrow` should move the focus from the input to the elements in the listbox.
* 

## Keyboard interactions

* `Down Arrow`
  * Moves focus to the next option, or to the first option if none was selected.
* `Alt + Down Arrow` (Optional)
  *If the popup is available but not displayed, displays the popup without moving focus.
* `Up Arrow`
  * Moves focus to the previous option. Moving focus to the first option if focus was originally on the last option.
* `Alt + Up Arrow` (Optional)
  * If the popup has focus, returns focus to the combobox, otherwise it closes the popup.
* `Enter`
  * If the combobox is editable and an autocomplete suggestion is selected in the popup, accepts the suggestion either by placing the input cursor at the end of the accepted value in the combobox or by performing a default action on the value. For example, in a messaging application, the default action may be to add the accepted value to a list of message recipients and then clear the combobox so the user can add another recipient.

## Additional Notes:

1. Look into using the native Autocomplete feature
  * see how that behaves with respect to location on the screen (including popover positioning behaviour)
  * Question - Can scroll be disabled on open with the native dropdown?
  * Question - Can the list be styled?
  * Question - can a “Clear selection” control be added in the input? 
2. Implement for use with a list of data, and a callback function that updates the list.

# Properties

| Name | Type | Required | Default | Description |
| :-- | :-- | :-- | :-- | :-: |
| `name` | `string` | No | | Must be unique in the rendered view and must match the children's name |
| `value` | `string` | No | | Stores the value of the item selected from the dropdown |
| `maxHeight` | `string` | No | 276px | Maximum height of the dropdown menu items popover. |
| `placeholder` | `string` | No | | The text displayed for the dropdown before a selection is made. |
| `default-value` | `string` | No | | The combo box will set this value as the default selection if it is found within the select options. |
| `width` | `string` | No | | Override the autosized menu width |
| `disabled` | `boolean` | No | false | Disable this control |
| `error` | `boolean` | No | false | Show an error state |
| `ariaLabel` | `string` | No | | Defines how the selected value will be translated for the screen reader. If not specified it will fall back to the name |

# Design

[Figma](https://www.figma.com/file/3pb2IK8s2QUqWieH79KdN7/%E2%9D%96-Components-%7C-D3?node-id=8360%3A264334&t=x7K7u0TneGG5q4Tg-1)

vanessatran-ddi:
# Description

Merge Combobox and Dropdown together. The dropdown has a new attribute "filterable" that allows users to query options. 

Fixes #1202 

## Type of change

- [x] Bug fix for current dropdown (keyboard behaviors & width)
- [x] New feature (Combobox)
- [x] Storybook updated 
- [x] Add arialabelledby for "input" and "dropdown"
- [x] Add "id" for "form-item"

# How Has This Been Tested?

image

image

image


libs/docs/src/components/common/Combobox.stories.mdx
```diff
@@ -0,0 +1,535 @@
...
+    description="Disable this control. **Use [attr.disabled] with [formControl]**"
+  />
+  <Prop
+    name="disabled"
+    lang="react"
+    type="boolean"
+    defaultValue="false"
+    description="Disable this control"
+  />
+  <Prop
```
vanessatran-ddi, line 84: Note:
I must add id (and fall back to name if id is not provided), so if anyone uses it with our goa-form-item which has label, the screen reader will read/voice out correctly.
The requirement is from Combobox Figma / Accessibility section.

libs/docs/src/components/common/Dropdown.stories.mdx
```diff
@@ -89,6 +89,12 @@ import { useState } from "react";
     type="string"
     description="Defines how the selected value will be translated for the screen reader. If not specified it will fall back to the name"
   />
+  <Prop
```
vanessatran-ddi, line 92: Note:
Same with combobox, I think we should add the id for the dropdown, that will help the screen reader voice out the form id label. 
Source: [w3](https://www.w3.org/WAI/tutorials/forms/labels/)

libs/docs/src/components/common/FormItem.stories.mdx
```diff
@@ -53,6 +61,18 @@ import {
     type="optional | required"
     description="Marks the field with optional/required label."
   />
+  <Prop
```
vanessatran-ddi, line 64: Note:
One of the requirements of the Combobox is to make sure when the screen reader reads the form label accordingly if hover the combobox. 
To make it works, I need to add a new attribute here so label of the goa-form-item knows what field id it should be for label "for" attribute.
Source: https://www.w3.org/WAI/tutorials/forms/labels/

libs/docs/src/components/common/Input.stories.mdx
```diff
@@ -148,11 +148,24 @@ import { Meta, Story } from "@storybook/addon-docs";
...
+    type="string"
+    description="Defines how the input will be translated for the screen reader. If not specified it will fall back to the name"
+  />
   <Prop
     name="ariaLabel"
+    lang="react"
     type="string"
     description="Defines how the input will be translated for the screen reader. If not specified it will fall back to the name"
   />
+  <Prop
```
vanessatran-ddi, line 163: Note:
One of the requirement for Combobox. Since I've added the attribute "id" for "dropdown" & "combobox", It will be good to add to our goa-input too. 
Source: https://www.w3.org/WAI/tutorials/forms/labels/

libs/web-components/src/components/popover/Popover.svelte
```diff
@@ -254,13 +254,13 @@
...
     <div class="popover-container">
-      <section 
+      <section
         bind:this={_popoverEl}
         data-testid="popover-content"
         class="popover-content"
         style={`
           ${cssVar("width", width)}
-          max-width: ${maxwidth};
+          max-width: ${maxwidth < width ? width : maxwidth};
```
vanessatran-ddi, line 263: Note:
Fix the error that it has now for the dropdown too. 
image

Radha-Alberta:
- [ ] focus should be active after selecting the option.
- [ ] unable to type in combo-box when NVDA is active.
- [ ] left padding on dropdown should be 12px, currently 8px. (it should use --goa-spacing-s)

vanessatran-ddi:
> * [x]  focus should be active after selecting the option.
> * [ ]  unable to type in combo-box when NVDA is active.
> * [ ]  left padding on dropdown should be 12px, currently 8px. (it should use --goa-spacing-s)

I tested NVDA on windows 11, looks like it is working. Maybe you have something set up on NVDA different from mine? 

https://github.com/GovAlta/ui-components/assets/120135417/328f6aa8-49a0-43d6-972c-d6dee69ba3af

Radha-Alberta:
> > * [x]   focus should be active after selecting the option.
> > * [ ]   unable to type in combo-box when NVDA is active.
> > * [ ]   left padding on dropdown should be 12px, currently 8px. (it should use --goa-spacing-s)
> 
> I tested NVDA on windows 11, looks like it is working. Maybe you have something set up on NVDA different from mine?
> 
>  combobox-nvda.mov

- [ ] **Issue-1:** Requires 2 'esc' to remove the content from input box
- [ ] **Issue-2:** After removing content with esc key, unable to type in input box

Steps to Reproduce:
1. Turn on NVDA.
2. type 'red' and select 'RED' from the suggestion box.
3. press 'esc' key and remove the content [ **Issue-1** ]
4. type anything in input box [ **Issue-2** ]

vanessatran-ddi:
> 

Hi @Radha-Alberta 
I can re-produce the above issue with the standard example of combobox from w3:

https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/
See the below video:


https://github.com/GovAlta/ui-components/assets/120135417/1230994c-4bd9-49f5-8d98-e4c9c44b3bfd

I checked the below issue from Microsoft: https://learn.microsoft.com/en-us/answers/questions/1289668/web-client-escape-key-intercepted-by-web-browser-i
The Escape key is misled by Chrome browser (and Edge too). I will suggest to close this comment as we cannot fix it for now.

libs/web-components/src/components/dropdown/Dropdown.svelte
```diff
@@ -147,236 +217,283 @@
...
   function closeMenu() {
-    _menuEl.removeEventListener("blur", closeMenu);
-    _menuEl.removeEventListener("mouseover", onHighlight);
-    setHighlightedIndexToSelected();
     _isMenuVisible = false;
   }
 
-  function setHighlightedIndexToSelected() {
-    _highlightedIndex = _options.findIndex(option => _values.includes(option.value));
+  function isMatched(option: Option, query: string) {
```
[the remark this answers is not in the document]
vanessatran-ddi: Hi @amar-ABgov I added logic to make sure isMatched isn't called if _filterable is false. 
For the 'debounce' suggestion, I make some research and it seems to be falling the case that needs to fetch data from backend which we must deal with network latency, so `autocomplete` with fetching data from API according to the query, will be the perfect case for adding 300ms delay. 
For our case, since we use local data, so debounce may not be necessary.

libs/web-components/src/components/dropdown/Dropdown.svelte
```diff
@@ -129,14 +175,43 @@
...
+   */
+  function changeHighlightedOption(position: number) {
+    let index = _highlightedIndex + position;
+    let items = !_filteredOptions?.length ? _options : _filteredOptions;
+    if (items.length === 0) return;
+    if (index < 0) {
+      index = items.length - 1; // Back to the last item
+    } else if (index >= items.length) {
+      index = 0; // back to the top item
+    }
```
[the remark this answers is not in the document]
vanessatran-ddi: It is a good catch. I will modify it.

Issue #1284, opened 2023-07-18
[Bug]: goa-button with 'type="submit"' does not render a button of 'type="submit"'


tchan101:
### Describe the bug

Using `<goa-button type="submit">` does not render a `<button type="submit">` in the shadow-root. It is a regular button. When using this type of goa-button with a form, the form doesn't get submitted when the button is clicked. 

### Provide Code

```javascript
import 'zone.js/dist/zone';
import { Component, CUSTOM_ELEMENTS_SCHEMA } from '@angular/core';
import { CommonModule } from '@angular/common';
import { bootstrapApplication, BrowserModule } from '@angular/platform-browser';
import { AngularComponentsModule } from '@abgov/angular-components';
import '@abgov/web-components';
import { FormsModule } from '@angular/forms';

@Component({
  selector: 'my-app',
  standalone: true,
  imports: [CommonModule, AngularComponentsModule, FormsModule],
  schemas: [CUSTOM_ELEMENTS_SCHEMA],
  template: `
    <h1>Test Form</h1>
    <form #formData="ngForm" (ngSubmit)="onSubmit(formData.value)">
      <div class="form-group">
        <div for="name">Name</div>
        <input type="text" class="form-control" id="name" name="name" ngModel>
      </div>
      <goa-button
        type="submit"
      >Submit</goa-button>
      <!-- <button type="submit">Submit</button> -->
    <form>
  `,
})
export class App {
  onSubmit(data: any) {
    alert(`Hello ${data.name}`);
  }
}

bootstrapApplication(App);
```

### To Reproduce

1. Fill out the form.
2. Click the submit button. 

**Expected result:** 
An alert that shows "Hello <name>", where name is the value you entered in the form. 

**Actual result:**
Nothing happens

### Additional context

I also included a basic submit button that can be switched out. Clicking on it gives the expected result.

PR #1314, opened 2023-08-10
chore(#1284): Update docs about button type=submit

amar-ABgov:
There is no easy fix for this. 
For now, 
* making sure that the native button gets the `type=submit` prop
* updating the docs

### DIFF
--- a/libs/web-components/src/components/button/Button.svelte
+++ b/libs/web-components/src/components/button/Button.svelte
@@ -64,6 +64,7 @@
   disabled={isDisabled}
   on:click={clickHandler}
   data-testid={testid}
+  type={type == "submit" ? type : "button"}
 >
   {#if type === "start"}
     <span class="text">
--- a/libs/docs/src/components/common/Button.stories.mdx
+++ b/libs/docs/src/components/common/Button.stories.mdx
@@ -7,7 +7,7 @@ import {
   Props,
   Prop,
 } from "@abgov/shared/storybook-common";
-import { GoAButton, GoAButtonGroup } from "@abgov/react-components";
+import { GoAButton, GoAButtonGroup, GoACallout } from "@abgov/react-components";
 
 <Meta title="Components/Button" />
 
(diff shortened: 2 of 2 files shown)


ArakTaiRoth (approved):

libs/web-components/src/components/button/Button.svelte
```diff
@@ -64,6 +64,7 @@
   disabled={isDisabled}
   on:click={clickHandler}
   data-testid={testid}
+  type={type == "submit" ? type : null}
```
vanessatran-ddi, line 67: Question: Should we include "reset" type too? If "submit" is one type that we consider to render, then we shouldn't ignore "reset".
https://www.w3schools.com/tags/att_button_type.asp

I suggest making "type" default as "button" instead of null, otherwise, we allow "reset" or "submit" to be rendered.

vanessatran-ddi (commented):
I leave 1 question.

libs/web-components/src/components/button/Button.svelte
ArakTaiRoth: @vanessatran-ddi This is outside the scope of this story, so not something we're going to do. But it is being discussed what to do about button types, as you've noticed it is currently inconsistent and doesn't make a ton of sense.

ArakTaiRoth: @vanessatran-ddi @amar-ABgov This can be referenced in this story #1321

vanessatran-ddi: Thank you @ArakTaiRoth 
Yah I see what you mean, the "submit" type should be the only one to be taken care of. 
I am a bit reluctant with type = null as default, because according to MDN (https://html.spec.whatwg.org/multipage/form-elements.html#the-button-element) if type = null, the button will be rendered and behaved as "submit", so I think we should make type = "button" as default. 
(I haven't tested the PR so I am not sure why it doesn't work in Angular submit form as described in the issue)

vanessatran-ddi: I will resolve this conversation.

vanessatran-ddi (approved):

ArakTaiRoth (approved):

Issue #1327, opened 2023-08-18
[Bug]: Table sort no longer working


ArakTaiRoth:
### Describe the bug

The `_sort` or `onSort` functionality is no longer firing.

### To Reproduce

This can be seen in our [Storybook](https://ui-components.alberta.ca/?path=/docs/components-table--sortable). Or on either of our Angular or React demo apps.

### What browsers are you seeing the problem on?

Firefox, Chrome, Safari, Edge

### Additional context

I changed the code in `Table.svelte` to the last version of the [file](https://github.com/GovAlta/ui-components/blob/5a11b616f953fe8905a42a8b69655bac3d0cea0a/libs/web-components/src/components/table/Table.svelte), and table sorting worked. So this problem was introduced in this [commit](https://github.com/GovAlta/ui-components/commit/9a1054f6e76e81ad328927ae30a5d33f648fffba)

mamadoutouregoa:
By the way the pagination no longer works as well

PR #1338, opened 2023-08-24
fix(#1327) Fix table sorting

amar-ABgov:
* Couldn't add a test case to validate the `_sort` event firing, as the `goa-table-sort-header` is within a slot and was hard to test with`@testing-library/svelte`. Open to suggestions on this
* Added a small test case on the react component but that isn't the actual the issue


Radha-Alberta:
- [ ] Sorting does not work in: React App, StoryBook.

chrisolsen (dismissed):

ArakTaiRoth (approved):

ArakTaiRoth (approved):

chrisolsen (approved):

Issue #1294, opened 2023-07-26
[Bug]: Accordion heading not responding correctly on smaller screens


twjeffery:
### Describe the bug

When stacking on smaller screens, the smaller text and badge is too close to the edge of the accordion container. 

Current:
<img width="418" alt="image" src="https://github.com/GovAlta/ui-components/assets/82968683/fd455524-37b7-4716-ba03-aab9f86f693c">

Design intent:
<img width="959" alt="image" src="https://github.com/GovAlta/ui-components/assets/82968683/6c0392f7-5ae0-4699-81e5-dd32996d0930">

### To Reproduce

Can see on Storybook
https://ui-components.alberta.ca/iframe.html?viewMode=docs&id=components-accordion--basic#secondary-text-and-badges

PR #1365, opened 2023-09-07
fix(#1294): Accordion heading not responding correctly on smaller scr…

vanessatran-ddi:
After fixing: 
image

### DIFF
--- a/libs/web-components/src/components/accordion/Accordion.svelte
+++ b/libs/web-components/src/components/accordion/Accordion.svelte
@@ -27,14 +27,26 @@
   export let ml: Spacing = null;
 
   let _hovering: boolean = false;
+  let _titleEl: HTMLElement = null;
+  let _headingContentSlotChildren: Element[] = [];
 
   $: isOpen = toBoolean(open);
-
   onMount(() => {
     validateRequired("GoAAccordion", {heading})
     validateHeadingSize(headingsize);
+    _headingContentSlotChildren = getChildren();
   })
-
+  function getChildren(): Element[] {
+    if (_titleEl) {
+      const slot = _titleEl.querySelector("slot") as HTMLSlotElement;
+      if (slot) {
+        return [...slot.assignedElements()];
+      } else {
+        return [..._titleEl.children] as Element[];  // unit tests
+      }
+    }
+    return [];
+  }
 </script>
 
 <!-- HTML -->
@@ -56,10 +68,12 @@
       <goa-icon type="chevron-forward"
         fillcolor={_hovering?"var(--goa-color-interactive-hover)": "var(--goa-color-interactive-default)"}
       ></goa-icon>
-      <div class="title">
+      <div class="title" bind:this={_titleEl}>
         <span class="heading heading-{headingsize}" data-testid={`${testid}-heading`}>{heading}</span>
         <span class="secondary-text">{secondarytext}</span>
-        <div class="heading-content">
+        <div class="heading-content"
+             class:heading-content-top={_headingContentSlotChildren.length}
+        >
           <slot name="headingcontent" />
         </div>
       </div>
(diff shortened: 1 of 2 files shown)


libs/web-components/src/components/accordion/Accordion.svelte
```diff
@@ -27,6 +27,7 @@
   export let ml: Spacing = null;
 
   let _hovering: boolean = false;
+  let _headingContentHeight = 0;
```
ArakTaiRoth, line 30: Maybe I'm stupid and just missing something obvious. But why are we setting this? As far as I can tell, there's no place this can change, nowhere actually sets this variable. So on line 64, we could just use 0, and line 65 doesn't make sense, because it will never be greater than 0.

vanessatran-ddi: @ArakTaiRoth yes it's a good question. 

The _headingContentHeight is binding with div class="heading-content" on line 65, so if slot name="headingcontent" is passed with at least a >0 px height dom element, then line 65 condition will be true for the class "heading-content-top" 

The issue I try to solve is the gap 8px between the heading and the slot "heading-content" 
image

At first I chose the solution: .heading-content {margin-top: 8px;}

It doesn't work in a case that accordion doesn't have "slot heading-content".

image

I tried CSS `:empty` for the div.heading-content and it doesn't work, the space isn't correct, because svelte always render the div.heading-content with a content inside which is an empty slot. 

So I made a decision to detect the inner-height of the div.heading-content to detect and assign the class name .heading-content-top whenever it is possible.

ArakTaiRoth (dismissed):

libs/web-components/src/components/accordion/Accordion.svelte
vanessatran-ddi: > https://dev.to/danawoodman/svelte-quick-tip-styling-conditional-named-slots-4817 Can we use any of these techniques to check if the slot is empty?
> 
> While it fixes the bug, checking the innerHeight seems hacky to me.

@amar-ABgov Yes let me try some other ways. 
I also agree that innerHeight is a bit hacky. I'll get back if css way can work.

libs/web-components/src/components/accordion/Accordion.svelte
```diff
@@ -202,6 +202,13 @@
     summary .title {
       flex-direction: column;
       align-items: flex-start;
+      padding-bottom: 0.875rem;
+    }
+    summary .title span {
+      padding-bottom: 0;
+    }
+    :global([slot="headingcontent"]) {
```
chrisolsen, line 210: Can someone explain what is happening when using `:global`? The previous solution was a little hacky, but it made sense. 

One other option to do what you are trying to do _might_ be to use a `:has` selector

```
[slot="headingcontent"]:has(*) {
  margin-top: var(--goa-space-xs);
}

vanessatran-ddi: @chrisolsen and @amar-ABgov 
This is a good learning experience for me. I admitted I was carelessly checking `:global` and think it worked. My browser has a cache in CSS. 

The way I did with `:global` after testing again, as well as using `:has` doesn't work in this case. So I did some research. 

- So `:has` may not work for some browsers. [Browser compatibility](https://developer.mozilla.org/en-US/docs/Web/CSS/:has#browser_compatibility) 
- `:global` seems to override the style defined under the slot (https://learn.svelte.dev/tutorial/component-styles). If I want to add a margin-top for the div which is a wrapper for the slot itself, it cannot traverse back to the parent and set the style. 

So I reverted back to this solution, since "margin-top" is only for the mobile site case, it seems to be simple enough.

chrisolsen: FYI: We already use `:has` in other places, since Firefox is the only one that doesn't have it.

chrisolsen (approved):

ArakTaiRoth (approved):

Issue #1359, opened 2023-09-05
FocusTrap allowing focus on element with tabindex=1


chrisolsen:
## Acceptance Criteria:

When tabbing within a calendar the last calendar date sometimes becomes focused even though it has a tabindex of -1.

Need to add
![image](https://github.com/GovAlta/ui-components/assets/41582/8919927d-1566-43c2-bc55-b65986836214)

```
+    if (element.tabIndex < 0 || element.getAttribute?.("tabindex") === "-1") return null;
```

PR #1368, opened 2023-09-08
fix(#1359): prevent focus-trap from allowing element with a tabindex=…

chrisolsen:
…-1 from attaining focus

### DIFF
--- a/libs/web-components/src/components/focus-trap/FocusTrap.spec.ts
+++ b/libs/web-components/src/components/focus-trap/FocusTrap.spec.ts
@@ -1,17 +1,29 @@
-import '@testing-library/jest-dom';
-import { render, cleanup, fireEvent } from '@testing-library/svelte';
-import FocusTrapTestComponent from './FocusTrapTestComponent.svelte';
+import "@testing-library/jest-dom";
+import {
+  render,
+  cleanup,
+  fireEvent,
+  createEvent,
+  waitFor,
+} from "@testing-library/svelte";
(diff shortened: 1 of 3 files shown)


chrisolsen:
Can be tested with the following. Where the last button should not be focusable (previous was when shift-tabbing)
```
    <goa-popover>
      <goa-button slot="target">Click me</goa-button>
      This is the popover content
      <button tabindex="0">Close</button>
      <button tabindex="0">Close</button>
      <button tabindex="0">Close</button>
      <button tabindex="-1">Close</button>
    </goa-popover>
```

libs/web-components/src/components/focus-trap/FocusTrap.spec.ts
```diff
@@ -1,17 +1,29 @@
...
+  cleanup,
+  fireEvent,
+  createEvent,
+  waitFor,
+} from "@testing-library/svelte";
+import FocusTrapTestComponent from "./FocusTrapTestComponent.svelte";
 
 afterEach(cleanup);
 
 // This test is blocked due to JEST handling slot elements differently than web browser
```
ArakTaiRoth, line 13: You've modified this test, and removed the skip. Is this comment still applicable?

chrisolsen: The test was modified in an attempt to get it working and the skip was just moved, but it's still there.

ArakTaiRoth (approved):

Issue #1380, opened 2023-09-22
Bug: Problems with Input Time


chrisolsen:
The storybook example for GoAInputTime, is using a value of type `Date`, but only a `string` type is valid.

https://ui-components.alberta.ca/?path=/docs/components-inputs--time#time

In addition, the default step value of 60 is not actually set.

https://ui-components.alberta.ca/?path=/docs/components-inputs--time-with-seconds

![image](https://github.com/GovAlta/ui-components/assets/41582/13fef775-518d-4fca-9a85-a9aab684b4aa)

## Acceptance Criteria:
1. Storybook example is changed to use a string value instead of Date
2. Default step value is changed to 60, so seconds should not be shown by default.

PR #1382, opened 2023-09-22
fix(#1380): Modify step value on Input. Modify time documentation to …

ArakTaiRoth:
…use string.


libs/web-components/src/components/input/Input.svelte
```diff
@@ -46,7 +46,7 @@
   export let arialabelledby: string = null;
   export let min: string = "";
   export let max: string = "";
-  export let step: number = 1;
+  export let step: number = 60;
```
chrisolsen, line 49: The default value needs to be `1`, as this value is what is used for a numeric input.

chrisolsen (approved):

PR #1405, opened 2023-10-16
adjust css for typography (do not merge, need decision)

vanessatran-ddi:
### Description: 
Fixing typography css.
image
Last paragraph shouldn't have `margin-bottom`
image


libs/styles/src/lib/reset.css
```diff
@@ -120,12 +120,15 @@ h6 {
...
   margin-top: -1rem;
 }
-
 p {
-  margin-bottom: 1rem;
+  margin: 0 0 1rem;
   font-size: var(--goa-font-size-4);
 }
 
+p:is(:last-child) {
```
ArakTaiRoth, line 128: I'm a little concerned about this one. This could have fairly far reaching implications for our users. Is this because of p tags being used in modals? If so, could we maybe limit this to p tags inside goa-modal for example?

ArakTaiRoth:
Don't need this anymore

Issue #1330, opened 2023-08-21
Target _blank for Microsite Header


ArakTaiRoth:
# Info
Currently the URL for our Microsite header guides users to the alberta.ca website. It does this in the current window/tab. It would be ideal for sites that it uses a new tab for it's linking.

## Acceptance Criteria:
1. The Microsite header URL uses `target="_blank"`
2. The Feedback URL property also uses `target="_blank"`

alihakak:
Would be even more helpful if we make it an optional prop and also `feedbackUrl` can open in new tab as well.

twjeffery:
Initial design added here: https://www.figma.com/file/G9Mkkio4zXDNIdba0UuLfF/Component---Microsite-header?type=design&node-id=1901-23676&mode=design&t=RF0x4myJ0phfYHxZ-4


![Image](https://github.com/GovAlta/ui-components/assets/82968683/576f5e21-90da-4a54-ac39-3557efdf43aa)

PR #1428, opened 2023-10-31
feat(#1330): Microsite header - open external urls in a new window

amar-ABgov:


libs/react-components/src/lib/microsite-header/microsite-header.tsx
```diff
@@ -41,6 +47,8 @@ export const GoAMicrositeHeader: FC<HeaderProps> = ({
       feedbackurl={feedbackUrl}
       data-testid={testId}
       maxcontentwidth={maxContentWidth}
+      feedbackurlexternal={feedbackURLExternal}
```
chrisolsen, line 50: Keep the type case of `URL` consistent with `feedbackUrl`

libs/react-components/src/lib/microsite-header/microsite-header.tsx
```diff
@@ -41,6 +47,8 @@ export const GoAMicrositeHeader: FC<HeaderProps> = ({
...
+      headerurlexternal={headerURLExternal}
```
chrisolsen, line 51: With this name it isn't obvious that it is a boolean variable. Maybe basing the name off the `<a target="_blank">` attribute would make things more clear. So using `feedbackUrlTarget="blank"` might make more sense.

libs/web-components/src/components/microsite-header/MicrositeHeader.html-data.json
```diff
@@ -20,6 +20,38 @@
...
+        {
+          "name": "self",
+          "description": "target=_self"
+        },
+        {
+          "name": "blank",
+          "description": "target=_blank"
+        }
+      ],
+      "type": "boolean"
```
chrisolsen, line 38: This type need to be updated to match the `self | blank`

libs/web-components/src/components/microsite-header/MicrositeHeader.html-data.json
```diff
@@ -20,6 +20,38 @@
       "default": "100%",
       "description": "Maximum width of the content area",
       "type": "string"
+    },
+    {
+      "name": "headerurltarget",
+      "default": "false",
```
ArakTaiRoth, line 26: Change to `default: self`

libs/web-components/src/components/microsite-header/MicrositeHeader.html-data.json
```diff
@@ -20,6 +20,38 @@
...
+        {
+          "name": "blank",
+          "description": "target=_blank"
+        }
+      ],
+      "type": "boolean"
+    },
+    {
+      "name": "feedbackurltarget",
+      "default": "false",
```
ArakTaiRoth, line 42: Change to `default: self`

libs/docs/src/components/goa/microsite-header.stories.mdx
```diff
@@ -34,6 +34,19 @@ import { Meta, Story } from "@storybook/addon-docs";
     defaultValue="100%"
     description="Maximum width of the content area"
   />
+  <Prop
+    name="feedbackUrlTarget"
+    type="self | blank"
+    defaultValue="self"
```
ArakTaiRoth, line 40: After talking to @Spark450  and @twjeffery it was decided that these default values should actually be blank instead, so by default they open up in separate windows. Make sure this gets changed here, the `json` file, and the `svelte` file. Probably need to make some changes in tests as well.

ArakTaiRoth (approved):

Issue #1346, opened 2023-08-28
maxwidth property for Modal


ArakTaiRoth:
# Info
Modify documentation for Max Width. And remove `scrollable` from React Modal component.

## Acceptance Criteria:
1. Max Width title is not duplicated in Storybook
2. Scrollable is removed from React Modal component (not testable)

PR #1436, opened 2023-10-31
chore(#1346): Update modal Storybook and react component

ArakTaiRoth:


Radha-Alberta:
- Callout variant is having both 'x' icon and action button, should only contain action buttons as per figma design

image

**Figma:**
image
image

Radha-Alberta:
> * [ ]  Callout variant is having both 'x' icon and action button, should only contain action buttons as per figma design
> 
> image
> 
> **Figma:** image image

We will create a new ticket to track this issue

chrisolsen (approved):

Issue #1446, opened 2023-11-03
Bug: Modal width is always 100%


ArakTaiRoth:
# Info
Right now, our modal width is always set to take up full 100% width no matter what. The `maxwidth` property we currently have does nothing, as it's only applied on the mobile breakpoint.

## Acceptance Criteria:
1. `maxwidth` should work as expected and limit the width of the modal to whatever is set

PR #1447, opened 2023-11-03
fix(#1446): Added maxwidth to modal CSS

ArakTaiRoth:


chrisolsen:
As I mentioned in our earlier conversation I thought the reason might have been an earlier made error, which it was. 
image

The existing `max-width` setting got accidentally changed to an incorrect media selector.

chrisolsen (approved):

chrisolsen (approved):

Issue #1398, opened 2023-10-05
KeyPress Event for Input Component


ArakTaiRoth:
# Info
Right now, there is no way to do anything on key press in our Input component. This means people are not allowed to track the Enter key being pressed in a form and being able to submit the form (like a search input).

## Acceptance Criteria:
1. New event for Input component - KeyPress
2. This event should pass along the value of the input that the event was started in

PR #1456, opened 2023-11-09
feat(#1398): Add keyPress event for input

amar-ABgov:


Radha-Alberta:
- inputbox value and value in event triggered are different, Is this expected ?

image

libs/react-components/src/lib/input/input.tsx
```diff
@@ -104,11 +105,13 @@ export interface InputProps extends BaseProps {
   step?: number;
   onFocus?: OnFocus;
   onBlur?: OnBlur;
+  onKeyPress?: OnKeyPress
```
ArakTaiRoth, line 108: Question, I must have missed this earlier. Why don't we move `onChange`, `onFocus`, `onBlur`, and now `onKeyPress` to `BaseProps`, so we don't have to repeatedly add it to every other Input type we have?

chrisolsen: For consistency it would make sense to also have the onKeyPress o on the TextArea as well.

ArakTaiRoth (dismissed):

libs/react-components/src/lib/input/input.tsx
```diff
@@ -194,17 +189,24 @@ export function GoAInput({
       onBlur?.(name, value);
     };
 
+    const keypressListner = (e: unknown) => {
```
chrisolsen, line 192: Small spelling mistake `Listner`

libs/react-components/src/lib/textarea/textarea.tsx
```diff
@@ -70,9 +72,16 @@ export function GoATextArea({
       onChange(name, value);
     };
 
+    const keypressListner = (e: unknown) => {
```
chrisolsen, line 75: Same spelling mistake

Issue #1360, opened 2023-09-05
Dropdown menu rendering


chrisolsen:
## Acceptance Criteria:
After merging in dropdown changes the menu no longer looks as it did.

After
![image](https://github.com/GovAlta/ui-components/assets/41582/32ee65bd-f046-400a-b5f2-218d49145789)

chrisolsen:
Also renders to narrow when small strings exist as labels

PR #1460, opened 2023-11-10
fix(#1360): Removed width from Pagination dropdown

ArakTaiRoth:
This should make the Pagination page select dropdown autosize, which should mean that 2 character numbers should fit now.


Radha-Alberta:
- [ ] 2 digits still cannot fit in pagination dropdown

image

chrisolsen (approved):

Issue #1459, opened 2023-11-10
Documentation: Pagination React code is wrong


ArakTaiRoth:
## Acceptance Criteria:
1. Items per page example should be updated
    * `pageUsers.map(u: any =>` should be `pageUsers.map((u:any) =>`
    * `variant={args.variant}` should be removed as `args` is specific to Storybook
2. Basic example should be updated
    * `pageUsers.map(u =>` should be `pageUsers.map((u) =>`

PR #1462, opened 2023-11-10
chore(#1459): Updated Pagination React code examples

ArakTaiRoth:


libs/docs/src/components/common/Pagination.stories.mdx
```diff
@@ -302,7 +302,7 @@ export const PaginationTableTemplate = (args) => {
             </tr>
           </thead>
           <tbody>
-            {pageUsers.map(u =>
```
chrisolsen, line 305: Adding braces is a minor syntactical change, where the previous version, is a common way of doing it. However, in this example since this is React code, the `User` type should be specified..but to do do this additional types need to be added within the code snippet. 

The most likely reason for the lack of types here is that the js code from within storybook was copied and pasted here.

libs/docs/src/components/common/Pagination.stories.mdx
```diff
@@ -495,7 +495,7 @@ export const PaginationTableTemplate = (args) => {
               </tr>
             </thead>
             <tbody>
-              {pageUsers.map(u:any =>
```
chrisolsen, line 498: The type should be set to `User` not `any`

chrisolsen (changes requested):

chrisolsen (approved):

PR #1546, opened 2024-01-02
Adding disabled and width attributes to the date-picker

hassananv:


libs/react-components/src/lib/date-picker/date-picker.tsx
```diff
@@ -24,10 +26,14 @@ export interface GoADatePickerProps extends Margins {
   value?: Date;
   min?: Date;
   max?: Date;
+  disabled?: boolean;
+  width?: string;
   onChange: (name: string, value: Date) => void;
 }
 
 export function GoADatePicker({
+  disabled = false,
```
chrisolsen, line 35: The default value will be set within the svelte component, so the `= false` can be removed here.

libs/web-components/src/components/date-picker/DatePicker.svelte
```diff
@@ -14,18 +15,21 @@
...
   export let mb: Spacing = null;
   export let ml: Spacing = null;
+  $: isDisabled = toBoolean(disabled);
 
   let _rootEl: Element;
   let _date: Date;
   let _showPopover: boolean = false;
 
   onMount(async () => {
+    await tick();
```
chrisolsen, line 32: Is this fixing an angular binding issue?

libs/web-components/src/components/date-picker/DatePicker.svelte
```diff
@@ -76,7 +80,11 @@
   }
 
   function showCalendar() {
-    _showPopover = true;
+    if (!isDisabled) {
```
chrisolsen, line 83: Can you change this to
```
_showPopover = !isDisabled;
```

chrisolsen (changes requested):

libs/web-components/src/components/date-picker/DatePicker.svelte
hassananv: Yes, without this line, Angular binding data does not take effect.

libs/react-components/src/lib/date-picker/date-picker.tsx
```diff
@@ -5,6 +5,8 @@ interface WCProps extends Margins {
   ref: React.RefObject<HTMLElement>;
   name?: string;
   value?: string;
+  disabled?: boolean;
```
chrisolsen, line 8: The only other thing that is need are some tests.
- validate that the disabled prop is in fact disabling it i.e. doesn't respond to user events (more difficult)
- validate that the width is binding to the internal text input (easy)

ArakTaiRoth:
Re-opened this as we can use this work and just need to add to it.

ArakTaiRoth:
Closing this PR, as part of the work for disabled has been done in this [PR](https://github.com/GovAlta/ui-components/pull/2091)

Issue #1379, opened 2023-09-21
Spike: Investigate options for Angular type safety


ArakTaiRoth:
# Info
Currently there is no type safety with Angular. We ask people to implement `CUSTOM_ELEMENTS_SCHEMA` when they import our packages. By doing this, it disables type safety for their application, which is not desirable. This disables things like failing builds if they misspell a property or use a property that doesn't exist. They expect the application to work, and it doesn't, but there was no build error, so they're left to figure it out on their own what the issue is.

We have seen teams build their own Angular wrappers around our components to combat this issue. But that's a lot of work to hoist onto every other Angular team.

## Options
We have come up with a few things that could be options:
1. [Custom Language Server](https://github.com/Matsuuu/custom-elements-language-server)
2. Implement our own Angular wrappers (can this be done in such a way that it doesn't constitute a major version change)
3. Implement our own custom VS Studio plugin (this could work, but only helps developers using VS Studio)

We aren't exclusive to the options listed above, if there is another option that we haven't presented that could be better, please investigate it.

## Acceptance Criteria:
We have a list of options that could work (what was listed above, we don't know it will actually work). With an idea for each one on effort to implement, how much it will effect our users (effort for them), and how effective it will be at solving the issue.

cjam:
what is the status of this issue?  Where does angular type safety fall on the roadmap for this project?

chrisolsen:
@cjam Angular wrapping components are currently in progress.

cjam:
I'm sure you're already aware of all the work Mitch R did, but in the event that you are not he has wrapped all / most of the components from this library over here:

https://github.com/GovAlta-EMU/aed-components/tree/main/libs/aed-ui-components/src/lib/goa-components

Perhaps it could just be pulled in and maintained in here by your team instead of externally.

chrisolsen:
Thanks @cjam, but I don't have an EMU account so I don't have access to that repo. Would it be possible to send me the zip file via Slack?

ArakTaiRoth:
This has been resolved with Angular wrappers

PR #1558, opened 2024-01-12
nx and svelte 4 upgrade

chrisolsen:


chrisolsen:
Issue 2: This is related to issue #1379. With the upgrade a index.d.ts file is being generated, but instead the types it just references the svelte files. This might be reason to include 1379 fixes in this PR.

Issue 3 is fixed :+1:

PR #1582, opened 2024-01-25
Angular automation testing

Radha-Alberta:
- Added data-cy & testid attribute in the Angular and React Components.
- These attributes used to avoid fetching components  `goa-input` via name attribute


ArakTaiRoth:
Closing as we aren't doing cypress automation like this anymore

Issue #1496, opened 2023-12-06
Dropdown menu doesn't work when set to a % value


chrisolsen:
# Info

This was previously marked as fixed, it is not.

When you use a percentage for the width, the dropdown width doesn't actually get determined by the parent container

## Acceptance Criteria

1. If you offer percentage of width it should be based off of the parent container

ArakTaiRoth:
Just adding a comment here for transparency. I spoke to Amar about this and when I went to show him the issue, I was unable to replicate the issue. So this might actually be fixed and was opened erroneously. Amar is going to investigate a little further, but I said if he can't replicate then move to Blocked and I will look at when I'm back in the office next week.

PR #1633, opened 2024-02-13
fix(#1496): Restrict dropdown width to parent container - code review needed

amar-ABgov:
### Issue
In scenarios where dropdown width is set to be more than the parent container the dropdown should respect the parent container width. This applies when there isn't enough screen real-estate. 

The videos show how the below piece of React code is rendered before and after this change
```jsx
      <GoADropdown name="colors" value={"red"} onChange={onChange} width="700px">
        <GoADropdownItem value="red" label="Red" />
        <GoADropdownItem value="green" label="Green" />
        <GoADropdownItem value="blue" label="Blue" />
      </GoADropdown>
```

#### Note
The issue still persists when the screen size is changed. This only computes the parent container width on initial render. 

#### Before

https://github.com/GovAlta/ui-components/assets/119442489/84a1bc1c-6d40-4dba-ae71-2247e350fd33

#### After

https://github.com/GovAlta/ui-components/assets/119442489/2094bcd8-5289-41a4-a21e-3be7ba468d99

### DIFF
--- a/libs/web-components/src/components/dropdown/Dropdown.spec.ts
+++ b/libs/web-components/src/components/dropdown/Dropdown.spec.ts
@@ -75,7 +75,7 @@ describe("GoADropdown", () => {
       });
 
       // show menu
-      dropdownIcon && await fireEvent.click(dropdownIcon);
+      dropdownIcon && (await fireEvent.click(dropdownIcon));
       await waitFor(() => {
         expect(popover?.getAttribute("open")).toBe("true");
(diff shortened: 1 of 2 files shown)


lizhuomeng71:
Issue is verified on chrome, edge and firefox

lizhuomeng71:
Screenshot 2024-02-23 142946

ArakTaiRoth (approved):

Issue #1510, opened 2023-12-11
[Bug]: Search Input in Filterable Dropdown is Being Autofilled


francisagyapong-goa:
### Describe the bug

When Dropdown is autofilled, a value isn't actually set in the Dropdown, the change event isn't fired.

### Provide Code

Please use https://purchasing.dev.alberta.ca/supplier-registration to confirm this behaviour. Recording of a walkthrough has also been attached here.
[supplier-registration-autofill-behaviour-2023-12-11_08.56.44.webm](https://github.com/GovAlta/ui-components/assets/94571832/410933c0-67fc-402a-8c3d-1506f657bb09)

NOTE: I can't get autofill to work in StackBlitz but here's code similar to what we have in our application: https://stackblitz.com/edit/b1due8?file=src%2Fapp%2Fapp.component.html

### To Reproduce

Accept browser's autofill suggestion

### What browsers are you seeing the problem on?

Chrome, Edge

## Acceptance Criteria

1. Make sure change event is fired when Dropdown is auto-filled so that Dropdown actually has a value.

chrisolsen:
I can't replicate this issue and neither can their code snippet. 

The autocomplete was disabled web-component v1.18 which is the same version set in the stackblitz code example, so I am wondering if the original bug is in an older version before the fix. @francisagyapong-goa Can you let us know what version of the web-component library you are using in the app?

francisagyapong-goa:
@chrisolsen We're still able to reproduce it - we're using web-components version 1.18.1 (the linked stackblitz uses 1.18.0).

The issue is that a filterable dropdown's search input is autofilled causing the user to think a value has been selected when in fact there's no value bound to it.

francisagyapong-goa:
@chrisolsen @vanessatran-ddi Was this only released for web-components, v1.17.x? We're currently on version 1.18.1.

BumbleB2na:
[...]

PR #1636, opened 2024-02-14
fix(#1510): dispatch value filterable dropdown when browser autofil

vanessatran-ddi:
Issue described: https://github.com/GovAlta/ui-components/issues/1510 

I tested with Chrome. You can set the contacts as the below:
image

The code I referred to can be found here: 
https://stackblitz.com/edit/b1due8?file=src%2Fapp%2Fapp.component.html

**Notice**: In order to test it, you must use the code in an angular app that doesn't navigate by router. (Our playground example app.component.html) is a good place to start. 

https://github.com/GovAlta/ui-components/assets/120135417/c95e600d-7768-4163-b528-81ed4f3b4a61

### DIFF
--- a/libs/web-components/src/components/dropdown/Dropdown.svelte
+++ b/libs/web-components/src/components/dropdown/Dropdown.svelte
@@ -1,7 +1,7 @@
 <svelte:options customElement="goa-dropdown" />
 
 <script lang="ts">
-  import { onMount } from "svelte";
+  import {onMount, tick} from "svelte";
 
   import type { GoAIconType } from "../icon/Icon.svelte";
   import type { Spacing } from "../../common/styling";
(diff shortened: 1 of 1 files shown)


ArakTaiRoth (approved):

libs/web-components/src/components/dropdown/Dropdown.svelte
```diff
@@ -258,6 +259,11 @@
       syncFilteredOptions();
       _isMenuVisible = true;
       _inputEl?.focus();
+
+      // the browser autocomplete filled in value while user isn't touching it
+      if (_filteredOptions.length === 1 && !_isTouched) {
+        onSelect(_filteredOptions[0]);
+      }
```
chrisolsen, line 266: Which event handler is calling the `showMenu` function? It might make more sense moving this logic there instead of this function which is called from quite a few different places. This way **may** also eliminate the need to have the `_isTouched` state.

chrisolsen: Of the events you listed it is only in the `onInputKeyUp` and the `onInputKeyDown` that you are changing the `_isTouchedStatus` on, so things are narrowed down to 2 events. However, since the keyup and keydown always happen together I don't think it needs to be in both of them, which narrows things back down to one place.

chrisolsen (approved):

ArakTaiRoth (approved):

Issue #1549, opened 2024-01-04
[Bug]: Spacing between radio items in horizontal orientation


Spark450:
## Acceptance criteria

- Spacing between radio options in the horizontal orientation should be 32px. Include (--goa-space-xl) on the right of each radio option when options are in horizontal orientation
 
![image](https://github.com/GovAlta/ui-components/assets/51723535/3795174c-054b-4efb-b000-5ea779ecfd22)

### To Reproduce

View radios in horizontal orientation in [Storybook docs](https://ui-components.alberta.ca/?path=/docs/components-radio--basic)

PR #1638, opened 2024-02-15
fix(#1549): Radio item padding adjustment

amar-ABgov:


libs/web-components/src/components/radio-group/RadioGroup.svelte
```diff
@@ -196,7 +196,7 @@
   }
 
   .goa-radio-label {
-    padding: 0 0.5rem;
+    padding: 0 var(--goa-space-xl) 0 var(--goa-space-xs);
```
chrisolsen, line 199: It would be more clear to add the additional padding in the scope of the `.goa-radio-group--horizontal` selector

```
.goa-radio-group--horizontal .goa-radio-label {
   padding-right: var(--goa-space-xl);
}
```

chrisolsen (approved):

ArakTaiRoth (approved):

Issue #1639, opened 2024-02-15
Web component's package.json `module` is set to non-existent index.mjs file


chrisolsen:
# Info

Angular is having a problem because our web-components package is pointing to a non-existent file `index.mjs`.

## Acceptance Criteria

1. Specifying the `module` as `index.js` works since the web-components aren't compiled to esm format

PR #1649, opened 2024-02-20
fix(#1639): web-components module file update

amar-ABgov:


ArakTaiRoth (approved):

chrisolsen (approved):

libs/web-components/package.json
```diff
@@ -14,7 +14,7 @@
   "private": true,
```
chrisolsen, line 14: Can you also remove this line (#14)

chrisolsen (changes requested):

chrisolsen (approved):

Issue #1492, opened 2023-12-01
Create table header column for number columns


chrisolsen:
# Info
We need a class to use for table headers that aligns content to the right. Much like the class we already have for table content. We can extend our current class `goa-table-number-column` to be able to apply it to headers.

## Acceptance Criteria
1. `goa-table-number-column` can be applied to table headers (`<th>` elements) that aligns content to the right

![image](https://github.com/GovAlta/ui-components/assets/41582/75188e10-5c06-44b1-bbb1-9a9cbf0a62a2)

PR #1658, opened 2024-02-21
fix(#1492) : Table - Adjust styling for number table header - code review needed

amar-ABgov:
### Issue 
extend our current class `goa-table-number-column` to be able to apply it to headers

### Fix
This wasn't working only when a `tableSortHeader` is used within a `th` element.

```jsx
<th className="goa-table-number-column">
            <GoATableSortHeader name="age" direction="asc">
              Age
            </GoATableSortHeader>
          </th>
```
this should now align the table header element to the right

### DIFF
--- a/libs/web-components/src/components/table/TableSortHeader.svelte
+++ b/libs/web-components/src/components/table/TableSortHeader.svelte
@@ -5,10 +5,24 @@
 </script>
 
 <script lang="ts">
+  import { onMount } from "svelte";
   export let direction: GoATableSortDirection = "none";
+
+  // Private
+  let _rootEl: HTMLElement;
+  onMount(() => {
+    if (_rootEl) {
+      // Add styling if an ancestor has a class to style number columns,
+      const hostEl = _rootEl.getRootNode().host;
+      const ancestor = hostEl?.closest("th.goa-table-number-header");
+      if (ancestor) {
+        _rootEl.style.setProperty("--header-text-align", "flex-end");
+      }
+    }
+  });
 </script>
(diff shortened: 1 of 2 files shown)


libs/web-components/src/components/table/TableSortHeader.svelte
```diff
@@ -5,10 +5,24 @@
 </script>
 
 <script lang="ts">
+  import { onMount } from "svelte";
   export let direction: GoATableSortDirection = "none";
+
```
chrisolsen, line 10: Looks good. Can you format this code as all the single quotes will get changed the next time someone works on this file.

libs/web-components/src/components/table/TableSortHeader.svelte
```diff
@@ -5,10 +5,24 @@
...
+  // Private
+  let _rootEl: HTMLElement;
+  onMount(() => {
+    if (_rootEl) {
+      // Add styling if an ancestor has a class to style number columns,
+      const hostEl = _rootEl.getRootNode().host;
+      const ancestor = hostEl?.closest("th.goa-table-number-header");
+      if (ancestor) {
+        _rootEl.style.setProperty("--header-text-align", "flex-end");
```
chrisolsen, line 19: The alignment styles are different. Here it is using flex-end and in the css it uses text-align: right. I am sure they are both valid, but let's keep them consistent with one another.

ArakTaiRoth (approved):

chrisolsen (approved):

PR #1662, opened 2024-02-22
 1577: textfield border renderring on error and focus

vanessatran-ddi:
Continue from #1635 
Tested with the below cases:
* trailing, leading, and both
* trailing, leading and both under error case
* disabled
* no trailing, no leading
* goa-input used under dropdown, focus border remains the same.

https://github.com/GovAlta/ui-components/assets/120135417/ca0d96b2-2025-4e69-bba5-96b6e052f5f0

### DIFF
--- a/libs/web-components/src/components/input/Input.svelte
+++ b/libs/web-components/src/components/input/Input.svelte
@@ -188,8 +188,8 @@
   <div
     class="goa-input variant--{variant} type--{type}"
     class:input--disabled={isDisabled}
-    class:input-leading-content={_leadingContentSlot}
-    class:input-trailing-content={_trailingContentSlot}
+    class:leading-content={_leadingContentSlot}
+    class:trailing-content={_trailingContentSlot}
     class:error={isError}
   >
     {#if prefix}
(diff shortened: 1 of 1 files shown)


ArakTaiRoth (approved):

libs/web-components/src/components/input/Input.svelte
```diff
@@ -481,19 +481,19 @@
...
-  .input-leading-content:focus-within,
-  .input-trailing-content:active,
-  .input-trailing-content:focus,
-  .input-trailing-content:focus-within {
+  input.input-leading-content:active,
+  input.input-leading-content:focus,
+  input.input-leading-content:focus-within,
+  input.input-trailing-content:active,
+  input.input-trailing-content:focus,
+  input.input-trailing-content:focus-within {
```
chrisolsen, line 496: I don't think the additional selectivity is required.

chrisolsen: Using the same class name in the component for different elements is rather confusing.

chrisolsen (approved):

ArakTaiRoth (approved):

PR #1673, opened 2024-02-26
fix: include v16 in angular peer deps

ArakTaiRoth:


chrisolsen (approved):

ArakTaiRoth:
Closed because there are a lot more issues that are being pushed to alpha

PR #1677, opened 2024-02-28
fix(#1679): null querySelector errors in dropdown

chrisolsen:
This seemed to be an angular issue only when changing routes from a page that contained a dropdown component.

The error was due to the Dropdown component responding to the `slotchanged` event, which made reference to one of the elements within the page. However, by the time the reference was made the html element had been removed from the DOM; resulting in the null exception.


lizhuomeng71:
Issue was verified on react and angular for dropdown

ArakTaiRoth (approved):

PR #1690, opened 2024-03-06
fix: add @container queries to FormStepper and FormStep

chrisolsen:
Peek 2024-03-06 09-29


ArakTaiRoth (approved):

lizhuomeng71:
Tested for react and angular with default or explicit set container width, also tested for mobile and desktop

PR #1705, opened 2024-03-13
Popover offset

chrisolsen:
This PR fixes a few issues 

1. In certain conditions the `baseWidth` was unable to be calculated, this resulted in 0px width dropdown items being rendered.
2. The @container queries within the tabs (and probably any other element) prevented a modal, that is initiated from within a GoATab component, from going full screen, but instead confined it to within the GoATabs boundries.
3. When using a Popover within a @container query, it was found that the calculated offset values should be 0, to get things to work. It was then found that using the zero values still worked for the non-relative popovers. So these calculations were removed.


lizhuomeng71:
Issue is verified for react and angular, for tabs, and dropsdowns in various conditions

ArakTaiRoth (dismissed):

PR #1708, opened 2024-03-15
fix: tabs are now links instead of buttons

chrisolsen:
Hash urls will allow the url to change when the tab is chaned, allowing for `popstate` events to be handled for any page updates required. At a later time this will also allow the tab's `open` status to be set if a url with one of hashes is opened directly.

To test this, if the previous Tab tests function as they did before then it is all good.


lizhuomeng71:
issue is verified for chrome edge and firefox, and react and angular,

PR #1751, opened 2024-04-04
fix: replace no longer used @container query with @media

chrisolsen:
Before
image

After
image


lizhuomeng71:
issue is verified on firefox and chrome for angular and react for desktop and mobile site

syedszeeshan (approved):

Issue #1466, opened 2023-11-15
[Bug]:  Table's variant `relaxed` and `normal`


vanessatran-ddi:
### Describe the bug

The variant `relaxed` and `normal` make no difference. 
![image](https://github.com/GovAlta/ui-components/assets/120135417/8acafbb6-eabc-434c-9e30-71752643ce71)
![image](https://github.com/GovAlta/ui-components/assets/120135417/2a34e017-c511-4d90-a2fe-eda704879ee2)

They are no different, regarding about the padding in the table cell. 
Definition of "variant" 
variant: A relaxed variant of the table with more vertical padding for the cells

### To Reproduce

https://deploy-preview-27--abgov-ui-component-docs.netlify.app/components/table
Sandbox, when changing the variant property, makes no difference.

Spark450:
@vanessatran-ddi it looks like we can get both normal and relaxed table variants to display in storybook. Does that mean that this could be sandbox issue?

bdfranck:
The CSS selectors for relaxed rows (and a sticky header) don't work because the `goa-table` element doesn't get a `relaxed` or `sticky` class.

https://github.com/GovAlta/ui-components/blob/b69048853d5bc2b6f1e8bc96eece8759e1a08636/libs/web-components/src/assets/css/components.css#L83-L90

It looks like the wrapper `div` in the element's shadow DOM that get these classes 👀.

![Image](https://github.com/GovAlta/ui-components/assets/1479091/4eaaef1e-3dfc-49e5-8540-18895ddf32ce)

I'm not sure how best to approach it. I'll need to spend more time researching how to style a slotted element in this situation. 🤔

PR #1788, opened 2024-04-18
fix(#1466): apply padding when variant=relaxed

syedszeeshan:
Instead of using JS for dom manipulation and adding 'relaxed' class to all eligible tds, the customElement <goa-table> is forced to spit out 'variant' attribute, and then the css selector goa-table[variant='relaxed'] td does work. See Chris' last comment below

image

### DIFF
--- a/libs/web-components/src/assets/css/components.css
+++ b/libs/web-components/src/assets/css/components.css
@@ -80,10 +80,6 @@ goa-table table {
   border-collapse: collapse;
 }
 
-goa-table.relaxed td {
-  padding: 1rem;
-}
-
 goa-table.sticky thead {
   position: sticky;
   top: 0;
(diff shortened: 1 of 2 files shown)


bdfranck:
Awesome! You could remove the old unused CSS:

https://github.com/GovAlta/ui-components/blob/b69048853d5bc2b6f1e8bc96eece8759e1a08636/libs/web-components/src/assets/css/components.css#L83-L85

syedszeeshan:
screenshot below after the above changes, Table with relaxed vs normal spacing

Playground-Table-relaxed-vs-normal

lizhuomeng71:
Issues is verified on Chrome, Firefox, and edge for react and angular, for mobile and desktop site
Screenshot 2024-04-19 141319

chrisolsen:
I am pretty sure this can be fixed with the following updates to the component.css file

chrisolsen:
My code worked since I was testing via Angular which allowed the defined `variant` prop to be rendered in the DOM, then allowing my css selector to work; this did not work when testing in the Svelte playground.

To fix this problem the `svelte:option` needs to be changed to the following.
```
<svelte:options customElement={{
  tag: "goa-table",
  props: {
    variant: { reflect: true }
  }
}} />
```
This will ensure that the prop is visible in the DOM, and the previously defined css rule can take affect.

chrisolsen (approved):

Issue #1796, opened 2024-04-23
Update to Form Item spacing


twjeffery:
## Acceptance Criteria:
- Spacing around form item should be as outlined below in design.
- Large form label needs more space between text and input field than regular label size (0.75 rem/--goa-spacing-xs)
- Both Error text and Helper text should be using text style (--goa-typography-body-xs)

<img width="859" alt="image" src="https://github.com/GovAlta/ui-components/assets/82968683/1e0435f9-2284-4165-81dc-e2f0ed288a35">

[Design and specs in Figma here](https://www.figma.com/file/LUNrOt4guKPgQyicV27389/Form-item-alignment?type=design&node-id=8-73142&mode=design&t=O4TRfiUKeqtcwYOv-4)

PR #1808, opened 2024-04-29
fix(#1796): update form item spacing and font

bdfranck:
This PR addresses https://github.com/GovAlta/ui-components/issues/1796 with the following changes:
- Adds extra spacing between a large label and the input
- Uses the proper font values for the error message
- Vertically aligns the error icon with the new line height using a transform
- Replaces other spacing values in the `FormItem` component with design tokens

Tasks:
image

### DIFF
--- a/libs/web-components/src/components/form-item/FormItem.svelte
+++ b/libs/web-components/src/components/form-item/FormItem.svelte
@@ -81,39 +81,40 @@
 
   .label {
     display: block;
-    font-weight: var(--goa-font-weight-bold);
-    color: var(--goa-color-text-default);
-    font-size: var(--goa-font-size-4);
-    padding-bottom: 0.5rem;
+    font: var(--goa-typography-heading-s);
+    padding-bottom: var(--goa-space-xs);
   }
 
   .label.large {
     font: var(--goa-typography-heading-l);
+    padding-bottom: var(--goa-space-s);
   }
 
   .label em {
+    font: var(--goa-typography-body-xs);
     color: var(--goa-color-greyscale-700);
-    font-weight: var(--goa-font-weight-regular);
-    font-size: var(--goa-font-size-2);
-    line-height: var(--goa-line-height-1);
-    font-style: normal;
-  }
-
-  .form-item-input {
-    margin-bottom: 0.25rem;
-  }
-
-  .help-msg {
-    font-size: var(--goa-font-size-2);
-    color: var(--goa-color-text-default);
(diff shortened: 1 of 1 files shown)


bdfranck:
I've tested in Edge...
image

image

bdfranck:
I've tested in Chrome...
image

image

bdfranck:
I've tested in Firefox...
image

image

bdfranck:
I've tested in Safari...
image

image

bdfranck:
I also checked the error message and the font is correct...  
image

...but then I noticed that the icon isn't vertically aligned because the line height is shorter. 🤔 I'll fix that.

bdfranck:
image

I nudged up the icon with a transform. It seems to work with single and multi-line error messages. I'll check in other browsers too.

ArakTaiRoth:
@bdfranck A few issues that need to be resolved:

- [ ] Label size regular should be using `font: var(--goa-typography-heading-s);` instead of setting font-size, weight, and line-height separately (the line-height is currently wrong)
- [ ] Gap between error message and help text should be increased (.error-msg `gap: var(--goa-space-xs`))
- [ ] .form-item-input should have margin-bottom of `goa-space-xs`

bdfranck:
**Testing checklist:**

A form item has...
- [ ] `xs / 0.5rem` space between label and input
- [ ] `s / 0.75rem` space between input and help text
- [ ] `s / 0.75rem` space between input and error message
- [ ] `xs / 0.5rem` space between error message and help text

A form item with a large label has...
- [ ] `s / 0.75rem` space between label and input

bdfranck:
I noticed that the form item has a bottom margin of varied sizes depending on if we have an error message and help text. 

Here are the current production values:
- Input only: 2xs bottom margin
image
- Input & error: 2xs bottom margin
image
- Input & help text:  0 bottom margin
image

I chatted with @twjeffery and we agreed that there should never be a trailing bottom margin on the form input. It's easy for teams to add margins as needed.

bdfranck:
@ArakTaiRoth I've updated my PR with the following changes:
- Use `--goa-typography-heading-s` for label
- Use `--goa-typography-body-xs` for optional/required label
- Use correct spacing between input, error message, and help text
- Remove trailing bottom margin from input in all states (See above comment):
  - Input only
   - Input & error message
  - Input & help text
  - Input, error message, & help text.

chrisolsen (approved):

Issue #1703, opened 2024-03-12
Bug: Inconsistent behaviour between input components


ArakTaiRoth:
# Info

Reported by Francis Agyapong

Our Input and Textarea component has a `keyup` and a `change` event. Both fire on the same thing though, `onKeyUp`. Whereas with our Radio, and Checkbox components, the `change` event fires `onChange` instead of `onKeyUp`.

With our input component, this potentially creates an issue where validation occurs even if a user does nothing in an input and just tabs through it. But really, the bigger issue is the inconsistency, a user can tab through a whole form and only be alerted to errors in the Input component.

## Acceptance Criteria:

1. Make the behaviour consistent for `change` events between all our components. (probably modify `change` to fire `onChange`)

syedszeeshan:
Discussed with @chrisolsen , the change may result in "breaking change", so we might want to defer this till our major release?

ArakTaiRoth:
Closing this issue as it's no longer an issue. The instigating issue of this has been fixed previously according to Francis.

PR #1820, opened 2024-05-01
fix(#1703): input add onChange,modify Keyup,make TextArea consistent

syedszeeshan:
-For the input, modify the behavior of the KeyUp, add take the change event functionality out in a separate onChange handler. 
-Also, make the behavior of the TextArea consistent with this change wherein the keypress event isn't dealing with the change event anymore (so that both Textarea and input behave in similar way w.r.t. Keyup and Change).


ArakTaiRoth:
@syedszeeshan Problem that needs to be fixed:

- [ ] onChange and _change (React and Angular) are firing now when component loses focus, instead of firing when the value is changed

ArakTaiRoth:
Closing this PR as the story has been closed, this PR is no longer needed

Issue #1816, opened 2024-04-30
Allow GoAContainer to fit content width in addition to the existing 100% width


chrisolsen:
The `.goa-container` `flex` css can updated to allow for dynamically setting it to prevent flex growth.

```
  .goa-container {
    box-sizing: border-box;
    display: flex;
    flex: 0 0 auto;   /** here **/
    flex-direction: column;
  }
```

The these changes:
![image](https://github.com/GovAlta/ui-components/assets/41582/12629499-3ece-4800-9848-54d3ad41591d)

## Acceptance Criteria

1. New property that dictates how the container is sized (either full-width or to content)

twjeffery:
From this thread in the support channel: https://goa-dio.slack.com/archives/C02PLLT9HQ9/p1714490521425969

PR #1851, opened 2024-05-16
feat(#1816): width property for container

bdfranck:
This PR for https://github.com/GovAlta/ui-components/issues/1816 adds a width property to the Container component. It has two values:

| Name | Description | Screenshot |
|--------|--------|--------|
| Full | Expand to fit available horizontal width (default) | image |
| Content| Shrink to width of content. | image |

### DIFF
--- a/libs/react-components/src/lib/container/container.spec.tsx
+++ b/libs/react-components/src/lib/container/container.spec.tsx
@@ -10,6 +10,7 @@ describe("Container", () => {
         accent="thick"
         padding="relaxed"
         title={"Text title"}
+        width="content"
         mt="s"
         mr="m"
         mb="l"
@@ -30,6 +31,7 @@ describe("Container", () => {
     expect(el?.getAttribute("mr")).toBe("m");
     expect(el?.getAttribute("mb")).toBe("l");
     expect(el?.getAttribute("ml")).toBe("xl");
+    expect(el?.getAttribute("width")).toBe("content");
 
     expect(el?.querySelector("*[slot=title]")?.innerHTML).toContain("Text title");
     expect(
--- a/libs/react-components/src/lib/container/container.tsx
+++ b/libs/react-components/src/lib/container/container.tsx
@@ -10,11 +10,13 @@ export type GoAContainerType =
   | "important";
 export type GoAContainerAccent = "thick" | "thin" | "filled";
 export type GoAContainerPadding = "relaxed" | "compact";
+export type GoAContainerWidth = "full" | "content";
 
 interface WCProps extends Margins {
   type?: GoAContainerType;
   accent?: GoAContainerAccent;
   padding?: GoAContainerPadding;
+  width?: GoAContainerWidth;
 }
 
 declare global {
@@ -34,6 +36,7 @@ export interface GoAContainerProps extends Margins {
   padding?: GoAContainerPadding;
   actions?: ReactNode;
   children?: ReactNode;
+  width?: GoAContainerWidth;
   testId?: string;
 }
 
@@ -45,6 +48,7 @@ export function GoAContainer({
   children,
   actions,
   type,
+  width,
   mt,
   mr,
   mb,
@@ -57,6 +61,7 @@ export function GoAContainer({
       type={type}
       padding={padding}
       accent={accent}
+      width={width}
       mt={mt}
       mr={mr}
       mb={mb}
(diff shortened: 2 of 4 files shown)


bdfranck:
I updated the unit tests with the width property and they pass:
image

bdfranck:
I tested the changes on latest Chrome...
image

Latest Edge...
image

Latest Firefox...
image

Latest Safari...
image

lizhuomeng71:
When GoAAccordion inside <GoAContainer width="content">, this display is not correct

```
  <GoAContainer width="content">
  <GoAAccordion heading="Heading" headingSize="medium">
    Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi
  </GoAAccordion>
  </GoAContainer>
```

image

lizhuomeng71:
When GoAGrid inside , this display is not correct

```
  <GoAContainer width="content">
    <GoAGrid minChildWidth="12ch">
      <GoAFormItem
        label="Given name"
        id="givenName"
      >
        <GoAInputText name="givenName" maxLength={50} onChange={() => { }} />
      </GoAFormItem>
      <GoAFormItem
        label="Surname"
        id="surname"
      >
        <GoAInputText
          name="surname"
          onChange={() => { }}
          maxLength={50}
        />
      </GoAFormItem>
    </GoAGrid>
  </GoAContainer>
```

bdfranck:
@lizhuomeng71 I spent some time exploring why `goa-grid` didn't work as you expect.

When `goa-container` has a width of `content`, It collapses to the width of the child. For a `goa-grid`, this width is equal to a column's minimum width. The column width can either be...

1. The width specified in the `minChildWidth` property
2. `0px`, if no `minChildWidth` is set

In your above example, you set the `minChildWidth` to `12ch`. The amount is too narrow for the inputs, which have a default width of `30ch`. When I update `minChildWidth` to `30ch` then I see the following:

image

This is what I expect. If they want a two-column layout, they should wrap the `goa-grid` in an element with a set width.

Does that make sense?  I'm happy to chat about it if you have any questions.

lizhuomeng71:
Hi @bdfranck , thanks for the update, there is also another comment regarding GoAAccordion

bdfranck:
@lizhuomeng71 Yup! The problem is that the `.goa-accordion` width shrinks to `0` in this situation regardless of it's children. I'm still not sure why. 🤔 

image

Removing this line of CSS in `accordion` fixes the issue...

https://github.com/GovAlta/ui-components/blob/dc4066b2d67b0491558839683d73c8bb3586bb85/libs/web-components/src/components/accordion/Accordion.svelte#L121-L123

image

...but we need to keep that line for content queries in the Accordion component. I'll keep exploring.

bdfranck:
@lizhuomeng71 Here's my best explanation of what is going on:

Because `.goa-container` has `container-type: inline-size`, it [cannot be sized](https://stackoverflow.com/questions/73975889/container-query-collapses-width-of-element) by the width of its children. It effectively has a minimum width of 0.

You should see this issue when you put an accordion in a `flexbox` with `flex-shrink`. 

```html
<div style="display: flex; flex-shrink: 1;">
  <goa-accordion headingsize="small" heading="Accordion heading">
    This is some test content within the accordion.
  </goa-accordion>
</div>
``` 

For example, you can see this issue when you put an accordion in a `goa-block`:

```html
<goa-block>
  <goa-accordion headingsize="small" heading="Accordion heading">
    This is some test content within the accordion.
  </goa-accordion>
</goa-block>
``` 

image

I'm not sure if we want to fix this or just advise people to put the accordion in a sized element for these situations. I'll write a bug to start the discussion.

bdfranck:
@lizhuomeng71 I wrote a bug for the issue you discovered:

https://github.com/GovAlta/ui-components/issues/1864

Anything else?

lizhuomeng71:
Hi Team, issue is verified on Chrome and Firefox, issue raised is identified as non-issues, Move ticket to Ready for Review

chrisolsen (dismissed):

ArakTaiRoth (approved):

ArakTaiRoth (approved):

chrisolsen (approved):

Issue #1814, opened 2024-04-29
Form Item - error and helpertext converted to slots


Spark450:
## Background:

-[The use of italics is the GOA standard when referencing the names of legislation](https://www.alberta.ca/web-writing-style-guide-formatting#:~:text=bold%20text%20sparingly.-,Italics,lead%20of%20a%20web%20page.)
- In order to meet this standard we need to change our support of italics in the form item.

## Acceptance Criteria:

- `error` and `helpertext` properties also allow slot content
- Update the component documentation to mark properties as slots and make an example of this usage
- Defaults should continue to be "normal" for the form item.

<img width="907" alt="Screen Shot 2024-04-26 at 1 49 19 PM" src="https://github.com/GovAlta/ui-components/assets/51723535/63811c4f-f113-41bf-a109-15a5412044ea">

PR #1853, opened 2024-05-21
fix(#1814): form item slotted error and helptext

syedszeeshan:
Allow slotted content projection for error and helptext. But when slot isn't used, then continue to keep the existing fucntionality where the error and helptext show up as before in regular font type. Documentation changes will be in a separate PR.


lizhuomeng71:
Issue is verified on react and angular, chrome. firefox and edge

image
image

libs/web-components/src/components/form-item/FormItem.svelte
```diff
@@ -57,14 +57,22 @@
   <div class="form-item-input">
     <slot />
   </div>
-  {#if error}
+
```
chrisolsen, line 60: As I mentioned in the corresponding docs PR, the React component needs to accept the helpText as a `string | ReactNode`, then inject the prop into the appropriate slot.

syedszeeshan:
> Issue is verified on react and angular, chrome. firefox and edge
> 
> image image

Hi @lizhuomeng71 , just a quick comment. HTML is not supposed to be passed to the helptext and error proerties on <goa-form-item>, it will continue to be plain text. However, the html can be passed into the appropriate slots. It's either or. You either provide the plain text as before or use slot for custom html where italiziced text or any other custom html can be passed into.

libs/web-components/src/components/form-item/FormItem.svelte
syedszeeshan: This comment is resolved in the lastest code change. Tested in Angular and React playground.

Below is a quick reference for examples on the usage. 

This is the plain text example:
`<GoAFormItem label="First name" **helpText="This help text is in regular font plain-text"**>
  <GoAInput name="firstName" ..... ></GoAInput>
</GoAFormItem>`

OR

This is the slotted (custom html) example:
`<GoAFormItem label="First name (slotted)">
 <GoAInput name="firstName" ..... ></GoAInput>
  <div **slot="helptext"**>
    This is a <i>slotted</i><b>help text</b>.
  </div>
</GoAFormItem>`

Same goes for 'error' property.

syedszeeshan: This is the slotted (custom html) example:
`<GoAFormItem label="First name (slotted)">
 <GoAInput name="firstName" ..... ></GoAInput>
  <div slot="helptext">
    This is a <i>slotted</i><b>help text</b>.
  </div>
</GoAFormItem>`

Same goes for 'error' property.

libs/react-components/src/lib/form-item/form-item.tsx
```diff
@@ -7,8 +7,8 @@ interface WCProps extends Margins {
   label?: string;
   labelsize?: GoAFormItemLabelSize;
   requirement?: GoAFormItemRequirement;
-  error?: string;
-  helptext?: string;
+  error?: string | React.ReactNode;
+  helptext?: string | React.ReactNode;
```
chrisolsen, line 11: This could be simplified to just `ReactNode` since it includes the `string` type. I missed that when we were talking about it earlier.

syedszeeshan: Done.

chrisolsen (approved):

ArakTaiRoth (approved):

Issue #1872, opened 2024-05-28
Modal: Modal Overlay colour is incorrect


ArakTaiRoth:
# Current Modal Overlay

![image](https://github.com/GovAlta/ui-components/assets/1081991/f90a1045-ae53-41a2-953c-4336735e5ac1)

# Correct Modal Overlay

![image](https://github.com/GovAlta/ui-components/assets/1081991/671bd41b-3c6f-426d-b0b5-e7b41692d402)

## Acceptance Criteria:

1. Modal overlay background colour should be `#333` --> `--goa-color-greyscale-black`
2. Modal overlay background colour should have an opacity of `50%`

PR #1873, opened 2024-05-28
fix(#1872): Changing Modal overlay background colour

ArakTaiRoth:


bdfranck:
@ArakTaiRoth We could use the modal opacity design token here:
https://design.alberta.ca/design-tokens/opacity

lizhuomeng71:
issue is verified on firefox and chrome, for react and angular

image

chrisolsen (approved):

PR #1874, opened 2024-05-29
fix(#1814): allow form-item slotted content for helptext and error

syedszeeshan:
GoAFormItemProps also need to be updated...in the form-item.tsx

```
error?: React.ReactNode;
helpText?: React.ReactNode;

```


lizhuomeng71:
Issue is passed for react

image

ArakTaiRoth (approved):

Issue #1868, opened 2024-05-23
Accordion and Detail: When expanded the body content isn't read by screen readers


ArakTaiRoth:
# Info

A screen reader will happily read the `heading` property and anything in the `headingcontent` slot. It will also read the status of the accordion, open or closed. However, when an Accordion is opened, the body content is not read at all.

## Acceptance Criteria:

1. Body content in an accordion and detail components should be read by a screen reader when the Accordion is expanded

## More Info

Current working theory is to apply `aria-expanded` to the content `div` and set it to true or false based on the open and closed state.

PR #1877, opened 2024-05-30
fix(#1868): enhance accordion & details accessibility

vanessatran-ddi:
Ticket is https://github.com/GovAlta/ui-components/issues/1868
**Note: The AC "Body content in an accordion and detail components should be read by a screen reader when the Accordion is expanded" cannot be fulfilled because this is against what NVDA suggested.**

Reference: [Should Accordion Header expanding read Accordion Panel by default?](https://github.com/w3c/aria-practices/issues/1280)

What this ticket is doing:
* We will follow the standard practice at https://www.w3.org/WAI/ARIA/apg/patterns/accordion/examples/accordion/
* Also the asking to fix is https://goa-dio.slack.com/archives/C02PLLT9HQ9/p1716488793498059 mentioned as they need `aria-label`, `aria-labelledby`

Screenshot: 
image

image

### DIFF
--- a/libs/web-components/src/components/accordion/Accordion.svelte
+++ b/libs/web-components/src/components/accordion/Accordion.svelte
@@ -8,6 +8,7 @@
     typeValidator,
     toBoolean,
     validateRequired,
+    generateRandomId,
   } from "../../common/utils";
   import type { Spacing } from "../../common/styling";
 
@@ -24,6 +25,7 @@
   export let heading: string = "";
   export let secondarytext: string = "";
   export let headingsize: HeadingSize = "small";
+  export let id: string = "";
   export let testid: string = "";
   export let mt: Spacing = null;
   export let mr: Spacing = null;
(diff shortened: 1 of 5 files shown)


lizhuomeng71:
ariaLabel and ariaLabelledBy doesn't seems shows up for react

image

lizhuomeng71:
issue is verified on firefox and chrome, it will take id if id exist or heading if id not exist

image

ArakTaiRoth (dismissed):

libs/web-components/src/components/accordion/Accordion.svelte
```diff
@@ -39,6 +40,7 @@
   // Reactive
 
   $: isOpen = toBoolean(open);
+  $: accordionId = id || `${heading.replace(/\s+/g, "-")}`;
```
chrisolsen, line 43: Is there a reason we are creating a public prop for `id` and not just using a random number or just the heading?

chrisolsen (approved):

ArakTaiRoth (approved):

Issue #1515, opened 2023-12-12
Chip: Long Text


jim-muirhead:
### Is your feature request related to a problem? Please describe

Due to the limited screen width available on mobile phones and the lengthy text strings required by the chips used to represent the filters currently applied to a list of procurement opportunities and notices, we need a way to view each chip’s value(s) within a limited chip width.

### Describe the solution you'd like

See the Figma mockups.

https://www.figma.com/file/ITHqvarhTtYal257Z4JzCB/Share-with-DS-Team?type=design&node-id=1072-116681&mode=design&t=f1GBRRudhd4O9aj8-4

# Info

![image](https://github.com/GovAlta/ui-components/assets/82968683/cfc8a00a-53bd-4b91-8768-3482e24380d6)

## Acceptance Criteria

1. Chip is designed to work as it is in the second example in the provided Figma mockups, but close icon is top aligned instead of center

PR #1881, opened 2024-06-03
fix(#1515): handle long text in chip

vanessatran-ddi:
allow long text in chip
image

### DIFF
--- a/libs/web-components/src/components/chip/Chip.svelte
+++ b/libs/web-components/src/components/chip/Chip.svelte
@@ -91,8 +91,8 @@
     vertical-align: middle;
     align-items: center;
     background-color: var(--goa-color-greyscale-white);
-    border-radius: 99px;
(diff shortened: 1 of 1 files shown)


lizhuomeng71:
issue is verified on firefox and chrome
alpha

libs/web-components/src/components/chip/Chip.svelte
```diff
@@ -91,19 +91,21 @@
     vertical-align: middle;
     align-items: center;
     background-color: var(--goa-color-greyscale-white);
-    border-radius: 99px;
-    border: 1px solid #949494; /* TODO: change this to a defined color value when one is defined in the design specs */
+    border-radius: 16px;
```
chrisolsen, line 94: use `1rem` instead

ArakTaiRoth (approved):

Issue #1840, opened 2024-05-08
Input and Form Item - Linking Form Item to Input for Accessibility


ArakTaiRoth:
# Info

Right now we have an `arialabel` property for our form components. This is because currently our Form Item field isn't actually read by a screen reader, so we need some way to announce a form component and what it's intended for. This is not the recommended way of having accessibility for form elements. The W3C recommends having a label element (Form Item) with a `for` property that is the `id` property of the associated form element. The W3C only recommends using an `aria-label` property if visual users can ascertain the intent of the form element through context of where it's located on the site, and thus don't need a label.

If you look at Angular Material, they have an interesting way of dealing with this issue. They use a `label` element with a `for` attribute, targeting an auto generated id on the associated Input element.

Right now we don't even use a `label` element, we have a div with a class of label that contains the text.

**This ticket applies to all form "input" elements i.e. input, radio, dropdown, checkbox, textarea**

## Acceptance Criteria:

1. Investigate if we can:
    * Use a `label` element instead of a div for our Form Item component
    * Generate a random `id` for our Input component
    * Attach that randomly generated `id` to the `for` property for the above `label` element

PR #1888, opened 2024-06-05
fix(#1840): link form-item to Input for accessibility

syedszeeshan:
The parent element sets the aria-label in the child (Input) element, but only when aria-label wasn't supplied in the input element i.e. it will not be overwritten.

Note: We decided not to set the 'for' attribute of the <label> tag in the parent i.e. FormItem... to be same as the id of the input. Because setting aria-label is enough. Tested locally with Chrome and Firefox using NVDA.

Some example code for angular playground. The `<div>` is added b/c `<goa-input>` may or may not always be the immediate child of `<goa-form-item>`
```
<h2>Case 1a - No arialabel provided (Expected: aria-lable gets set)</h2>
<goa-form-item label="DDD Id Provided" helptext="Maximum characters allowed are 10.">
  <div>
    <goa-input name="inputNameDesignAlberta1" maxlength="10" id="abcd-1234"></goa-input>
  </div>
</goa-form-item>

<h2>Case 1b - aria-label NOT provided (Expected: aria-lable gets set irrespective of whether input element's Id was provided or not)</h2>
<goa-form-item label="DDD No Id" helptext="Maximum characters allowed are 10.">
  <div>
    <goa-input name="inputNameDesignAlberta2" maxlength="10"></goa-input>
  </div>
</goa-form-item>

<h2>Case 2 - Both the (label and arialabel) provided at the same time - (Expected: Aria-label in the child should be the winner)</h2>
<goa-form-item label="Some Label" helptext="Maximum characters allowed are 10.">
  <div>
    <goa-input name="inputNameDesignAlberta4" maxlength="10" arialabel="aria label from child element"></goa-input>
  </div>
</goa-form-item>
```

### DIFF
--- a/libs/web-components/src/components/form-item/FormItem.svelte
+++ b/libs/web-components/src/components/form-item/FormItem.svelte
@@ -1,6 +1,12 @@
 <svelte:options customElement="goa-form-item" />
 
 <!-- Script -->
+<script lang="ts" context="module">
+  export type FormItemChannelProps = {
+    el: HTMLElement;
+  };
+</script>
+
 <script lang="ts">
   import { onMount } from "svelte";
   import type { Spacing } from "../../common/styling";
@@ -36,23 +42,41 @@
   export let helptext: string = "";
   export let error: string = "";
   export let requirement: RequirementType = "";
-  export let id: string = "";
+  export let id: string = ""; // @deprecated: no longer used
+
+  let _rootEl: HTMLElement;
 
   onMount(() => {
     validateRequirementType(requirement);
     validateLabelSize(labelsize);
+
+    _rootEl?.addEventListener("input:mounted", handleInputMounted);
   });
+
+  function handleInputMounted(e: Event) {
+    const ce = e as CustomEvent<FormItemChannelProps>;
+
+    // Check if aria-label is present and has a value in the child element
+    const ariaLabel = ce.detail.el.getAttribute("aria-label");
+    if (!ariaLabel || ariaLabel.trim() === "") {
+      ce.detail.el.setAttribute("aria-label", label);
+    }
+  }
 </script>
 
 <!-- HTML -->
-<div data-testid={testid} style={calculateMargin(mt, mr, mb, ml)}>
+<div
+  data-testid={testid}
+  style={calculateMargin(mt, mr, mb, ml)}
+  bind:this={_rootEl}
+>
   {#if label}
-    <div class={`label ${labelsize}`} {id}>
+    <label class={`label ${labelsize}`}>
       {label}
       {#if requirement && REQUIREMENT_TYPES.includes(requirement)}
         <em>({requirement})</em>
       {/if}
-    </div>
+    </label>
   {/if}
   <div class="form-item-input">
     <slot />
--- a/libs/web-components/src/components/input/Input.spec.ts
+++ b/libs/web-components/src/components/input/Input.spec.ts
@@ -1,6 +1,7 @@
 import { render, fireEvent, waitFor, cleanup } from "@testing-library/svelte";
 import GoAInput from "./Input.svelte";
 import GoAInputWrapper from "./Input.test.svelte";
+import GoAInputFormItemWrapper from "./InputFormItemWrapper.test.svelte";
 import { it, describe } from "vitest";
 
 afterEach(cleanup);
@@ -8,7 +9,7 @@ afterEach(cleanup);
 describe("GoAInput Component", () => {
   it("should render", async () => {
     const el = render(GoAInput, { testid: "input-test", id: "test" });
-    const input = await el.findByTestId('input-test');
+    const input = await el.findByTestId("input-test");
     expect(input).toBeTruthy();
     expect(input.getAttribute("id")).toBe("test");
   });
@@ -107,14 +108,42 @@ describe("GoAInput Component", () => {
       expect(root).toBeTruthy();
     });
 
-    it("defaults to the name property if arialabel is not supplied", async () => {
-      const el = render(GoAInput, { testid: "input-test", name: "firstName" });
-      const root = el.container.querySelector('[aria-label="firstName"]');
-      expect(root).toBeTruthy();
+    it("should generate an aria-label value when not provided", async () => {
+      const result = render(GoAInputFormItemWrapper, {
+        inputId: "",
+        inputName: "inputName",
+        label: "DDD Alberta",
+        testIdFormItem: "formItem-testid",
+        testIdInput: "input-testid",
+      });
+      const inputEl = result.queryByTestId("input-testid");
+
+      await waitFor(() => {
+        expect(inputEl?.getAttribute("aria-label")).toBe("DDD Alberta");
+      });
+    });
+
+    it("shouldn't overwrite aria-label if a value has already been assigned", async () => {
+      const result = render(GoAInputFormItemWrapper, {
+        inputId: "inputId",
+        inputName: "inputName",
+        arialabel: "DONT OVERWRITE ME",
+        label: "DDD Alberta",
+        testIdFormItem: "formItem-testid",
+        testIdInput: "input-testid",
+      });
+      const inputEl = result.queryByTestId("input-testid");
+
+      await waitFor(() => {
+        expect(inputEl?.getAttribute("aria-label")).toBe("DONT OVERWRITE ME");
+      });
     });
 
     it("accepts an arialabelledby property", async () => {
-      const el = render(GoAInput, { testid: "input-test", arialabelledby: "firstName" });
+      const el = render(GoAInput, {
+        testid: "input-test",
+        arialabelledby: "firstName",
+      });
       const root = el.container.querySelector('[aria-labelledby="firstName"]');
       expect(root).toBeTruthy();
     });
@@ -177,7 +206,7 @@ describe("GoAInput Component", () => {
       keypress();
     });
 
-    await fireEvent.keyUp(input, { target: { value: "foobar" }, key: 'r' });
+    await fireEvent.keyUp(input, { target: { value: "foobar" }, key: "r" });
     await waitFor(() => {
       expect(change).toBeCalledTimes(1);
       expect(keypress).toBeCalledTimes(1);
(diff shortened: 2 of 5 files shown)


lizhuomeng71:
currently GoAInput have a property of ariaLabel and ariaLabelledBy, this will be overwrite by GoAFormItem, which is a break change, 

the logic should by if only GoAInput's ariaLabel  and ariaLabelledBy is not set , then GoAFormItem should apply

lizhuomeng71:
issue is verified for react and angular

image

libs/web-components/src/components/form-item/FormItem.svelte
```diff
@@ -38,21 +39,67 @@
...
+    slot?.addEventListener("slotchange", () => {
+      if (!_rootEl) return;
+
+      linkFormItemToInputForAccessibility();
+    });
+  });
+
+  function linkFormItemToInputForAccessibility() {
+    // Remove previous listener if it exists
+    _rootEl?.removeEventListener("input:mounted", inputMountedHandler);
```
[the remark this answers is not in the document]
syedszeeshan: Done.

libs/web-components/src/common/utils.ts
```diff
@@ -113,3 +113,7 @@ export function clamp(value: number, min: number, max: number): number {
       ? min
       : value
 }
+
+export function generateRandomId() {
+  return `input-${Math.random().toString(36).substr(2, 9)}`;
```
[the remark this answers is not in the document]
syedszeeshan: replaced with slice

libs/web-components/src/components/form-item/FormItem.svelte
```diff
@@ -38,21 +39,52 @@
...
+  let _rootEl: HTMLElement;
+  let _options: InputOption[] = [];
+
   onMount(() => {
     validateRequirementType(requirement);
     validateLabelSize(labelsize);
+
+    const slot = _rootEl.querySelector("slot");
+    slot?.addEventListener("slotchange", () => {
+      if (!_rootEl) return;
```
chrisolsen, line 51: This line is not needed since you use the `?` (optional chaining) operator

libs/web-components/src/components/input/Input.spec.ts
```diff
@@ -107,9 +107,9 @@ describe("GoAInput Component", () => {
       expect(root).toBeTruthy();
     });
```
chrisolsen, line 109: There should be a test for this verifying that the event is being raised and/or a test with the form-item wrapping the input that verifies they `for` attribute is being set.

libs/web-components/src/components/input/Input.svelte
```diff
@@ -2,6 +2,10 @@
 
 <script lang="ts" context="module">
   export type GoAInputVariant = "goa" | "bare";
+
+  export type InputOption = {
```
chrisolsen, line 6: Since the form item will be talking to a number of other form elements, let's define this type in the FormItem component, using a more generic name like `FormItemChannelProps`.

libs/web-components/src/components/form-item/FormItem.svelte
```diff
@@ -6,6 +6,7 @@
   import type { Spacing } from "../../common/styling";
   import { calculateMargin } from "../../common/styling";
   import { typeValidator } from "../../common/utils";
+  import type { InputOption } from "../input/Input.svelte";
```
chrisolsen, line 9: See my comment with the Input.svelte file on the change required here.

libs/web-components/src/components/input/Input.spec.ts
syedszeeshan: Test created, event issue resolved. Will be adding few more tests......in the next commit.

libs/web-components/src/components/form-item/FormItem.svelte
```diff
@@ -38,21 +39,52 @@
...
   onMount(() => {
     validateRequirementType(requirement);
     validateLabelSize(labelsize);
+
+    _rootEl?.addEventListener("input:mounted", handleInputMounted);
+
+    const slot = _rootEl.querySelector("slot");
+    slot?.addEventListener("slotchange", () => {
+      _rootEl?.addEventListener("input:mounted", handleInputMounted);
+    });
```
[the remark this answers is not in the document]
syedszeeshan: I think we can do without lines 51-54.... it was to handle a situation where the user makes some changes e.g. programatically updates the id of the input directly and/or remove/modify the aria-label....etc. But I guess in that case the onus is on the user to set aria-label, etc... and may be an intended action from the user. I am gonna remove and update the PR.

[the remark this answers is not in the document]
syedszeeshan: If the user does something that triggers the slot change, we dont need to re-run this logic b/c if the user is changing the id and or aria-label etc then may be that's their inention. and its upto them.

Code updated, resolving this comment.

libs/web-components/src/components/form-item/FormItem.svelte
```diff
@@ -38,21 +39,47 @@
   export let requirement: RequirementType = "";
   export let id: string = "";
 
+  let _rootEl: HTMLElement;
+  let _options: FormItemChannelProps[] = [];
```
chrisolsen, line 43: This is probably a mix up from looking at the dropdown component's code. In the case of a dropdown the `_options` array is holding references to each of the options to allow for later deselecting of them when a different option is selected.

In this case the `_options` are being set, but never referenced afterwards, I would think that VSCode would be showing a lint error for this. 

Since, the currently logic isn't tuned to handle more than one child event the `_options` can be removed.

libs/web-components/src/components/input/Input.spec.ts
```diff
@@ -107,14 +108,73 @@ describe("GoAInput Component", () => {
...
+      const result = render(GoAInputFormItemWrapper, {
+        inputId: "",
+        inputName: "inputName",
+        label: "DDD Alberta",
+        testIdFormItem: "formItem-testid",
+        testIdInput: "input-testid",
+      });
+      const inputEl = result.queryByTestId("input-testid");
+      expect(inputEl?.getAttribute("id")).not.toBe("");
+    });
```
chrisolsen, line 139: This test is close to duplicating the above test, so it can be removed.

libs/web-components/src/components/input/Input.spec.ts
```diff
@@ -107,14 +108,73 @@ describe("GoAInput Component", () => {
...
+        inputName: "inputName",
+        arialabel: "DONT OVERWRITE ME",
+        label: "DDD Alberta",
+        testIdFormItem: "formItem-testid",
+        testIdInput: "input-testid",
+      });
+      const inputEl = result.queryByTestId("input-testid");
+
+      await waitFor(() => {
+        expect(inputEl?.getAttribute("aria-label")).not.toBe("DDD Alberta");
```
chrisolsen, line 168: No need for this check

libs/web-components/src/components/input/Input.svelte
```diff
@@ -2,6 +2,10 @@
 
 <script lang="ts" context="module">
   export type GoAInputVariant = "goa" | "bare";
+
+  export type FormItemChannelProps = {
+    el: HTMLElement;
+  };
```
chrisolsen, line 8: As mentioned, this should be moved into the FormItem component.

libs/web-components/src/components/form-item/FormItem.svelte
```diff
@@ -38,21 +44,44 @@
...
 <!-- HTML -->
-<div data-testid={testid} style={calculateMargin(mt, mr, mb, ml)}>
+<div
+  data-testid={testid}
+  style={calculateMargin(mt, mr, mb, ml)}
+  bind:this={_rootEl}
+>
   {#if label}
-    <div class={`label ${labelsize}`} {id}>
+    <label for={id} class={`label ${labelsize}`} {id}>
```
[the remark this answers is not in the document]
syedszeeshan: As discussed, I have removed the id and also the for attribute. The 'for' will be set later inside the event handler.

lizhuomeng71:
Verified with the latest change for the label for

libs/web-components/src/common/utils.ts
```diff
@@ -115,5 +115,5 @@ export function clamp(value: number, min: number, max: number): number {
 }
 
 export function generateRandomId() {
-  return `${Math.random().toString(36).substring(2, 9)}`;
```
chrisolsen, line 118: Since this is no longer being used this change can be removed, and taking a substring starting at index `6` doesn't make sense.

libs/web-components/src/components/input/Input.spec.ts
```diff
@@ -107,14 +108,42 @@ describe("GoAInput Component", () => {
       expect(root).toBeTruthy();
     });
 
-    it("defaults to the name property if arialabel is not supplied", async () => {
-      const el = render(GoAInput, { testid: "input-test", name: "firstName" });
-      const root = el.container.querySelector('[aria-label="firstName"]');
-      expect(root).toBeTruthy();
+    it("Input's aria-label gets a value, when not provided", async () => {
```
chrisolsen, line 111: The tests should read as a sentence, so in this case it should be
```
it("should generate an aria-label value when not provided", ...
```

The same is true for the other specs below.

libs/web-components/src/components/form-item/FormItem.svelte
```diff
@@ -38,21 +44,39 @@
   export let requirement: RequirementType = "";
   export let id: string = "";
```
chrisolsen, line 45: Can you mark this as `@deprecated` since it is no longer being used within the code.

libs/web-components/src/common/utils.ts
syedszeeshan: The generateRandomId() method is in use by couple fo other components. I have revereted my change b/c its not needed.

libs/web-components/src/components/form-item/FormItem.svelte
syedszeeshan: Done.

```
/**
   * @deprecated: `requirement` prop is deprecated.
   */
  export let requirement: RequirementType = "";
```

chrisolsen: It is the `id` that is deprecated. And to keep things readable just use the `//` comments to the right of the prop

syedszeeshan: Code and PR updated, thanks.

chrisolsen (approved):

lizhuomeng71:
Verified on the new pull request

Issue #1887, opened 2024-06-05
Previous change needs to be reverted back


chrisolsen:
FormItem.tsx
![image](https://github.com/GovAlta/ui-components/assets/41582/1d01d75c-1616-4f4b-bd9f-f5487d9d3373)

PR #1889, opened 2024-06-05
fix(#1887): Reverting error and helptext type change

ArakTaiRoth:
This means we need to test story #1814 again with this change


lizhuomeng71:
issue verified on firefox and angular

image
image

chrisolsen (approved):

Issue #1867, opened 2024-05-23
Block: 100% width of inputs not working


twjeffery:
To replicate: see example on design system website: https://design.alberta.ca/components/button#tab-0:~:text=Ask%20a%20user%20for%20an%20address,-Street%20Address
<img width="895" alt="image" src="https://github.com/GovAlta/ui-components/assets/82968683/3df05267-86ce-4731-9fa9-eeeb0302c151">

## Acceptance Criteria:
Inputs should go 100% within block when set to 100%

PR #1893, opened 2024-06-06
fix(#1867): allow child element 100% width inside block

vanessatran-ddi:
allow `input` 100% inside `block`

Example code: 
```
<goa-block gap="xl" direction="column">
  <goa-container>Test</goa-container>
  <goa-form-item label="Phone number">
    <goa-input width="100%" (_change)="onChange($event)" [value]="value" name="phone">
      <div slot="leadingContent">
        <goa-block direction="row" alignment="center" gap="none">
          <goa-icon type="add" size="small"></goa-icon> <span class="medium-text">1</span>
        </goa-block>
      </div>
    </goa-input>
  </goa-form-item>
  </goa-block>
```

For the `ui-components-docs`:

Before fix: 
image

After fix: 
image

### DIFF
--- a/libs/web-components/src/components/block/Block.svelte
+++ b/libs/web-components/src/components/block/Block.svelte
@@ -6,14 +6,16 @@
 
   export let gap: Spacing = "m";
   export let direction: "row" | "column" = "row";
-  export let alignment: "center" | "start" | "end" = "start";
+  export let alignment: "center" | "start" | "end" | "normal" = "normal";
 
   $: _alignment =
     alignment === "start"
       ? "flex-start"
       : alignment === "end"
         ? "flex-end"
-        : "center";
+        : alignment === "center"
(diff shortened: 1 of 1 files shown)


syedszeeshan (approved):

lizhuomeng71:
When block sent to  alignment="center" the 100% width will not take affect
image

There is also an edge case, 
When block sent to  direction="row", and there is multiple input, if will not fill the entire lenght

image

lizhuomeng71:
The issue has been verified on both React and Angular.

Note to Reviewer: This fix is only applied to the following cases:

Input width set to 100%
Block direction set to column
Block alignment not set
For all other cases, the behavior remains the same as in the current main branch.

chrisolsen (approved):

ArakTaiRoth (approved):

Issue #1892, opened 2024-06-06
Callout: Component fixes


Spark450:
### Describe the problem

**Acceptance criteria:**

- line height is incorrect in development on heading when stack to second line. should be coming from heading medium token 
(--goa-typography-heading-m)
(--got-line-height-4)

PR #1904, opened 2024-06-07
fix(#1892): Callout Header style change

ArakTaiRoth:


lizhuomeng71:
issue is verified on react and angular

image
image
Notice, these 2 styles are only apply to size="large"
(--goa-typography-heading-m)
(--got-line-height-4)

when size="medium", these style is used
--goa-typography-heading-xs
line-height = 24px

chrisolsen (approved):

Issue #1957, opened 2024-06-24
Date Picker: Year selection not appearing


ArakTaiRoth:
# Info

The year selection on the Datepicker currently isn't showing up.
![image](https://github.com/GovAlta/ui-components/assets/1081991/da98ce82-8611-4d7c-9a4e-f80628fdc8f1)

This is currently happening in both Alpha and Production

## Acceptance Criteria:

1. Year selection on DatePicker should appear

PR #1965, opened 2024-06-25
fix(#1957): date picker year selection not appearing

syedszeeshan:
# Before (the change)
image

# After (the change)
image

### DIFF
--- a/libs/web-components/src/components/dropdown/Dropdown.svelte
+++ b/libs/web-components/src/components/dropdown/Dropdown.svelte
@@ -554,7 +554,7 @@
             `}
           data-testid="input"
           bind:this={_inputEl}
-          value={_selectedOption?.label ?? _selectedOption?.value ?? ""}
+          value={_selectedOption?.label || _selectedOption?.value || ""}
           type="text"
           role="combobox"
           autocomplete="off"


libs/web-components/src/components/calendar/Calendar.svelte
```diff
@@ -319,7 +319,7 @@
         on:_change={setYear}
       >
         {#each _years as year}
-          <goa-dropdown-item value={year} />
+          <goa-dropdown-item value={year} label={year} />
```
chrisolsen, line 322: Because the `value` of the `input` falls back to the `value` prop, I am not 100% sure this fixes the issue.

syedszeeshan: It does fix the issue as you can see in the before and after images. I will connect with you on slack.

syedszeeshan: @chrisolsen 
I ran couple of checks and turns out that the label is an empty string that's why the 'value' isn't used...I am modifying this in the Dropdown comp...hope this is not considerd a breaking change

and do we still want to explicitly set the label in the year dropdown in the Calendar component?

libs/web-components/src/components/dropdown/Dropdown.svelte
```diff
@@ -564,7 +566,7 @@
             `}
             data-testid="input"
             bind:this={_inputEl}
-            value={_selectedOption?.label ?? _selectedOption?.value ?? ""}
+            value={(_selectedOption?.label && _selectedOption?.label !== '') ? _selectedOption.label : (_selectedOption?.value ?? "")}
```
chrisolsen, line 569: This seems redundant. if `_selectedOption?.label` is truthy then `_selectedOption?.label` it is not equal to an empty string.

ArakTaiRoth (approved):

chrisolsen (dismissed):

syedszeeshan:
image

chrisolsen (approved):

Issue #1954, opened 2024-06-19
Dropdown onChange event is firing when re-selecting the current option


chrisolsen:
# Info

Dropdown `onChange` event is firing even if the currently selected option is clicked again.

## Acceptance Criteria

1. `onChange` event only fires when the actual value is changed, not when anything is clicked

![image](https://github.com/GovAlta/ui-components/assets/41582/991a0529-a8ab-43ea-8122-b42abfec4358)

chrisolsen:
An existing issue that was found will also be fixed in this PR.
https://jam.dev/c/b4610511-3917-4e65-934c-ee5934cb72c2

PR #1971, opened 2024-06-28
fix(#1954): prevent dropdown firing _change event  when the value is reselected

chrisolsen:
# Before (the change)
https://jam.dev/c/9a8ae231-c3d9-4762-bff1-34ee4c03efe8

# After (the change)
https://jam.dev/c/8db31b77-a73a-4fe8-9c69-c0ede712b240

## Steps needed to test

- [ ] For non-filterable dropdown validate that when selecting the same value repeatedly, that events are only fired the first time.
- [ ] For the filterable dropdown validate that the when clearing the selection that events are fired with a blank value and that an event is fired when you delete the filter text and Tab away


lizhuomeng71:
Issue is verified on angular and react, for autocompleted dropdown and filter

ArakTaiRoth (approved):

PR #2045, opened 2024-07-31
[Major release] Angular wrappers and React name changed

vanessatran-ddi:
I created this drafted pull request to compare/update with `alpha` branch.


.github/workflows/pull-request.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       - run: npm ci
       # Run with the target branch as the base.
       - name: Lint
-        run: npx nx affected --target=lint --base=origin/${{ github.base_ref }}
+        run: npm run lint
```
chrisolsen, line 28: What is the reason for this change? I can't remember exactly why things were changed to the previous code without looking back through the history, but there was a reason :)

chrisolsen (commented):
Can you shorten your breaking change commit messages using the following format
```
fix(#999): [some commit message]

BREAKING CHANGE: [what is causing the breaking change]
```

package.json
```diff
@@ -42,35 +49,42 @@
...
+    "jest": "^29.4.1",
+    "jest-environment-jsdom": "^29.4.1",
+    "jest-preset-angular": "~14.1.1",
     "jsdom": "~22.1.0",
     "jsonc-eslint-parser": "^2.1.0",
+    "postcss": "^8.4.5",
     "ng-packagr": "18.0.0",
-    "nx": "19.3.0",
-    "postcss": "8.4.38",
+    "nx": "^19.3.1",
```
chrisolsen, line 87: Any nx version changes should/have to be done via running the `nx migrate`

chrisolsen (dismissed):

libs/web-components/src/components/checkbox/Checkbox.svelte
```diff
@@ -184,9 +184,7 @@
       {/if}
     </div>
     <div class="text" data-testid="text">
-      <slot>
         {text}
-      </slot>
```
chrisolsen, line 189: This could be a breaking change

chrisolsen (approved):

Issue #2075, opened 2024-08-21
Pagination: id of root being duplicated


PaigeDDD:
# Info

Issue raised by Andrew Filipe Naves de Barros

In our Pagination component we have a `<goa-block>` using an `id="root"`. This could potentially cause issues with React setups that is also used automatically on the main `<div id="root">` in the app. Semantic HTML states only one component can have a distinct `id` value, and they should never be shared.

This could be a bigger issue for people trying an SSR implementation. And he's personally seen issues where click events don't work as expected on other elements, because it's not sure where to assign them sometimes.

## Acceptance Criteria

1. `<goa-block>` inside Pagination has it's `id` property changed from "root" to something else

ArakTaiRoth:
I can't find anything about this happening with our Checkbox component. We have a class of "root" in our Checkbox, but no id of "root". We do however have `<goa-block id="root">` in our Pagination component, so this is indeed correct. I don't know how much of a problem this actually is though. Wil reach out to @PaigeDDD tomorrow to get some answers.

ArakTaiRoth:
Spoke with him, updated the story as needed.

chrisolsen:
Our id of `root` is within the shadow root, so it will not cause issue where id selectors. This is not an issue.

lizhuomeng71:
issue is verified on react and angular

ArakTaiRoth:
Closing this story for @chrisolsen 's comment above

PR #2163, opened 2024-09-25
fix(#2075): Changed Pagination block ID

ArakTaiRoth:
# Before

`id` property of `goa-block` element was "root"

# After

`id` property of `goa-block` element is "goa-pagination"

## Additional

Made a separate commit to add a comment to each package, so that they release on next push to main


lizhuomeng71:
issue is verified on react and angular

chrisolsen:
See my comment in the original ticket.

ArakTaiRoth:
Closing this PR as this story is not necessary

PR #2174, opened 2024-10-03
Feature/2162 & 2177 app header - privacy portal components + app header to trigger fly out menu

vanessatran-ddi:
# Before (the change)
If we want to have the look for those below 5 components:
* Accordion
* Button
* Callout
* Container
* Table
* SideMenuGroup
Like this figma design: https://www.figma.com/design/NRFUYjCwVoI9qMbFEG6s7J/Component-tokens?node-id=0-1&node-type=canvas&m=dev

It is impossible to override CSS right now, and if we override `--goa-text-default-color` for example, we will affect other components. 
This pull request is to add more specific design token variable names for those above 5 components, at the same time assigning values for those adding tokens, so it won't affect our components, but at the same time, allow other teams (specifically Privacy Portal Team) to override variable design token CSS (For example: https://github.com/vanessatran-ddi/latest-react-demo/blob/privacy-portal-components/src/override.css) 

# After (the change)
They can change the CSS variable token values on their app, example: https://vanessatran-ddi.github.io/latest-react-demo/

## Steps needed to test
The below is how we can test: @lizhuomeng71 :

Step 1:  Build the project: `npm run build:all`

Step 2: Check The 6 above components to see if their looks are different with existing components. We don't want to have anything different. 
(You can check with `ui-component-docs` by `npm link`)

Accordion: 
image

New attribute:
* iconPosition
image

Button: (remain same)
image

Callout: (style remains the same but has a new attribute `iconTheme`)
image

Container (remains the same)
image

Side Menu Group (remains the same with a new attribute `icon`)
image

(notice: for padding issue it has a PR https://github.com/GovAlta/design-tokens/pull/54 to fix)

AppHeader (styles remain the same)
image

If `onMenuClick` function is passed, then when it is on Tablet/Mobile, clicking the burger menu won't show up the AppHeaderMenu, but trigger the function `handleMenuClick`

### DIFF
--- a/libs/react-components/src/lib/accordion/accordion.tsx
+++ b/libs/react-components/src/lib/accordion/accordion.tsx
@@ -2,6 +2,7 @@ import { ReactNode } from "react";
 import { Margins } from "../../common/styling";
 
 export type GoAHeadingSize = "small" | "medium";
+export type GoAIconPosition = "left" | "right";
 
 interface WCProps extends Margins {
   open?: boolean;
@@ -11,6 +12,7 @@ interface WCProps extends Margins {
   headingContent?: ReactNode;
   maxwidth?: string;
   testid?: string;
+  iconposition?: GoAIconPosition;
 }
 
 declare global {
@@ -30,6 +32,7 @@ export interface GoAAccordionProps extends Margins {
   headingContent?: ReactNode;
   maxWidth?: string;
   testid?: string;
+  iconPosition?: GoAIconPosition;
   children: ReactNode;
 }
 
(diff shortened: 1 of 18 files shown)


lizhuomeng71:
https://jam.dev/c/e1428fa0-38ed-463c-b228-92416f4ecb66

I noticed something for the button font size is different with this PR, please see above jaber

libs/react-components/src/lib/accordion/accordion.tsx
```diff
@@ -11,6 +12,7 @@ interface WCProps extends Margins {
   headingContent?: ReactNode;
   maxwidth?: string;
   testid?: string;
+  iconposition?: string;
```
chrisolsen, line 15: This should also have the GoAIconPosition  type

chrisolsen (approved):

ArakTaiRoth (approved):

Issue #1973, opened 2024-06-28
Side Menu: No active link on page load


ArakTaiRoth:
# Info

This code:
```html
<goa-side-menu>
  <a routerLink="/accordion"> Accordion </a>
  <a routerLink="/button"> Button </a>
  <a routerLink="/callout"> Callout </a>
  <a routerLink="/checkbox"> Checkbox </a>
  <a routerLink="/chip"> Chip </a>
  <a routerLink="/container"> Container </a>
</goa-side-menu>
```
Reproduces the issue. If you click a link on the side menu, it becomes the active link as is proper. However if you load a page via URL, or reload a page, the active link doesn't show up. If you add `goa-side-menu-group` to the code, like this:
```html
<goa-side-menu>
  <goa-side-menu-group heading="Components">
    <a routerLink="/accordion"> Accordion </a>
    <a routerLink="/button"> Button </a>
    <a routerLink="/callout"> Callout </a>
    <a routerLink="/checkbox"> Checkbox </a>
    <a routerLink="/chip"> Chip </a>
    <a routerLink="/container"> Container </a>
  </goa-side-menu-group>
</goa-side-menu>
```
Then the active links work properly on reload, loading via URL, or loading by clicking a link.

## Acceptance Criteria:

1. Active side menu link should be shown even if not using `goa-side-menu-group`

PR #2176, opened 2024-10-03
fix(#1973): side menu active link issues

syedszeeshan:
# Before (the change)

# After (the change)

### DIFF
--- a/libs/web-components/src/components/side-menu/SideMenu.svelte
+++ b/libs/web-components/src/components/side-menu/SideMenu.svelte
@@ -16,6 +16,7 @@
   onMount(async () => {
     await tick();
     getChildren();
+    setCurrentUrl();
     addEventListeners();
   });
 
@@ -34,6 +35,7 @@
       .filter((el) => el.tagName === "A")
       .map((el) => {
         el.classList.remove("current");
+        el.addEventListener("click", setCurrentUrl);
         return el;
       });
 
@@ -96,12 +98,6 @@
       observer = null;
     }
     window.removeEventListener("popstate", setCurrentUrl);
-    if (_rootEl) {
-      _rootEl.removeEventListener(
-        "sidemenugroup:mounted",
-        handleSideMenuGroupMount,
-      );
-    }
   }
 </script>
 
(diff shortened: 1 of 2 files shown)


lizhuomeng71:
This issue is verified for react and angular

libs/web-components/src/components/side-menu/SideMenu.svelte
```diff
@@ -30,14 +30,28 @@
 
     if (slotChildren.length === 0) return;
 
+    updateSideMenuLinks(slotChildren);
+
+    observer = new MutationObserver((mutations) => {
```
chrisolsen, line 35: Since links aren't continually being added throughout it's lifetime, is an observer necessary? This seems to be a place that the, now commonly used pattern, of receiving an event from child nodes would work best.

syedszeeshan: updated the PR.

libs/web-components/src/components/side-menu/SideMenu.svelte
```diff
@@ -46,6 +48,10 @@
     setCurrentUrl();
   }
 
+  function handleLinkClick() {
+    setCurrentUrl();
+  }
+
```
chrisolsen, line 54: Instead of creating a new function just call the `setCurrentUrl()` directly.

libs/web-components/src/components/side-menu/SideMenu.svelte
```diff
@@ -102,6 +108,9 @@
         handleSideMenuGroupMount,
       );
     }
+    _sideMenuLinks.forEach((el) => {
+      el.removeEventListener("click", handleLinkClick);
+    });
```
chrisolsen, line 113: This isn't really needed, since the event listeners are attached to the element being removed. The same is true for lines 105-110.

libs/web-components/src/components/side-menu/side-menu.spec.ts
```diff
@@ -20,3 +20,30 @@ describe.skip("SideMenu should render with children and set highlighted menu ite
...
+    window.location = new URL("http://localhost/get-started/designers") as any;
+  });
+
+  afterEach(() => {
+    window.location = originalLocation;
+  });
+
+  it("should set active link correctly", async () => {
+    const { container } = render(SideMenuWrapper);
+    await new Promise((resolve) => setTimeout(resolve, 0));
```
chrisolsen, line 39: The testing library has a `waitFor` function, or you could use the `tick` function.

chrisolsen (approved):

ArakTaiRoth (approved):

Issue #2214, opened 2024-10-24
Radio: minor visual alignment issue


twjeffery:
### Describe the bug

Minor issue: the text in the radio item is not visually centered with the radio circle. Text is slightly below center.

Current
<img width="288" alt="image" src="https://github.com/user-attachments/assets/7b2ec11c-eea1-404c-a231-4b7318c5bb70">

Corrected
<img width="288" alt="image" src="https://github.com/user-attachments/assets/4931483c-f28d-49b4-b26e-b5e816a87250">

### To Reproduce

https://design.alberta.ca/components/radio#tab-0

### Additional context

Potentially fixed here: https://github.com/GovAlta/ui-components/pull/2215

PR #2215, opened 2024-10-24
fix: visual design issues on radio item

twjeffery:
Fix for https://github.com/GovAlta/ui-components/issues/2214

# Before (the change)
image

# After (the change)
image


lizhuomeng71:
This issue is verified on react and angular

image

Issue #2272, opened 2024-11-20
Privacy Portal component updates: Stepper progress bar and stepper step hover treatment


Spark450:
## Acceptance Criteria:

- Based on the need to show a complete status on the final step of the stepper (through a 5th hidden step), update the progress bar so that it accurately underlines the bounds of each stepper step
- Stepper steps should no longer be interactive, remove hover treatment from all stepper steps

Spark450:
@lizhuomeng71 This issue does not need to be tested, only code reviewed by @chrisolsen .

PR #2274, opened 2024-11-21
fix(#2272): pp form stepper, allow final step excluded from progress …

vanessatran-ddi:
Allow pp team to have the hidden 5th step as the Confirmation message 

Hidden step: https://github.com/vanessatran-ddi/latest-react-demo/blob/privacy-portal-components/src/components/2175/form-stepper.tsx#L52

How it is like: 
https://vanessatran-ddi.github.io/latest-react-demo/

### DIFF
--- a/libs/web-components/src/experimental/ExperimentalFormStep.svelte
+++ b/libs/web-components/src/experimental/ExperimentalFormStep.svelte
@@ -50,15 +50,12 @@
     // receive notification from parent of resize
     _rootEl.addEventListener("form-stepper:resized", (e: Event) => {
       const { mobile } = (e as CustomEvent).detail;
-      console.log("form-stepper:resized", mobile);
       _isMobile = mobile;
     });
 
     // receive parent el information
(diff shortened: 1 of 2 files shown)


syedszeeshan (dismissed):

libs/web-components/src/experimental/ExperimentalFormStep.svelte
```diff
@@ -183,20 +180,20 @@
...
-
-  label:not([aria-disabled="true"]):not([aria-current="step"]):hover {
-    background-color: rgba(0, 0, 0, 0.05);
-    cursor: pointer;
-  }
-  label.mobile:not([aria-disabled="true"]):not([aria-current="step"]):hover {
-    background-color: rgb(232 232 232);
-    cursor: pointer;
-  }
+  /*label:not([aria-disabled="true"]):not([aria-current="step"]):has(*/
```
ArakTaiRoth, line 183: If we don't need this stuff, maybe remove it instead of commenting it out?

syedszeeshan (approved):

PR #2286, opened 2024-12-02
Vite config for prod only minification

chrisolsen:
This change will allow for one to step through the code within the inspector, but still publish a minified file to npm.

## Steps needed to test

1. Compile the web-components lib via `npx nx reset; npm run build` and note the file size (~760KB).
2. Create a `.env` file and add `VITE_DEV=true`
3. Recompile the project `npx nx reset; npm run build`. The new size of the build would be ~1.2MB


BumbleB2na (approved):
Can confirm the compiled size changes: ui-components\dist\libs\web-components

Issue #2211, opened 2024-10-22
Dropdown: Can't select using keyboard on filterable


ArakTaiRoth:
# Info

Reported by Sergey Fedyanin

## Expected Output

Using a keyboard to navigate through filtered options, and then hitting enter, I expect the highlighted item to be the selected item in the Dropdown

## Actual Output

Using a keyboard to navigate through filtered options, and then hitting enter, the selected option is actually not the highlighted item.

## Replication Steps

1. Using a Dropdown with the `filterable` property set to `true`
2. Type to filter the suggestions down
3. Use the down arrow key on your keyboard to select an item
4. Hit enter on the item you want to select

[jam.dev](https://jam.dev/c/ff52d630-ae4a-4316-965a-021611dab02d)

## Acceptance Criteria:

1. The item highlighted using the arrow keys when the enter button is hit, should be the selected item in the Dropdown

PR #2303, opened 2024-12-10
fix(#2211): filterable dropdown selected option

BumbleB2na:
# Before (the change)

Filterable dropdown selected option broken when a) clicking an option to select, b) using arrows keys to highlight an option then pressing Enter to select, and c) when filterable dropdown input loses focus. Also, selected option in the DOM would not have "aria-selected=true".

# After (the change)

Filterable dropdown selected option fixed when a) clicking an option to select, b) using arrows keys to highlight an option then pressing Enter to select, and c) when filterable dropdown input loses focus. Also, selected option in the DOM has "aria-selected=true".

## Steps needed to test

Testing considerations: Tested in Chrome, Firefox

These Autofill Dropdown examples may be helpful for svelte, react and angular: https://github.com/user-attachments/files/19014103/2025-02-26-playground-examples.zip

### DIFF
--- a/libs/web-components/src/components/dropdown/Dropdown.spec.ts
+++ b/libs/web-components/src/components/dropdown/Dropdown.spec.ts
@@ -3,6 +3,14 @@ import GoADropdown from "./Dropdown.svelte";
 import GoADropdownWrapper from "./DropdownWrapper.test.svelte";
 import { describe, it } from "vitest";
 import { tick } from "svelte";
+import userEvent from "@testing-library/user-event";
+import type { UserEvent } from "@testing-library/user-event/dist/types/setup/setup";
+
+let user: UserEvent;
+
+beforeEach(() => {
+  user = userEvent.setup();
+});
 
 afterEach(() => {
   cleanup();
@@ -85,7 +93,7 @@ describe("GoADropdown", () => {
       });
     });
 
-    it.skip("should render a filterable dropdown", async () => {
+    it("should render a filterable dropdown", async () => {
       const result = render(GoADropdownWrapper, {
         name,
         value: "orange",
@@ -96,7 +104,7 @@ describe("GoADropdown", () => {
       const dropdown = result.queryByTestId("favcolor-dropdown");
       const popover = result.container.querySelector("goa-popover");
       const inputField = dropdown?.querySelector("input");
-      const dropdownIcon = result.container.querySelector("goa-icon");
+      let dropdownIcon = result.container.querySelector("goa-icon");
 
       expect(popover).toBeTruthy();
       expect(inputField).toBeTruthy();
@@ -134,7 +142,7 @@ describe("GoADropdown", () => {
         expect(ul?.getAttribute("role")).toBe("listbox");
         expect(ul?.getAttribute("style")).toContain("max-height: 276px"); // default height
         expect(ul?.getAttribute("tabindex")).toBe("-1");
-        expect(ul?.querySelectorAll("li").length).toBe(1);
+        expect(ul?.querySelectorAll("li").length).toBe(items.length);
 
         // Check options
         const option = result.container.querySelector("li#orange");
@@ -148,11 +156,18 @@ describe("GoADropdown", () => {
         expect(option).toHaveTextContent("orange");
       });
 
-      // show menu
-      inputField && (await fireEvent.click(inputField));
+      // show menu by clearing selected value
+      let clearIcon = result.queryByTestId("clear-icon");
+      expect(clearIcon).toBeTruthy();
+      clearIcon && (await user.click(clearIcon));
       await waitFor(async () => {
-        await tick();
+        const popover = result.container.querySelector("goa-popover");
+        expect(popover?.getAttribute("open")).toBe("true");
         expect(inputField?.getAttribute("aria-owns")).toBe("menu-favcolor");
+        clearIcon = result.queryByTestId("clear-icon");
+        expect(clearIcon).toBeFalsy();
+        dropdownIcon = result.container.querySelector("goa-icon");
+        expect(dropdownIcon).toBeTruthy();
         expect(dropdownIcon?.getAttribute("ariaexpanded")).toBe("true");
         expect(dropdownIcon?.getAttribute("type")).toBe("chevron-up");
       });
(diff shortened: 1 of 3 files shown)


libs/web-components/src/components/dropdown/Dropdown.svelte
```diff
@@ -526,6 +516,9 @@
 
     handleKeyDown(e: KeyboardEvent) {
       switch (e.key) {
+        case "Enter":
+          this.onEnter(e);
+          break;
```
chrisolsen, line 521: Since the `onEnter` function is already being called in the keyup event add it here will result in it being called twice.

libs/web-components/src/components/dropdown/Dropdown.svelte
```diff
@@ -364,26 +364,16 @@
...
-      setDisplayedValue();
-      setTimeout(() => {
-        hideMenu();
-      }, 10);
-    } else {
-      _selectedOption = undefined;
-      setDisplayedValue();
-      dispatchValue("");
-    }
+    setDisplayedValue();
```
chrisolsen, line 376: This is replacing a lot of logic (that should've had a comment), but from the commit it was there to prevent dropdown from firing _change event when the value is reselected.

Can we verify this [previous bug fix](https://github.com/govalta/ui-components/issues/1954)  isn't being undone

libs/web-components/src/components/dropdown/Dropdown.spec.ts
```diff
@@ -225,6 +225,181 @@ describe("GoADropdown", () => {
...
+      dropdown?.addEventListener("_change", (e: Event) => {
+        const ce = e as CustomEvent;
+        const d = ce.detail;
+        onChange(d.name, d.value);
+      });
+
+      const option = result.queryByTestId("dropdown-item-blue");
+      expect(option).toBeTruthy();
+
+      option?.addEventListener("click", (e: Event) => {
```
chrisolsen, line 263: Is it necessary to override the click event of the dropdown item, as there are other tests on this page that don't?

libs/web-components/src/components/dropdown/Dropdown.spec.ts
```diff
@@ -225,6 +225,181 @@ describe("GoADropdown", () => {
...
+      input.addEventListener("keydown", (e: KeyboardEvent) => {
+        if (e.key !== "Enter") {
+          return;
+        }
+        const ce = new CustomEvent("_change", {
+          detail: { name, value: "blue" },
+          bubbles: true,
+        });
+        input.dispatchEvent(ce);
+      });
```
chrisolsen, line 325: This is another place that were adding/overriding event handlers within the component. If without it we can't get the test to work because of a limitation of the testing-library then the test will be ok until we shift over to the headless tests, then this test needs to be changed.

This would be a good test to rewrite now in the headless testing branch.

chrisolsen:
Can you prefix your first two commits in this PR with `chore: `

libs/web-components/src/components/dropdown/Dropdown.svelte
BumbleB2na: Yes that previous fix is still good. Checked again, but also noticed that previous fix in git history and tested for it before. The `setDisplayedValue()` resets the input to the last selected value or empty string and that's what prevents re-selecting.

BumbleB2na: To clarify what was checked exactly: it does not reach this onChange handler when reselecting

libs/web-components/src/components/dropdown/Dropdown.spec.ts
BumbleB2na: No it wasn't necessary. Thanks and removed some `await tick()` from the other new tests.

libs/web-components/src/components/dropdown/Dropdown.svelte
BumbleB2na: Thanks was able to remove it from keyup, so now it's been "moved" from keyup to keydown.

libs/web-components/src/components/dropdown/Dropdown.spec.ts
BumbleB2na: Could not easily remove it and like the idea of tackling this in other headless branch that's in progress

libs/web-components/src/components/dropdown/Dropdown.svelte
```diff
@@ -526,6 +513,10 @@
 
     handleKeyDown(e: KeyboardEvent) {
       switch (e.key) {
+        case "Enter":
+        console.log('enter3')
```
chrisolsen, line 517: Just remove the console.log calls (three of them)

chrisolsen (changes requested):
The commits need to be squashed.

chrisolsen (dismissed):

libs/web-components/src/components/dropdown/Dropdown.svelte
```diff
@@ -673,7 +670,6 @@
           {name}
           on:keydown={onInputKeyDown}
           on:keyup={onInputKeyUp}
-          on:change={onChange}
```
chrisolsen, line 686: You should talk to Vanessa about this change as the `_onChange` event was previously added to properly handle autofilled data changes.

BumbleB2na: Please try again. It handles onChange event again to support browser autofill

BumbleB2na:
Demonstration of how #2304 was resolved:
<video src="https://github.com/user-attachments/assets/b889f136-f46e-4f62-b00a-ee2194e59833" width="200"></video>

ArakTaiRoth:
@BumbleB2na This was tied also to story #2304 . However, that issue isn't fixed for non-filterable Dropdowns.

libs/web-components/src/components/dropdown/Dropdown.svelte
```diff
@@ -268,7 +268,6 @@
   }
 
   function setHighlightedToSelected() {
-    if (!_filterable) return;
```
BumbleB2na: Resolves https://github.com/GovAlta/ui-components/pull/2303#issuecomment-2617184496 because, basic dropdown was working same as filterable dropdown re: scroll to selected option

BumbleB2na:
@vanessatran-ddi thank you it's now ready to be re-tested

BumbleB2na:
@vanessatran-ddi it's fixed now

BumbleB2na:
Thank you, @vanessatran-ddi will be adding test coverage for clearing selected after backspace completely. Then building on test coverage for autofill that was added earlier in this PR so that it will cover use case such as Country and State/Province dynamic selection. E.g. One dropdown dynamically populates another (or swaps dropdown with input field if no match on first dropdown) and autofill must continue to work on dynamic dropdown or input.

syedszeeshan (approved):

Issue #2200, opened 2024-10-16
Textarea: Setting width using a percentage makes the textarea artificially smaller


ArakTaiRoth:
# Info

If you set the width of a TextArea component as `100%`, there are no issues. If you use `ch` instead of `%`, there are also no issues, but then the TextArea component doesn't scale properly to window size.

If you set the width to anything less than 100 when using `%`, you get something like this:
![image](https://github.com/user-attachments/assets/0a9c98fb-18c1-4aa9-b492-b1cee4a4e7f4)
Where the width set on the TextArea also becomes the available width inside the TextArea (in this example, 90%). If you set no width on the TextArea, then you get the following:
![image](https://github.com/user-attachments/assets/5719942c-9dd5-4d9c-98f5-21309c803813)

## Acceptance Criteria:

1. Available width inside the TextArea should be the full TextArea

PR #2308, opened 2024-12-13
fix(#2200): textarea width issues

syedszeeshan:
# Before (the change)

# After (the change)

### DIFF
--- a/libs/web-components/src/components/text-area/TextArea.svelte
+++ b/libs/web-components/src/components/text-area/TextArea.svelte
@@ -9,8 +9,11 @@
     relay,
     toBoolean,
   } from "../../common/utils";
-  import type { Spacing } from "../../common/styling";
-  import { calculateMargin } from "../../common/styling";
+  import {
+    calculateMargin,
+    injectCss,
+    type Spacing,
+  } from "../../common/styling";
   import { onMount } from "svelte";
   import {
     FieldsetResetErrorsMsg,
@@ -74,10 +77,12 @@
   onMount(() => {
     addRelayListener();
     sendMountedMessage();
+    injectCss(_rootEl, ":host", {
+      width: width.includes("%") ? width : `min(${width}, 100%)`,
+    });
   });
 
   // functions
-
   function addRelayListener() {
     receive(_textareaEl, (action, data) => {
       switch (action) {
@@ -200,6 +205,7 @@
 
     box-sizing: border-box;
     font-family: var(--goa-font-family-sans);
+    display: inline-block;
   }
 
   #container {
@@ -208,10 +214,11 @@
 
   .root {
     position: relative;
-    width: var(--width, 100%);
+    width: 100%;
     padding-bottom: var(--char-count-padding);
     border: var(--goa-border-width-s) solid var(--goa-color-greyscale-700);
     border-radius: 3px;
+    transition: width 0.3s ease-in-out;
   }
 
   textarea {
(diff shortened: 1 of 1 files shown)


syedszeeshan:
image

libs/web-components/src/components/text-area/TextArea.svelte
```diff
@@ -200,10 +200,13 @@
 
     box-sizing: border-box;
     font-family: var(--goa-font-family-sans);
+    display: block;
```
chrisolsen, line 203: `inline-block` might be a better choice as it will allow for the setting of the width, but doesn't break the textarea's default behaviour (native textarea is also inline-block)

libs/web-components/src/components/text-area/TextArea.svelte
```diff
@@ -200,10 +200,13 @@
...
+    width: 100%;
```
chrisolsen, line 204: Ideally, this width is the one that should be set, but since it is within the :host scope it can't be set via css vars. However, there is a, `injectCss` helper method that will create :host block with the desired value.

```
    injectCss(rootEl, ":host", { width });
```

libs/web-components/src/components/text-area/TextArea.svelte
```diff
@@ -229,19 +236,25 @@
...
-      min-width: 0;
-      width: var(--width, 100%);
+
+    .root[style*="%"] {
+      width: var(--width);
+    }
+
+    .root:not([style*="%"]) {
+      width: min(var(--width), 100%);
     }
```
BumbleB2na, line 258:
```suggestion
```
Based on some testing done those styles can be removed because the max-width style on Line 249 controls width.

Tested in react playground and in playground\web\src\pages\TextareaPage.svelte:
`<goa-textarea name="item" value="read this" width="75%"></goa-textarea>`

Try removing those styles and if it works for you, too then it looks like Lines 212-259 can be simplified even more:
```
  .root {
    position: relative;
    width: var(--width, 100%);
    padding-bottom: var(--char-count-padding);
    border: var(--goa-border-width-s) solid var(--goa-color-greyscale-700);
    border-radius: 3px;
    transition:
      width 0.3s ease-in-out,
      box-shadow 0.2s ease;
  }
  
  @container self (--mobile) {
    .root {
      width: 100%;
    }
  }

  textarea {
    display: block;
    box-sizing: border-box;
    outline: none;
    border: none;
    border-radius: 3px;
    color: var(--goa-color-greyscale-black, #ccc);
    padding: var(--textarea-padding-vertical) var(--textarea-padding-horizontal);
    font-family: var(--goa-font-family-sans);
    width: 100%;
    resize: vertical;
  }
```

BumbleB2na (changes requested):

libs/web-components/src/components/text-area/TextArea.svelte
syedszeeshan: PR updated

libs/web-components/src/components/text-area/TextArea.svelte
```diff
@@ -212,6 +215,10 @@
     padding-bottom: var(--char-count-padding);
     border: var(--goa-border-width-s) solid var(--goa-color-greyscale-700);
     border-radius: 3px;
+    box-sizing: border-box;
```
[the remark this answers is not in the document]
syedszeeshan: PR updated

libs/web-components/src/components/text-area/TextArea.svelte
syedszeeshan: PR updated

libs/web-components/src/components/text-area/TextArea.svelte
syedszeeshan: PR updated

BumbleB2na (approved):

PR #2353, opened 2025-01-15
Public form updates

chrisolsen:


libs/web-components/src/components/input/Input.svelte
```diff
@@ -144,39 +145,42 @@
...
     value = detail.value;
-    dispatch(
-      inputEl,
-      "_change",
-      { name, value: detail.value },
-      { bubbles: true },
-    );
+
+    // dispatch the change event to the form-item element to ensure the state is in sync
+    // dispatch(_inputEl, "_change", { name, value }, { bubbles: true });
```
chrisolsen, line 175: This might need to be uncommented

Issue #2003, opened 2024-07-16
Allow popover component to never need the `relative` prop


chrisolsen:

ArakTaiRoth:
@BumbleB2na Moving this to blocked until the #2344 is merged into alpha, and this associated PR can be rebased

BumbleB2na:
`relative` prop can be removed completely from Popover, but it needs to be deprecated in react and angular Dropdown and Date Picker

PR #2358, opened 2025-01-20
fix(#2003): popover remove relative prop

BumbleB2na:
Also see related ui-components-docs PR: https://github.com/GovAlta/ui-components-docs/pull/305

# Before (the change)

Popover had prop: "relative" that when enabled would add style, "position: relative;" to the popover container to help with positioning flow. Dropdown and DatePicker had "relative" prop that would be passed down to child popover component. Code was adjusted to always set style, "position: relative;" and to calculate the position of popover when displaying upwards.

# After (the change)

Popover always has "position: relative" (related to #2003):
<video src="https://github.com/user-attachments/assets/bdc9255a-c743-4e1f-8180-c2f78d77539e" width="200"></video>

Popover positioning is still okay when scrolling within a modal (related to #2466):
<video src="https://github.com/user-attachments/assets/cafca2ec-8c13-45e8-a652-52ba847d4249" width="200"></video>

Note: a popover cannot appear over top of modal because of z-index layer hierarchy
  - E.g. basic or filterable Dropdown (E.g. `<goa-dropdown filterable="true">`) cannot have its contents "drop down" outside of modal like a native Dropdown can, because native html select is not affected by z-index layering and will always show over top
  - this was already the behavior of popover when relative property was used before these changes (E.g. `<goa-dropdown relative="true">`)

## Steps needed to test

Try dropdown and datepicker, some with a wrapper that has "position: relative" and some inside a modal -- try resizing the viewport to see how it calculates space below to determine if it dropdown or datepicker should expand upwards -- and try this in the different playgrounds:
- [angular-modal.component.html.txt](https://github.com/user-attachments/files/18482932/angular-modal.component.html.txt)
- [react-modal.tsx.txt](https://github.com/user-attachments/files/18482933/react-modal.tsx.txt)
- [ModalPage.svelte.txt](https://github.com/user-attachments/files/18482934/ModalPage.svelte.txt)
  - Alternative [ModalPage.svelte.txt](https://github.com/user-attachments/files/19212549/2025-03-12-web-components-playground-ModalPage.svelte.txt) related to #2466

### DIFF
--- a/libs/angular-components/src/lib/components/date-picker/date-picker.ts
+++ b/libs/angular-components/src/lib/components/date-picker/date-picker.ts
@@ -46,6 +46,9 @@ export class GoabDatePicker implements ControlValueAccessor {
   @Input() max?: Date | string;
   @Input() error?: boolean;
   @Input() disabled?: boolean;
+  /***
+   * @deprecated This property has no effect and will be removed in a future version
+   */
   @Input() relative?: boolean;
   @Input() testId?: string;
   @Input() mt?: Spacing;
--- a/libs/angular-components/src/lib/components/dropdown/dropdown.ts
+++ b/libs/angular-components/src/lib/components/dropdown/dropdown.ts
@@ -59,6 +59,9 @@ export class GoabDropdown implements ControlValueAccessor{
   @Input() placeholder?: string;
   @Input() testId?: string;
   @Input() width?: string;
+  /***
+   * @deprecated This property has no effect and will be removed in a future version
+   */
   @Input() relative?: boolean;
   @Input() mt?: Spacing;
   @Input() mb?: Spacing;
--- a/libs/angular-components/src/lib/components/popover/popover.ts
+++ b/libs/angular-components/src/lib/components/popover/popover.ts
@@ -32,6 +32,9 @@ export class GoabPopover {
   @Input() minWidth?: string;
   @Input() padded = true;
   @Input() position?: GoabPopoverPosition;
+  /***
+   * @deprecated This property has no effect and will be removed in a future version
+   */
   @Input() relative?: boolean;
   @Input() testId?: string;
   @Input() mt?: Spacing;
--- a/libs/common/src/lib/common.ts
+++ b/libs/common/src/lib/common.ts
@@ -255,6 +255,9 @@ export interface GoabPopoverProps extends Margins {
   maxWidth?: string;
   padded?: boolean;
   position?: GoabPopoverPosition;
+  /***
+   * @deprecated This property has no effect and will be removed in a future version
+   */
   relative?: boolean;
 }
 
--- a/libs/react-components/src/lib/date-picker/date-picker.tsx
+++ b/libs/react-components/src/lib/date-picker/date-picker.tsx
@@ -30,6 +30,9 @@ export interface GoabDatePickerProps extends Margins {
   min?: Date;
   max?: Date;
   testId?: string;
+  /***
+   * @deprecated This property has no effect and will be removed in a future version
+   */
   relative?: boolean;
   disabled?: boolean;
   onChange: (detail: GoabDatePickerOnChangeDetail) => void;
--- a/libs/react-components/src/lib/dropdown/dropdown.tsx
+++ b/libs/react-components/src/lib/dropdown/dropdown.tsx
@@ -55,6 +55,9 @@ export interface GoabDropdownProps extends Margins {
   placeholder?: string;
   testId?: string;
   width?: string;
+  /***
+   * @deprecated This property has no effect and will be removed in a future version
+   */
   relative?: boolean;
 }
 
--- a/libs/react-components/src/lib/popover/popover.tsx
+++ b/libs/react-components/src/lib/popover/popover.tsx
@@ -27,6 +27,9 @@ export interface GoabPopoverProps extends Margins {
   padded?: boolean;
   position?: GoabPopoverPosition;
   children: ReactNode;
+  /***
+   * @deprecated This property has no effect and will be removed in a future version
+   */
   relative?: boolean;
 }
 
--- a/apps/angular-tests/src/app/dropdown/dropdown.component.html
+++ b/apps/angular-tests/src/app/dropdown/dropdown.component.html
@@ -32,8 +32,12 @@ value: {{reactiveFormCtrl.value | json}}
   </goab-form-item>
 
   <goab-form-item label="Country">
-    <goab-dropdown [filterable]="true" [relative]="true" name="country" width="30ch">
-      <goab-dropdown-item *ngFor="let opt of countries" [value]="opt.code" [label]="opt.name"></goab-dropdown-item>
+    <goab-dropdown [filterable]="true" name="country" width="30ch">
+      <goab-dropdown-item
+        *ngFor="let opt of countries"
+        [value]="opt.code"
+        [label]="opt.name"
+      ></goab-dropdown-item>
     </goab-dropdown>
   </goab-form-item>
 </form>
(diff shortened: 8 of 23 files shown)


ArakTaiRoth:
This PR needs to be rebased with alpha after #2344 is merged in, and then needs to be re-tested

BumbleB2na:
> This PR needs to be rebased with alpha after #2344 is merged in, and then needs to be re-tested

This is done. Then linked a second issue, #2466 and updated this PR description with a second video to demonstrate how that got resolved.

BumbleB2na:
Updated PR to deprecate `relative` property instead of removing it. However, the property no longer has any effect (every popover now has `postion: relative` by default) and developers will see a console warning if using either of the 3 affected components:

image

minhthytran:
Round 1 feedback: 
Issue #2466 is fixed on this PR. Has been tested with:

- [x] new angular
- [x] new react
- [x] firefox
- [x] chrome (mac) and Edge (windows)
- [x] safari

libs/angular-components/src/lib/components/date-picker/date-picker.ts
```diff
@@ -46,7 +46,7 @@ export class GoabDatePicker implements ControlValueAccessor {
   @Input() max?: Date | string;
   @Input() error?: boolean;
   @Input() disabled?: boolean;
-  @Input() relative?: boolean;
+  @Input() relative?: boolean; // @deprecated: property has no effect
```
dustin-nielsen-goa, line 49: Deprecate using Vanessa's suggestion above

libs/angular-components/src/lib/components/dropdown/dropdown.ts
```diff
@@ -59,7 +59,7 @@ export class GoabDropdown implements ControlValueAccessor{
   @Input() placeholder?: string;
   @Input() testId?: string;
   @Input() width?: string;
-  @Input() relative?: boolean;
+  @Input() relative?: boolean; // @deprecated: property has no effect
```
dustin-nielsen-goa, line 62: Deprecate using Vanessa's suggestion above

libs/angular-components/src/lib/components/popover/popover.ts
```diff
@@ -32,7 +32,7 @@ export class GoabPopover {
   @Input() minWidth?: string;
   @Input() padded = true;
   @Input() position?: GoabPopoverPosition;
-  @Input() relative?: boolean;
+  @Input() relative?: boolean; // @deprecated: property has no effect
```
dustin-nielsen-goa, line 35: Deprecate using Vanessa's suggestion above.

libs/common/src/lib/common.ts
```diff
@@ -255,7 +255,7 @@ export interface GoabPopoverProps extends Margins {
   maxWidth?: string;
   padded?: boolean;
   position?: GoabPopoverPosition;
-  relative?: boolean;
+  relative?: boolean; // @deprecated: property has no effect
```
dustin-nielsen-goa, line 258: I don't think we need to list as deprecated here, because it's being done in both the Angular components and the React components

libs/react-components/src/lib/date-picker/date-picker.tsx
```diff
@@ -8,7 +8,7 @@ interface WCProps extends Margins {
   error?: boolean;
   min?: string;
   max?: string;
-  relative?: boolean;
+  relative?: boolean; // @deprecated
```
dustin-nielsen-goa, line 11: I don't think we need to list it as deprecated here, because these aren't the props implemented by the wrapper

libs/react-components/src/lib/date-picker/date-picker.tsx
```diff
@@ -30,7 +30,7 @@ export interface GoabDatePickerProps extends Margins {
   min?: Date;
   max?: Date;
   testId?: string;
-  relative?: boolean;
+  relative?: boolean; // @deprecated: property has no effect
```
dustin-nielsen-goa, line 33: Deprecate as per Vanessa's suggestion above

libs/react-components/src/lib/date-picker/date-picker.tsx
```diff
@@ -47,7 +47,7 @@ export function GoabDatePicker({
   mr,
   mb,
   ml,
-  relative,
+  relative, // @deprecated
```
dustin-nielsen-goa, line 50: Not necessary here

libs/react-components/src/lib/dropdown/dropdown.tsx
```diff
@@ -20,7 +20,7 @@ interface WCProps extends Margins {
   placeholder?: string;
   value?: string;
   width?: string;
-  relative?: boolean;
+  relative?: boolean; // @deprecated
```
dustin-nielsen-goa, line 23: Not necessary here

libs/react-components/src/lib/dropdown/dropdown.tsx
```diff
@@ -55,7 +55,7 @@ export interface GoabDropdownProps extends Margins {
   placeholder?: string;
   testId?: string;
   width?: string;
-  relative?: boolean;
+  relative?: boolean; // @deprecated: property has no effect
```
dustin-nielsen-goa, line 58: Deprecate as per Vanessa's suggestion above

libs/react-components/src/lib/popover/popover.tsx
```diff
@@ -6,7 +6,7 @@ interface WCProps extends Margins {
   minwidth?: string;
   padded?: boolean;
   position?: GoabPopoverPosition;
-  relative?: boolean;
+  relative?: boolean; // @deprecated
```
dustin-nielsen-goa, line 9: Not necessary here

libs/react-components/src/lib/popover/popover.tsx
```diff
@@ -27,7 +27,7 @@ export interface GoabPopoverProps extends Margins {
   padded?: boolean;
   position?: GoabPopoverPosition;
   children: ReactNode;
-  relative?: boolean;
+  relative?: boolean; // @deprecated: property has no effect
```
dustin-nielsen-goa, line 30: Deprecate as per Vanessa's suggestion above

libs/react-components/src/lib/popover/popover.tsx
```diff
@@ -37,7 +37,7 @@ export function GoabPopover({
   minWidth,
   padded,
   position,
-  relative,
+  relative, // @deprecated
```
dustin-nielsen-goa, line 40: Not necessary here

dustin-nielsen-goa:
There's a problem using the Dropdown in the AppHeaderMenu component, it appears that the width isn't being calculated properly
Before:
before

After:
after

libs/common/src/lib/common.ts
BumbleB2na: Agreed but, decided to leave this one in for future cleanup

BumbleB2na:
> @BumbleB2na There's a problem using the Dropdown in the AppHeaderMenu component, it appears that the width isn't being calculated properly

Was not able to reproduce width problem but, this aided in the discovery that AppHeader and AppHeaderMenu depended on popover not having `position: relative` so, these two components override that style to turn it off, allowing them to display the same as before.

ArakTaiRoth (approved):

chrisolsen:
I am seeing the following

1. When a non-native dropdown is at the bottom of a modal (this was the reason for previously having the `relative` prop)
image

2. When wrapping the dropdown within an absolutely positioned parent element (the calendar belongs to the left most element)
image

chrisolsen (changes requested):
See my most recent comment

chrisolsen (approved):

Issue #2345, opened 2025-01-09
Reported Bug


ArakTaiRoth:
Issue reported by: Jim.muirhead@gov.ab.ca
Web component version: 1.27.0
Angular/React version: 3.0.2

## Expected Behaviour

When I navigate to a page from one of the navigation dropdowns in the app header, I expect that navigation option/dropdown to be highlighted with the blue top border while I'm on one of the pages within that section.

## Actual Behaviour

On a page where the URL also includes a query string, the corresponding navigation option/dropdown in the app header is NOT highlighted, nor is the page highlighted in the navigation dropdown menu.

Example Page/URL: https://purchasing.alberta.ca/search?npo=1

## Replication Steps

1. Navigate to https://purchasing.alberta.ca/
2. Select the "Browse" navigation option in the app header.
Note that the "Browse" navigation option is NOT highlighted in the app header.
The actual URL of the page is https://purchasing.alberta.ca/search?npo=1

A similar case also happens for a signed-in Supplier user when navigating to the Find Opportunities & Notices page.

PR #2382, opened 2025-01-29
fix(#2345): appheader links with query strings not showing as active

syedszeeshan:
# Before (the change)

# After (the change)

https://github.com/user-attachments/assets/1b4fd831-31c4-4018-8f2a-c9cc11a00f08

### DIFF
--- a/libs/web-components/src/common/urls.ts
+++ b/libs/web-components/src/common/urls.ts
@@ -2,8 +2,16 @@ export function isUrlMatch(windowUrl: URL | Location, testUrl: string): number {
   // path match
   if (testUrl === undefined) return -1;
 
+  let testPathname;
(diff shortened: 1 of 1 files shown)


ArakTaiRoth:
@syedszeeshan There's a problem with this, evidenced with this [jam](https://jam.dev/c/d15777fe-2bdc-4085-85e8-47af8dcf001b).

Basically, if you go to the link with the query string, so long as the query string is there it matches. But if you go to the page without the query string, the same link doesn't get highlighted, and it should because it's still taking the person to the same page. Take a look at how the SideMenu works, the link gets activated if it has a query string and if it doesn't, it should work the same way.

BumbleB2na (approved):

PR #2405, opened 2025-02-07
Public form fixes

chrisolsen:
The last commit includes a fix to the datepicker, in it not changing when the value it is bound to changes. This issue can be tested with the following 

Angular only
```ts
  someDate = "2024-06-12";

  resetDate() {
    this.someDate = "2026-02-13";
  }

  onDateChange(e: Event) {
    this.someDate = (e as CustomEvent).detail.value;
  }


```

```
    <goab-block direction="row" alignment="center">
      <goa-date-picker type="input" [value]="someDate" (_change)="onDateChange($event)"></goa-date-picker>
      <goa-link-button (_click)="resetDate()">Reset date</goa-link-button>
    </goab-block>
```

Before
[Screencast from 2025-02-12 14-28-35.webm](https://github.com/user-attachments/assets/82e925b1-57e3-436e-b5f6-7fbc10202776)

After
[Screencast from 2025-02-12 14-27-25.webm](https://github.com/user-attachments/assets/9382f82c-5021-44e5-8465-3cdab8c55255)

### DIFF
--- a/libs/common/src/lib/public-form-controller.spec.ts
+++ b/libs/common/src/lib/public-form-controller.spec.ts
@@ -1,0 +1,18 @@
+import { describe, it, expect } from "vitest";
+import { PublicFormController } from "./public-form-controller";
+
+describe("PublicFormController", () => {
+  const pfc = new PublicFormController("details");
+
+  describe("clean", () => {
+    const data = JSON.parse(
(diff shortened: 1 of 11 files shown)


syedszeeshan:
Shared my findings with Chris in a screenshare....
-The date gets set to 'Feb 1st, 2026' the first time the button is clicked, but not after. 
-Also, the type="input" was giving an error about [value] assignment "Can't bind to 'value' since it isn't a known property of 'goa-date-picker'".

syedszeeshan (dismissed):

ArakTaiRoth (approved):

Issue #2413, opened 2025-02-12
React 19 not supported by react-components


vanessatran-ddi:
### Describe the bug

npm complains about the peer dependency
```
 "devDependencies": {
    "@eslint/js": "^9.19.0",
    "@types/react": "^19.0.8",
    "@types/react-dom": "^19.0.3",
    "@vitejs/plugin-react": "^4.3.4",
    "eslint": "^9.19.0",
    "eslint-plugin-react-hooks": "^5.0.0",
    "eslint-plugin-react-refresh": "^0.4.18",
    "globals": "^15.14.0",
    "typescript": "~5.7.2",
    "typescript-eslint": "^8.22.0"
  }
```

If we install `npm install @abgov/react-components`
and same with `npm install @abgov/ui-components-common`

Error is as below:
```
npm install @abgov/react-components@latest
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR! 
npm ERR! While resolving: vite-react-typescript-starter@0.0.0
npm ERR! Found: @types/react@19.0.8
npm ERR! node_modules/@types/react
npm ERR!   dev @types/react@"^19.0.8" from the root project
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer @types/react@"^17.0.0 || ^18.0.0" from @abgov/react-components@6.1.0
npm ERR! node_modules/@abgov/react-components
npm ERR!   @abgov/react-components@"6.1.0" from the root project
npm ERR! 
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
npm ERR! 
npm ERR! 
npm ERR! For a full report see:
npm ERR! /home/.npm/_logs/2025-02-12T15_14_50_239Z-eresolve-report.txt

npm ERR! A complete log of this run can be found in: /home/.npm/_logs/2025-02-12T15_14_50_239Z-debug-0.log
```

```
 npm install @abgov/ui-components-common
npm ERR! code ERESOLVE
npm ERR! ERESOLVE could not resolve
npm ERR! 
npm ERR! While resolving: @abgov/react-components@6.1.0
npm ERR! Found: @types/react@19.0.8
npm ERR! node_modules/@types/react
npm ERR!   peer @types/react@"^19.0.0" from @types/react-dom@19.0.3
[...]

Issue #1780, opened 2024-04-15
Breaking Changes


ArakTaiRoth:
# Info

This story is intended to gather all breaking changes we potentially want to make to our components. This is to be used when we have to make a breaking change to one of our libraries, we can include more to make it more "worthwhile".

When adding a breaking change under one of our libraries, please include as much context as possible, as it could be many months before we ever even look at these, and memories suck.

## Web Components

1. Make the `header` property required for Modal component
2. Make the `arialabel` property required for Badge component if `content` property isn't provided
    * This is also included under React components
3. Update default value for `step` property for Input component to -1. I'm not sure why this one is being asked for anymore, if anyone is able to provided context for this, that would be great. 

## React Components

1. Add support for event parameters for React. I'm not sure what this one is getting at either, if anyone can provide context for this, that would be great.
2. Remove the inclusion of `web-components` from the React library. This means teams would have to manually include the web-components library in their project.

twjeffery:
Update to typography has been waiting on a breaking change. 
https://github.com/GovAlta/ui-components/issues/1410

This is an update to: 
- default margin top and bottom for each type
- max length
- mobile type scale
- default spacing for specific text pairings

Design issue here:
https://goa-dio.atlassian.net/browse/DDIDS-1399?atlOrigin=eyJpIjoiNDU4MDlhNTQwN2Y0NGMzZWIyMmNhYjc2MmQ3ODM3MTAiLCJwIjoiaiJ9

Design file here: 
https://www.figma.com/file/7xYbs6rAM4cwBdalPNE8lH/Text-combinations---spacing?type=design&node-id=0-1&mode=design

BumbleB2na:
If it's not too soon to add this under, _Web Components:_
- Remove deprecated `relative` property from Dropdown, Date Picker and Popover components
  - This also needs to be included under React and Angular components (see #2003)

ArakTaiRoth:
Closing as we have new ways of tracking stories requiring Breaking Changes

PR #2460, opened 2025-03-06
fix(#2413): react 19 upgrade

BumbleB2na:
Run `npm ci` or `npm install`

# Before (the change)

Could not use react 19 in a project when importing @abgov/react-components without a peer dependency conflict.

# After (the change)

Should no longer encounter a peer dependency conflict when using react 19 in a project that imports @abgov/react-components. An upgrade to React 19 was required.

## Upgrade to React 19:
1. followed react 19 upgrade guide: https://react.dev/blog/2024/04/25/react-19-upgrade-guide
1. upgraded 4 packages to react version 19.0.0:
    ```bash
    npm install --save-exact react@^19.0.0 react-dom@^19.0.0 @types/react@^19.0.0 @types/react-dom@^19.0.0
    ```
1. ran codemon for react
    ```bash
    npx codemod@latest react/19/migration-recipe
    ```
    - no changes
1. ran codemon for react typescript:
    ```bash
    npx types-react-codemod@latest preset-19 ./_templates/react
    ```
    - it changed "ref" objects in .tsx files (in React 19 "useRef" works more like "useContext"). Some manual changes were needed, like in form.tsx for example:
        - `ref?: React.MutableRefObject<HTMLElement | undefined>;` --> `ref?: React.RefObject<HTMLElement | null>;` because MutableRefObject is deprecated.
        - `const el = useRef<HTMLElement>();` --> `const el = useRef<HTMLElement>(null);` because initializing RefObject is required and `undefined` is disallowed.
    - See React 19 Upgrade guide: https://react.dev/blog/2024/04/25/react-19-upgrade-guide#useref-requires-argument
1. attempted build, then had to manually changed each react component that defines a custom element to be used in its return statement. Take form.tsx for example:
    - before the change it would throw an error during `nx run react-components:build:production': "error TS2339: Property 'goa-simple-form' does not exist on type 'JSX.IntrinsicElements'."
    - The error and solution are somewhat similar to what is described here: https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/71395#discussioncomment-11525908
    - See React 19 Upgrade guide: https://react.dev/blog/2024/04/25/react-19-upgrade-guide#the-jsx-namespace-in-typescript
1. needed to change all boolean values in react components to convert to string when passed to related web component.
    - Changed boolean in a consistent way so that react components do not set a default, allowing props for setting boolean values to optional.
    - Updated all related tests to
      - render a component with only required value and check boolean flags to make sure they're optional
      - render a component with non-required properties and set any boolean flags to true to make sure they can be set
1. built without errors, served react playground
1. upgraded libs/react-components/package.json: allow peer dependencies for react ^19.0.0
    - Should no longer encounter a peer dependency conflict when using react 19 in a project that imports @abgov/react-components.

## Steps to test
- run `npm ci` to upgrade locally and confirm there are no peer dependency conflicts
- optional: could try each component in react playground and toggle settings related to changed boolean flags (although there's test coverage for all of the changed boolean flags)


libs/react-components/src/lib/input/input.spec.tsx
```diff
@@ -34,8 +46,10 @@ describe("Input", () => {
       disabled: true,
       readonly: true,
       focused: true,
+      error: true,
       placeholder: "placeholder",
-      prefix: "foo",
+      // TODO: remove deprecated property or fix spec failure
+      // prefix: "foo",
```
BumbleB2na, line 52: Confirmed that `prefix` still works in react playground. However, this was causing test to fail after upgrade. It is acting like `prefix` is a reserved word. Confirmed that changing it to `prefix2` works.

libs/react-components/src/utils.ts
```diff
@@ -0,0 +1,12 @@
...
+  settings: { omitIfFalse?: boolean } = {},
+): string | undefined {
+  return typeof value === "undefined"
+    ? undefined
+    : value
+      ? "true"
+      : settings.omitIfFalse
+        ? undefined
+        : "false";
+}
```
chrisolsen, line 12: Is this needed? Syed had a similar issue just the other week and the resolution was just to set the property with 
```
<SomeComponent
   someBoolProp={prop || undefined}
/>
```
Does this not work with React 19?

BumbleB2na: That would work with React 19, but don't want to omit if it's false. The exception is `disabled` property.

BumbleB2na: Rebased to resolve conflict in [button.tsx](https://github.com/GovAlta/ui-components/pull/2460/files#diff-1da93302ffe441c36c58a6674a82416a229ef4ac6a65c77fa7795f9ae4eb625f), see Line 87 in that file for example of `disabled` property

chrisolsen: Omitting if it's false is ok. So let's remove that helper func.

BumbleB2na: The helper function is needed to return bool value as string because React 19 upgraded typescript

libs/react-components/src/lib/input/input.spec.tsx
BumbleB2na: Create issue #2539 that will resolve this up by removing deprecated property

BumbleB2na:
@chrisolsen formatting changes were moved to a separate commit

libs/react-components/src/lib/input/input.spec.tsx
BumbleB2na: Update: will remove property in #1780

chrisolsen:
@BumbleB2na Can you rebase this PR?

chrisolsen:
@BumbleB2na Thanks for rebasing, unfortunately after some other merges it is out of sync again :)

libs/react-components/src/utils.ts
BumbleB2na: Thought about this some more then removed the helper function. It would be worth mentioning that we did not always omit if false and that is what it is doing now but, I have not rebased the last commit in case we want to undo that. There was only one exception to omitting if false and that was Popover.svelte where `export let padded = "true"` is the default.  Then drawer.tsx had to be updated because Drawer.svelte uses boolean for `open`.

libs/react-components/src/lib/table/table.tsx
```diff
@@ -23,7 +26,7 @@ declare global {
 export interface GoabTableProps extends Margins {
   width?: string;
   onSort?: (detail: GoabTableOnSortDetail) => void;
-  // stickyHeader?: boolean; TODO: enable this later
+  stickyHeader?: boolean;
```
chrisolsen, line 29: Uncommenting this line will then make it visible to other devs, so leaving it as commented out, or remove it all together would probably be best.

chrisolsen (approved):

libs/react-components/src/lib/app-header/app-header.tsx
```diff
@@ -1,16 +1,16 @@
...
+import { useEffect, useRef, type JSX } from "react";
 
 interface WCProps {
   heading?: string;
   url?: string;
   maxcontentwidth?: string;
   fullmenubreakpoint?: number;
-  hasmenuclickhandler?: boolean;
-  ref: React.RefObject<HTMLElement>;
+  hasmenuclickhandler?: string;
```
chrisolsen, line 8: Why are some of the previous boolean values getting changed to strings?

BumbleB2na: This property is a string in AppHeader.svelte and it needed to be changed here to match, because of typescript upgrade in react 19

chrisolsen: Will this result in a breaking change for existing React 18 apps?

BumbleB2na: The `hasmenuclickhandler` is not exposed as an attribute of this react component so, it will not

ArakTaiRoth (approved):

Issue #2384, opened 2025-01-30
Button rendered with disabled=false (React only)


ArakTaiRoth:
Issue reported by: ben.lam@gov.ab.ca
Web component version: 1.31.0
Angular/React version: 5.4.1

This was also reported in the UIAM accessibility report

# Info

If I use a GoAButton in React, it is rendered with `disabled="false"`. This causes issues with screen readers reading the button out as disabled, when it isn't.

## Code for Replication

```javascript
<GoabButton>Not set</GoabButton>
```
OR
```javascript
<GoAButton>Not set</GoAButton>
```

## Acceptance Criteria

1. Using the Button component in React should not render `disabled="false"`, any rendering of disabled causes screen readers to believe the button is disabled.

ArakTaiRoth:
Moving back to In Progress until a test is written to test that this indeed works, in both PR's (both branches, alpha and LTS)

PR #2467, opened 2025-03-12
fix(#2384): (LTS) button rendered with disabled=false react 

syedszeeshan:
# Before (the change)

# After (the change)

### DIFF
--- a/libs/react-components/src/lib/button/button.tsx
+++ b/libs/react-components/src/lib/button/button.tsx
@@ -2,12 +2,7 @@ import { ReactNode, useEffect, useRef } from "react";
 import { Margins } from "../../common/styling";
 import { GoAIconType } from "../icon/icon";
 
-export type GoAButtonType =
-  | "primary"
-  | "submit"
-  | "secondary"
-  | "tertiary"
-  | "start";
+export type GoAButtonType = "primary" | "submit" | "secondary" | "tertiary" | "start";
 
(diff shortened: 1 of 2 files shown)


libs/react-components/src/lib/button/button.tsx
```diff
@@ -93,7 +88,7 @@ export function GoAButton({
       type={type}
       size={size}
       variant={variant}
-      disabled={disabled}
+      {...(disabled ? { disabled: true } : {})}
```
BumbleB2na, line 91:
```suggestion
      disabled={typeof value === 'undefined' ? undefined : value ? "true" : "false"}
```
Suggesting this change to allow for a) true, b) false, or c) undefined. if prop not passed in it will not include it in the DOM

BumbleB2na (approved):
Have updated #2460 so that `disabled={false}` will continue to work as it does here

libs/react-components/src/lib/button/button.tsx
chrisolsen, line 91: Can this can be simplified to the following?
```
disabled={disabled || undefined}
```

syedszeeshan: Doing this will still render it....Acceptance criteria is to not render `disabled `attribute at all (when its false), to avoid Screen Reader issues.

chrisolsen (approved):

Issue #2482, opened 2025-03-19
Re-arrange common.ts and React specific exports


syedszeeshan:
## Acceptance Criteria:
-Move `Margins `out of `common.ts` into a new `types.ts`
-Any `Props` (e.g. `GoabTabsProps`, `GoabTableProps`, etc), move them out of common.ts into their corresponding tsx components if not already there
-replace existing Margins import statements to import from "`common/types.ts`";
-update `index.ts` to export common/types.ts

When creating a new React app,
2. We want to be able to build without @abgov/ui-components-common in package.json
3. Import only from the @abgov/react-components

**Desired Result**
After the above changes, it eliminates the dependency on "`@abgov/ui-components-common`"
and all you need is import "`@abgov/react-components`"

syedszeeshan:
@Spark450 @ArakTaiRoth 
I mentioned this in our daily standup, but I think I should have mentioned it somewhere in this new user story also, that it is a continuation/improvement of #2389 

Vanessa, Chris and I discussed and we decided to do the changes that you see in the new PR #2484 ..... but since US 2389 was already merged and closed, I created this new US for myself, just for tracking purpose. I could have made these changes without creating a new US also.

I have updated the Acceptance criteria.

Spark450:
Okay, thanks @syedszeeshan !

BumbleB2na:
QA Testing

In ui-components/playground/react there is intellisense to import types:
![Image](https://github.com/user-attachments/assets/45989dae-4517-4c41-a7e9-50ceebda25db)

In a separate react project there is no intellisense to import types (in my opinion there should be intellisense):
![Image](https://github.com/user-attachments/assets/55b30530-367d-4674-890f-7531d2bf69c3)

However, in that separate project if we install "@abgov/ui-components-common" then intellisense comes along with that:
![Image](https://github.com/user-attachments/assets/60aefdbd-160a-4e1c-90ae-00f3eaf4f275)

A workaround without needing intellisense in VSCode is to use Quick fix the import the type. But, in my opinion this shouldn't be needed. There should be intellisense.

syedszeeshan:
As discussed in the Dev meeting:

I think the PR does the job as outlined in the acceptance criteria. I will let the team decide on whether this VSCode specific issue is a blocker for this PR. 

Below are the details for future reference

**WebStorm**
1-manually importing the type works
2-Hover mouse on the missing reference, and **click "Insert import" works**
[...]

Issue #2389, opened 2025-02-03
Angular: Importing common components imports from angular-components instead of ui-components-common


ArakTaiRoth:
# Info

If you import any common components, such as `GoabBadgeType` while using Angular, it defaults to importing from `@abgov/angular-components`, instead of `@abgov/ui-components-common`. This is the default behaviour in React.

## Acceptance Criteria:

1. Angular should default to import common components from `@abgov/ui-components-common`

syedszeeshan:
Tested using below locally and on stackblitz
Vanessa's stackblits and demo react app links below for reference #2444 
-https://stackblitz.com/~/github.com/vanessatran-ddi/react-demo-app
-https://github.com/vanessatran-ddi/react-demo-app.git

```
    "@abgov/react-components": "^6.2.1-alpha.5",
    "@abgov/ui-components-common": "^1.2.1-alpha.1",
    "@abgov/web-components": "^1.32.1-alpha.6",
```

PR #2484, opened 2025-03-19
fix(#2482): re-arrange common.ts and react specific exports

syedszeeshan:
# Before (the change)

# After (the change)

## Steps needed to test
In your react app's package.json, this `"@abgov/ui-components-common":` should not be required anymore.

```
"dependencies": {
    "@abgov/react-components": ".....",
    "@abgov/web-components": "....",
```

Your React app should not have any imports referencing `@abgov/ui-components-common`, rather it should be importing from `@abgov/react-components` only


syedszeeshan:
#2389

BumbleB2na (approved):

BumbleB2na (changes requested):
In a separate react project there is no intellisense to import types:
image

Here's a way to make intellisense work:
image

libs/react-components/src/lib/table/table.tsx
```diff
@@ -29,6 +33,14 @@ export interface GoabTableProps extends Margins {
...
 }
 
+// Alerady exists
+// export interface GoabTableProps extends Margins {
+//   width?: string;
+//   onSort?: (sortBy: string, sortDir: number) => void;
+//   variant?: GoabTableVariant;
+//   testId?: string;
+// }
+
```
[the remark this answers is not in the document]
syedszeeshan: Thanks, removed.

syedszeeshan:
@BumbleB2na 
I made the exact changes as line 2 and line 12 in your screenshot, and then tried the ctrl+space shortcut to import... in an external React app, but no luck with the intellisense as you mentioned in your screenshot. 

Other ways of import continue to work as before in VSCode. And the Webstorm was always working.

chrisolsen:
@syedszeeshan Once this conflict is resolved it can be merged.

BumbleB2na:
> @syedszeeshan Once this conflict is resolved it can be merged.

Please wait because, I've requested a change where @syedszeeshan and I have discussed offline

libs/react-components/src/lib/input/input.tsx
```diff
@@ -8,9 +8,9 @@ import {
   GoabInputOnChangeDetail,
   GoabInputOnFocusDetail,
   GoabInputOnKeyPressDetail,
```
BumbleB2na, line 10: Since types have been moved wouldn't it make sense to import from `"../../common/types"` just like `Margin`?

BumbleB2na (changes requested):

libs/react-components/src/lib/side-menu-group/side-menu-group.tsx
```diff
@@ -1,6 +1,6 @@
 import { ReactNode } from "react";
-import { GoabIconType, Margins } from "@abgov/ui-components-common";
-
+import { GoabIconType,  } from "@abgov/ui-components-common";
+import { Margins } from "../../common/types";
```
BumbleB2na, line 3: Leftover comma
```suggestion
import { GoabIconType } from "@abgov/ui-components-common";
import { Margins } from "../../common/types";
```
Furthermore it may make sense to import this way and other .tsx files, too:
```suggestion
import type { GoabIconType } from "../../common/types";
import { Margins } from "../../common/types";
```

syedszeeshan: The lonely comma has been removed.

BumbleB2na: The reason for the second suggestion to import types from "../../common/types" in all .tsx files is that over time while maintaining react-components it helps to decrease the risk of forgetting to add any new types to /react-components/src/common/types.ts and the consequence of forgetting is that the type won't be available in other projects, so I'd recommend it.

BumbleB2na (commented):
@syedszeeshan Before noticing this PR was moved to draft mode I found a better way to test and found a problem. Those types work as expected in a separate project when importing 'ui-components-common':
image
But, as you can see they don't work as expected when using 'react-components':
image

chrisolsen:
@BumbleB2na I get an error within the playground. Is this only seen within an external project?
image

syedszeeshan:
https://github.com/user-attachments/assets/dd33d24b-a03b-413d-9a53-30deef3efd0f

@chrisolsen @BumbleB2na 
Here is the comparison b/w playground (left) and external React app (right)

The import works but it does not give errors on the allowed values in the external app scenario

chrisolsen:
@BumbleB2na Does it show errors on build?

BumbleB2na:
> @BumbleB2na I get an error within the playground. Is this only seen within an external project? 

Same here. Playground gives errors, but not in external react vite project

syedszeeshan:
> @BumbleB2na Does it show errors on build?

The extenal react app BUILDS fine, as you can see in the updated video.

chrisolsen:
Including the types within React is not working as expected, so this PR is no longer needed and any previous type imports will be removed. For now, React projects will need to include the ui-components-common lib in the package.json file. 

@syedszeeshan will create a new PR that contains the reverts to the already existing commont types in React

PR #2506, opened 2025-03-25
2410 modal accessibility

BumbleB2na:
# Before (the change)

When modal opened it would set initial focus to the first focusable element inside the dialog:
<video src="https://github.com/user-attachments/assets/809bbeea-caa6-42a2-af3f-791f4ff99e2b" width="200"></video>

# After (the change)

When modal opens it sets initial focus to the dialog itself:
<video src="https://github.com/user-attachments/assets/826911b1-762f-4d01-b56a-5d403059c041" width="200"></video>

Tested to ensure that still works if custom heading is used instead:
<video src="https://github.com/user-attachments/assets/c904312b-501d-4244-b1c8-d549b5bcfe9a" width="200"></video>

## Other info

- When modal is open Tab forwards or backwards remains within focus trap
- Recommendation from W3C for accessibility when dialog opens is to either a) set focus to first focusable element inside the dialog, or b) set focus to the dialog container. We had previously went with the former recommendation but, this was causing a problem in a Modal with lots of content where vertical scrollbar needed to be moved to set focus to first focusable element. The change here goes with the latter recommendation.
- Previously there was logic to allow skipping elements after the first focus, such as on a close Modal button. These were misleadingly named "ignore-focus" and now "first-focus" is used. There were changes to focus trap, currently used by Modal and Popover components.

## Steps needed to test

Code snippet provided in original issue can be used for testing


BumbleB2na:
@twjeffery suggestion for a follow up issue that would introduce a breaking change:
- if role is "alertdialog" then prevent allowing it to be closeable so that screenreader will continue read until first focusable element. Ideally that first focusable element is a button in the actions slot where user can confirm they understand or is given a choice, but we give developers control over what goes in there. It's a breaking change because developers may have been relying on their "alertdialog" to be closable with the "x" icon button.

chrisolsen:
@BumbleB2na  Remember, if a PR is blocked to move it back to draft mode

Issue #1219, opened 2023-06-14
Input: date type not rendering correctly on iOS


ArakTaiRoth:
## Summary of issue

- The Input type date is not working as intended on safari (iOS). 
- The height of the input field is incorrect when input is empty 

## More Info

- Height issue is likely related to OS, as it exists in both Chrome and Safari on MacOS and iOS

![image (28)](https://github.com/user-attachments/assets/6b322794-887a-4466-9cba-b0442ab9f230)

## Acceptance Criteria

- [ ] Height of the input field is correct when input is empty

Spark450:
Spoke with Andrew Mitchel (issue reporter Sept 2024). 

The team will try to use the date picker component instead and report back.

Most users of their product will be using IPhones or IPads.

Spark450:
Followup with Andrew Mitchell and Jennifer Chan, they communicated that the date format of YYYY MM DD was a hard requirement for their stakeholders and users. As a result the custom date picker control would not be suitable and they would need this fix to be applied in order for them to use our GOA input component.

ArakTaiRoth:
@syedszeeshan This will need to be rebased after PR #2296 is merged into Alpha, then re-tested and potentially worked on again

PR #2669, opened 2025-04-25
fix(1219): fixed height of input type date

twjeffery:
This is a fix for issue https://github.com/GovAlta/ui-components/issues/1219

This also fixes an additional bug where input type=date, datetime-local, and time, is 44px tall (should be 42px)

# Before (the change)
image

# After (the change)
image

### DIFF
--- a/libs/web-components/src/components/input/Input.svelte
+++ b/libs/web-components/src/components/input/Input.svelte
@@ -410,6 +410,7 @@
     display: inline-flex;
     align-items: stretch;
     width: 100%;
+    height: 42px;
     z-index: 1;
     background-color: var(--goa-text-input-color-bg);
     /* default border */
@@ -465,7 +466,6 @@
   }
 
   input {
-    display: inline-block;
     color: var(--goa-text-input-color-text);
     font: var(--goa-text-input-typography);
     padding: var(--goa-text-input-padding);
@@ -527,6 +527,7 @@
     display: flex;
     align-items: center;
     white-space: normal;
+    height: 42px;
   }
 
   .leading-content-slot :global(::slotted(div)),


ArakTaiRoth (approved):

libs/web-components/src/components/input/Input.svelte
```diff
@@ -466,6 +467,7 @@
 
   input {
     display: inline-block;
+    align-items: center;
```
chrisolsen, line 470: This line won't have an effect since the display is being overridden with a  `block` value.
image

libs/web-components/src/components/input/Input.svelte
```diff
@@ -466,6 +467,7 @@
 
   input {
     display: inline-block;
```
chrisolsen, line 469: According to the browser inspector, this `display` setting is being overwritten by the `input.input--goa` selector.
image

twjeffery: should I remove it?

libs/web-components/src/components/input/Input.svelte
twjeffery: should we remove it?

libs/web-components/src/components/input/Input.svelte
chrisolsen: I think so :+1:

libs/web-components/src/components/input/Input.svelte
chrisolsen: yes

ArakTaiRoth (approved):

chrisolsen (approved):

PR #2708, opened 2025-05-08
fix: input width missing 1ch, date picker too small

twjeffery:
# Before (the change)
image

image

# After (the change)
image
image


chrisolsen:
These changes add to #2706

Issue #2722, opened 2025-05-13
Add text-align option to goab-input component


chrisolsen:
## Description
The `goab-input` component currently always left-aligns its content. For numeric fields (e.g. currency, quantities, percentages), it’s often more readable to right-align the text. We need a way to configure the input’s text alignment.

## Background
In Slack, Victor Shulman asked whether it’s possible to `text-align: right` on `goab-input`. Jenna Granitto followed up, “Would it be possible to add that option eventually for when the user is inputting numbers?”

### Proposed solution

* Introduce an optional `align` (or `textAlign`) prop on `goab-input`
* Supported values: `left` (default) and `right`
* When `align="right"`, apply `text-align: right` to the native `<input>`

## Acceptance criteria

* `goab-input` API includes a new `align` prop accepting `left and right`
* Without `align` specified, input remains left-aligned
* With `align="right"`, input text is right-aligned in all supported browsers
* Unit and visual regression tests cover each alignment option
* Documentation example added showing numeric input with right alignment
* functionality add to LTS is also required

syedszeeshan:
docs: https://github.com/GovAlta/ui-components-docs/pull/383

PR #2791, opened 2025-06-10
fix(#2722): LTS - add text-align option to goa-input

syedszeeshan:
# Before (the change)

# After (the change)

### DIFF
--- a/libs/react-components/src/lib/input/input.spec.tsx
+++ b/libs/react-components/src/lib/input/input.spec.tsx
@@ -144,5 +144,19 @@ describe("Input", () => {
       })
     );
     expect(mockOnChangeHandler).toBeCalledWith("numberInput", decimalValue);
-  })
+  });
+
+  describe("Text Alignment", () => {
+    it("passes textAlign prop through to web component", () => {
+      const props: GoAInputProps = {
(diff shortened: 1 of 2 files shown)


libs/web-components/src/components/input/Input.svelte
```diff
@@ -2,6 +2,7 @@
 
 <script lang="ts" context="module">
   export type GoAInputVariant = "goa" | "bare";
+  export type GoAInputTextAlign = "left" | "right";
```
chrisolsen, line 5: Any changes to the svelte file will cause version issues when publishing. So all LTS changes must be only on React code.

libs/web-components/src/components/input/Input.svelte
```diff
@@ -492,6 +501,15 @@
     cursor: pointer;
   }
 
+  /* Text alignment classes */
```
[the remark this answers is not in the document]
syedszeeshan: PR will be updated shortly.

libs/web-components/src/components/input/Input.svelte
syedszeeshan: PR will be updated shortly.

Issue #2766, opened 2025-05-29
Input onBlur property is wrong


ArakTaiRoth:
# Info

In our `common.ts`, we have `GoaInputOnBlurDetail`, that should be `GoabInputOnBlurDetail`

## Acceptance Criteria:

1. `GoaInputOnBlurDetail` should be changed to `GoabInputOnBlurDetail`
    * Create a type alias for this, to not make this a breaking change
    * Make a deprecated comment about `GoaInputOnBlurDetail`

PR #2798, opened 2025-06-10
fix(#2766): Added export for GoabInputOnBlurDetail

ArakTaiRoth:
# Before (the change)

You would have to use `GoaInputOnBlurDetail` for the `blur` event on the Input component

# After (the change)

Now you can use `GoabInputOnBlurDetail` instead, `GoaInputOnBlurDetail` also still exists so existing use cases aren't wrecked.

### DIFF
--- a/libs/common/src/lib/common.ts
+++ b/libs/common/src/lib/common.ts
@@ -11,12 +11,15 @@ export type GoabInputOnChangeDetail<T = string> = {
   name: string;
   value: T;
 };
+
+// @deprecated GoaInputOnBlurDetail has been deprecated. Use GoabInputOnBlurDetail instead.
+export type GoaInputOnBlurDetail = GoabInputOnBlurDetail;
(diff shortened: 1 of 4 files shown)


libs/common/src/lib/common.ts
```diff
@@ -13,6 +13,7 @@ export type GoabInputOnChangeDetail<T = string> = {
 };
 export type GoabInputOnFocusDetail<T = string> = GoabInputOnChangeDetail<T>;
 export type GoaInputOnBlurDetail<T = string> = GoabInputOnChangeDetail<T>;
```
chrisolsen, line 15: Can you mark the `Goa` variant as `@deprecated` and for better clarity use
```
export type GoaInputOnBlurDetail = GoabInputOnBlurDetail;
```

ArakTaiRoth: @chrisolsen This has been resolved

Issue #2666, opened 2025-04-24
table-sort-header background not covering full header


ArakTaiRoth:
# Info

Reported by Daryl C

When running a Table header in our Table component, that goes really long (either in Desktop or in Mobile, doesn't matter), if the header needs to wrap to a second line, the background-colour when the sortable header is hovered over, doesn't cover the whole header:
![Image](https://github.com/user-attachments/assets/50191fa9-ab0c-43bf-b1b8-978ff2a62100)

This is both an issue for Angular and React, and only tested on `goab` components.

## Code to Replicate

```
<goab-table width="100%" mb="xl" (onSort)="handleSort($event)">
  <thead>
    <tr>
      <th>
        <goab-table-sort-header name="firstName"
          >First name and really long header</goab-table-sort-header
        >
      </th>
      <th>Another really long header</th>
      <th>Age</th>
      <th>Random Column 1</th>
      <th>Random Column 2</th>
    </tr>
  </thead>
  <tbody>
    <tr *ngFor="let user of users; index as i">
      <td>{{ user.firstName }}</td>
      <td>{{ user.lastName }}</td>
      <td>{{ user.age }}</td>
      <td>
        <goab-popover [relative]="true" [target]="target">
          <p>This is a popover</p>
          It can be used for a number of different contexts.
          <ng-template #target>
            <goab-button type="secondary" size="compact">Click me</goab-button>
          </ng-template>
        </goab-popover>
      </td>
      <td>
        <goab-popover [relative]="true" [target]="target">
          <p>This is a popover</p>
          It can be used for a number of different contexts.
          <ng-template #target>
            <goab-button type="secondary" size="compact">Click me</goab-button>
          </ng-template>
        </goab-popover>
      </td>
    </tr>
  </tbody>
</goab-table>
```

```
this.users = [
    {
      firstName: "Christian",
      lastName: "Batz",
      age: 18,
    },
    {
      firstName: "Brain",
      lastName: "Wisozk",
      age: 19,
    },
    {
      firstName: "Neha",
      lastName: "Jones",
      age: 23,
    },
    {
      firstName: "Tristin",
      lastName: "Buckridge",
      age: 31,
    },
  ];
}

handleSort(event: GoabTableOnSortDetail) {
  const { sortBy, sortDir } = event;
  this.users.sort((a: any, b: any) => (a[sortBy] > b[sortBy] ? 1 : -1) * sortDir);
}
```

## Acceptance Criteria:

[...]

PR #2857, opened 2025-07-04
#2666: fix table-sort-header background not covering full header

BumbleB2na:
# Before (the change)

# After (the change)

image


syedszeeshan:
@BumbleB2na 
image

One minor issue with the height of the header background (see the screenshot, side by side, when mouse is hovered on the 2 columns)

twjeffery:
@ArakTaiRoth Can we update to extend the clickable area of the headers to be all the same size? (This would make them all larger when there is a wrapping header, that's ok with me.) I don't think that this wasn't accounted for in the initial design.

They should all be the same height so that the interaction is straight across at a consistent size and height across the table, like this:
image

chrisolsen:
Since all the headers are only one line I would think that the highlighted area was also one line, but in the following it differs from that.
image

chrisolsen:
And it seems that the issue Tom mentioned still exists
image

BumbleB2na:
> Since all the headers are only one line I would think that the highlighted area was also one line, but in the following it differs from that. 

Yes there are still problems with current PR that I've brought up in a PM but you were going to meet with Thomas to make a decision. Now that I think you are interested in the direction it's going I can put more work in to it.

twjeffery (dismissed):
@BumbleB2na So the whole header changes on hover, but only the heading text is clickable+hovers for the sort? I think that solution works for me for now, especially as this is covering a less common use case with long heading sizes.

Thanks!

twjeffery (approved):

Issue #2878, opened 2025-07-10
Strange behaviour with DatePicker type input with onChange


ArakTaiRoth:
# Info

## Angular

1. Anytime anything is changed, a console error appears "date.toLocaleDateString is not a function"
2. If you change the month, then make a change to day, month reverts back to January
3. Typing in the year input is very strange:
    * When day is selected (or changed), year is automatically populated with "0000"
    * When you then tab to year, it changes to "0"
    * When you start typing, it re-populates to "000x" where x is whatever number you just used
    * It then builds the year from right to left

## React

1. It's far more catastrophic, the whole page just disappears as soon as you make a change anywhere, and you get a console error, "value?.toISOString is not a function"

Can't test any of the other Angular issues because the whole page just stops working

## Acceptance Criteria:

1. Get the `onChange` functionality working with input type and making sure it fires only when the full date is filled out
2. Document the property
    * Property added to properties accordion
    * Sandbox updated with the property
    * Example created using it
3. Fix the Angular use case as with date and month selection
4. Fix the Angular use case with year selection (empty input default, fill in left to right)

PR #2886, opened 2025-07-14
fix(#2878): date picker input type

BumbleB2na:
# Before (the change)

It was not correctly validating the input type date picker on initialization from value prop or on change

# After (the change)

## Steps needed to test

Angular:
```ts
import {
  GoabBlock,
  GoabDatePicker,
  GoabDatePickerOnChangeDetail,
  GoabFormItem,
} from "@abgov/angular-components";
import { Component, CUSTOM_ELEMENTS_SCHEMA } from "@angular/core";
import { FormsModule } from "@angular/forms";
import { ReactiveFormsModule } from "@angular/forms";

@Component({
  standalone: true,
  selector: "abgov-date-picker",
  templateUrl: "./datePicker.component.html",
  imports: [FormsModule, ReactiveFormsModule, GoabBlock, GoabDatePicker, GoabFormItem],
  schemas: [CUSTOM_ELEMENTS_SCHEMA],
})
export class DatePickerComponent {
  constructor() {
    console.log("date picker component loaded");
  }

  jsDateNow = new Date();

  errors: string[] = [];

  dateOnChange(event: GoabDatePickerOnChangeDetail) {
    // show error when date is invalid after a change
    const elName = event.name;
    if (elName) {
      const hasError = event.value === null;
      if (hasError && !this.errors.includes(elName)) {
        this.errors = [...this.errors, elName];
      } else if (!hasError && this.errors.includes(elName)) {
        this.errors = this.errors.filter((e) => e !== elName);
      }
    }
  }
}
```
```html
<main>
  <goab-block direction="column" gap="m">
    <goab-form-item label="Date Picker">
      <goab-date-picker
        (onChange)="dateOnChange($event)"
        name="datePicker1"
        [error]="errors.includes('datePicker1')"
      ></goab-date-picker>
    </goab-form-item>
    <goab-form-item label="Date Picker and value in ISO format">
      <goab-date-picker
        (onChange)="dateOnChange($event)"
        name="datePicker2"
        [error]="errors.includes('datePicker2')"
        value="2025-07-07T07:25:44-06:00"
      ></goab-date-picker>
    </goab-form-item>
    <goab-form-item label="Date Picker and value in JavaScript Date.toString() format">
      <goab-date-picker
        (onChange)="dateOnChange($event)"
        name="datePicker3"
        [error]="errors.includes('datePicker3')"
        value="Mon Jul 07 2025 07:25:44 GMT-0600 (Mountain Daylight Time)"
      ></goab-date-picker>
    </goab-form-item>
    <goab-form-item label="Date Picker: input type with no value">
      <goab-date-picker
        (onChange)="dateOnChange($event)"
        name="datePicker4"
        [error]="errors.includes('datePicker4')"
        type="input"
      ></goab-date-picker>
    </goab-form-item>
    <goab-form-item label="Date Picker: input type with valid value">
      <goab-date-picker
        (onChange)="dateOnChange($event)"
        name="datePicker5"
        [error]="errors.includes('datePicker5')"
        value="2025-07-07"
        type="input"
      ></goab-date-picker>
    </goab-form-item>
    <goab-form-item label="Date Picker: input type with valid value format">
      <goab-date-picker
        (onChange)="dateOnChange($event)"
        name="datePicker6"
        [error]="errors.includes('datePicker6')"
        [value]="jsDateNow"
        type="input"
      ></goab-date-picker>
    </goab-form-item>
    <goab-form-item
      label="Date Picker: input type with invalid value format (not in YYYY-MM-DD format)"
    >
      <goab-date-picker
        (onChange)="dateOnChange($event)"
        name="datePicker7"
        [error]="errors.includes('datePicker7')"
        value="Mon Jul 07 2025 07:25:44 GMT-0600 (Mountain Daylight Time)"
        type="input"
      ></goab-date-picker>
    </goab-form-item>
    <!-- NOTE: In react this was failing but in angular it was not a problem: -->
    <goab-form-item
      label="Date Picker: input type with invalid value format (not in YYYY-MM-DD format)"
    >
      <goab-date-picker
        (onChange)="dateOnChange($event)"
        name="datePicker8"
        [error]="errors.includes('datePicker8')"
        value="2025-12345"
        type="input"
      ></goab-date-picker>
    </goab-form-item>
    <goab-form-item
      label="Date Picker: input type with invalid value (Feb 31st does not exist)"
    >
      <goab-date-picker
        (onChange)="dateOnChange($event)"
        name="datePicker9"
        [error]="errors.includes('datePicker9')"
        value="2025-02-31"
        type="input"
        testid="datePicker9"
      ></goab-date-picker>
    </goab-form-item>
  </goab-block>
</main>
```

React:
```tsx
import { GoabBlock, GoabDatePicker, GoabFormItem } from "@abgov/react-components";
import { GoabDatePickerOnChangeDetail } from "@abgov/ui-components-common";
import { useState } from "react";

export function DatePickerRoute() {
  const [errors, setErrors] = useState<string[]>([]);
  function onChange(event: GoabDatePickerOnChangeDetail) {
    // show error when date is invalid after a change
    const elName = event.name;
    if (elName) {
      setErrors((prev) => {
        const hasError = event.value === null;
        const errors = prev.filter((e) => e !== elName);
        return hasError ? [...errors, elName] : errors;
      });
    }
  }

  return (
    <main>
      <GoabBlock direction="column" gap="m">
        <GoabFormItem label="Date Picker">
          <GoabDatePicker
            onChange={onChange}
            name="datePicker1"
            error={errors.includes("datePicker1")}
            value={new Date(2025, 6, 7)}
          ></GoabDatePicker>
        </GoabFormItem>
        {/* Reproduced the react bug: */}
        <GoabFormItem label="Date Picker: input type, value in wrong format for input type">
          <GoabDatePicker
            onChange={onChange}
            name="datePicker2"
            error={errors.includes("datePicker2")}
            value={new Date("2025-12345")}
            type="input"
          ></GoabDatePicker>
        </GoabFormItem>
        {/* NOTE: This does not fail but, February 31st gets changed to March 3rd: */}
        <GoabFormItem label="Date Picker: input type, value provided is a date that does not exist (Feb 31st, although in react input type date picker it converts to March 3rd)">
          <GoabDatePicker
            onChange={onChange}
            name="datePicker3"
            error={errors.includes("datePicker3")}
            value={new Date("2025-02-31")}
            type="input"
          ></GoabDatePicker>
        </GoabFormItem>
        <GoabFormItem label="Date Picker: input type, no value">
          <GoabDatePicker
            onChange={onChange}
            name="datePicker4"
            error={errors.includes("datePicker4")}
            type="input"
          ></GoabDatePicker>
        </GoabFormItem>
      </GoabBlock>
    </main>
  );
}
```

### DIFF
--- a/libs/web-components/src/components/date-picker/DatePicker.svelte
+++ b/libs/web-components/src/components/date-picker/DatePicker.svelte
@@ -139,12 +139,22 @@
       return;
     }
 
-    if (type === "input") {
-      const [year, month, day] = value.split("-");
-      _inputDate = { year: year, month: `${+month - 1}`, day: day };
-    } else if (type === "calendar") {
+    if (type === "calendar") {
       _date = startOfDay(new Date(value));
+    } else if (type === "input") {
+      const [year = "", month = "", day = ""] = value.split("T")[0].split("-");
+
+      // save without padded zeroes
+      _inputDate = { year: `${+year}`, month: `${+month - 1}`, day: `${+day}` };
+
+      if (!isInputDateValid()) {
+        resetInputDate();
+      }
     }
+  }
+
+  function resetInputDate() {
+    _inputDate = { day: "", month: "", year: "" };
   }
 
   function onCalendarChange(e: CustomEvent) {
@@ -252,6 +262,32 @@
     e.stopPropagation();
   }
 
+  function getInputDateWithPaddedZeroes(): string {
+    return `${padLeft(_inputDate.year, 4, 0)}-${padLeft(+_inputDate.month + 1, 2, 0)}-${padLeft(_inputDate.day, 2, 0)}`;
+  }
+
+  function isInputDateValid(): boolean {
+    if (
+      _inputDate.year === "" ||
+      _inputDate.month === "" ||
+      _inputDate.day === ""
+    ) {
+      return false;
+    }
+
+    const date = getInputDateWithPaddedZeroes();
+    if (!new Date(date)?.getTime() || isNaN(new Date(date).getTime())) {
+      return false;
+    }
+
+    if (date !== new Date(date).toISOString().split("T")[0]) {
+      // E.g. "2025-02-31" would be invalid because the date does not exist
+      return false;
+    }
+
+    return true;
+  }
+
   // _change event handler for the text/dropdown inputs for the `input` date format
   function onInputChange(e: Event) {
     e.stopPropagation();
@@ -260,13 +296,9 @@
       e as CustomEvent<{ name: string; value: string }>
     ).detail;
 
-    _inputDate = { ..._inputDate, [elName]: +value };
+    _inputDate = { ..._inputDate, [elName]: value };
 
-    if (!new Date(+_inputDate.year, +_inputDate.month, +_inputDate.day)) {
-      return;
-    }
-
-    const date = `${padLeft(_inputDate.year, 4, 0)}-${padLeft(+_inputDate.month + 1, 2, 0)}-${padLeft(_inputDate.day, 2, 0)}`;
+    const date = isInputDateValid() ? getInputDateWithPaddedZeroes() : null;
 
     dispatch(
       _rootEl,
@@ -345,7 +377,8 @@
           value={_inputDate.day}
           min="1"
           max="31"
-          {_error}
+          {error}
+          testid="day-input"
         />
       </goa-form-item>
       <goa-form-item helptext="Year (YYYY)">
@@ -358,6 +391,7 @@
           min="1800"
           max="2200"
           {error}
+          testid="year-input"
         />
       </goa-form-item>
     </goa-block>
--- a/libs/web-components/src/components/date-picker/date-picker.spec.ts
+++ b/libs/web-components/src/components/date-picker/date-picker.spec.ts
@@ -5,8 +5,15 @@ import {
   render,
   waitFor,
 } from "@testing-library/svelte";
+import userEvent, { UserEvent } from "@testing-library/user-event";
 import { addDays, format, addMonths, addYears } from "date-fns";
 import { it, expect, vi } from "vitest";
+
+let user: UserEvent;
+
+beforeEach(() => {
+  user = userEvent.setup();
+});
 
 it("it renders", async () => {
   const { container } = render(DatePicker);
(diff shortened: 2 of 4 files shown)


BumbleB2na:
Regarding acceptance criterion:
> 1. Get the `onChange` functionality working with input type and making sure it fires only when the full date is filled out

What was done in this PR:
> 1. Get the `onChange` functionality working with input type ~and making sure it fires only when the full date is filled out~

Because calendar type date picker will always fire onChange whenever it closes or a change is made. Therefore I think input type date picker should do something similar, always firing onChange and returning `"YYYY-MM-DD"` when valid or `null` when invalid. Benefit to returning `null` is so that day or year inputs can be cleared.

BumbleB2na:
> @BumbleB2na @chrisolsen I can't actually replicate most of the issues described in this bug anymore on Alpha. The only issue I can still replicate is when you tab to the day and year input, it autofills with 0 first (which is not wanted). Which makes me think most of this PR likely isn't needed. Is anyone else able to confirm?

@ArakTaiRoth from Steps to Reproduce in this PR this demonstrates the react bug:
```
        {/* Reproduced the react bug: */}
        <GoabFormItem label="Date Picker: input type, value in wrong format for input type">
          <GoabDatePicker
            onChange={onChange}
            name="datePicker2"
            error={errors.includes("datePicker2")}
            value={new Date("2025-12345")}
            type="input"
          ></GoabDatePicker>
        </GoabFormItem>
```

libs/web-components/src/components/date-picker/DatePicker.svelte
```diff
@@ -252,6 +262,30 @@
     e.stopPropagation();
   }
 
+  function validateInputDate(): boolean {
+    if (
```
ArakTaiRoth, line 266: This if statement isn't necessary. It's already done in `setDate` on line 148.

libs/web-components/src/components/date-picker/DatePicker.svelte
```diff
@@ -132,18 +132,28 @@
...
+    const [year = "", month = "", day = ""] = value.split("T")[0].split("-");
+
+    if (year === "" || month === "" || day === "") {
+      _inputDate = { day: "", month: "", year: "" };
+    }
+
+    // save without padded zeroes
+    _inputDate = { year: `${+year}`, month: `${+month - 1}`, day: `${+day}` };
+
+    if (!validateInputDate()) {
```
ArakTaiRoth, line 155: Move this function call inside the if statement on line 148, that way we can remove the validation check I called out below.

libs/web-components/src/components/date-picker/DatePicker.svelte
```diff
@@ -132,18 +132,28 @@
...
-      _date = startOfDay(new Date(value));
+    // else type === "input"
+    const [year = "", month = "", day = ""] = value.split("T")[0].split("-");
+
+    if (year === "" || month === "" || day === "") {
+      _inputDate = { day: "", month: "", year: "" };
+    }
+
+    // save without padded zeroes
+    _inputDate = { year: `${+year}`, month: `${+month - 1}`, day: `${+day}` };
```
ArakTaiRoth, line 153: Run this inside an else statement for the above if statement

libs/web-components/src/components/date-picker/DatePicker.svelte
```diff
@@ -252,6 +262,30 @@
...
+    if (
+      _inputDate.year === "" ||
+      _inputDate.month === "" ||
+      _inputDate.day === ""
+    ) {
+      // input date is not complete
+      return false;
+    }
+
+    const date = `${padLeft(_inputDate.year, 4, 0)}-${padLeft(+_inputDate.month + 1, 2, 0)}-${padLeft(_inputDate.day, 2, 0)}`;
```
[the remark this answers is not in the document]
BumbleB2na: Added comment. Would prefer to return a bool and reason is that it's a common practice with form validation for multiple fields. Most of the AB design system components do not have multiple fields but in this case it does.

libs/web-components/src/components/date-picker/DatePicker.svelte
BumbleB2na: I know what you mean but it did not work because of how validation depends on what `_inputDate` is set to

libs/web-components/src/components/date-picker/DatePicker.svelte
BumbleB2na: I know what you mean but could not move to else statement because of how validation depends on what `_inputDate` is set to

libs/web-components/src/components/date-picker/DatePicker.svelte
BumbleB2na: I know what you mean but it is still needed for the other place it gets called:
```svelte
    const date = isInputDateValid()
      ? `${padLeft(_inputDate.year, 4, 0)}-${padLeft(+_inputDate.month + 1, 2, 0)}-${padLeft(_inputDate.day, 2, 0)}`
      : null;
```

libs/web-components/src/components/date-picker/date-picker.spec.ts
```diff
@@ -67,6 +74,135 @@ it("dispatches a value on date selection", async () => {
   });
 });
 
+it("dispatches date or null value when any field in input type date picker changes", async () => {
```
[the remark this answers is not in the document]
chrisolsen: The react browser tests are also testing the web-component and they should be created over the web-component unit tests.

libs/web-components/src/components/date-picker/DatePicker.svelte
```diff
@@ -252,6 +267,31 @@
     e.stopPropagation();
   }
 
+  function isInputDateValid(): boolean {
```
ArakTaiRoth, line 270: None of this validation is necessary on our end. All we have to check for is empty values in the input elements, and return a date once they all have something.

Let the users validate their own dates in whatever way they want to support.

BumbleB2na: After adding new test coverage I see what you mean and the validation has been simplified

libs/web-components/src/components/date-picker/date-picker.spec.ts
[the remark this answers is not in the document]
BumbleB2na: Moved that behavioral test to a react browser test. Much easier and was able to improve it.

libs/web-components/src/components/date-picker/DatePicker.svelte
```diff
@@ -260,13 +281,11 @@
...
+    _inputDate = { ..._inputDate, [elName]: `${value && +value}` };
 
-    if (!new Date(+_inputDate.year, +_inputDate.month, +_inputDate.day)) {
-      return;
-    }
-
-    const date = `${padLeft(_inputDate.year, 4, 0)}-${padLeft(+_inputDate.month + 1, 2, 0)}-${padLeft(_inputDate.day, 2, 0)}`;
+    const date = isInputDateValid()
+      ? `${padLeft(_inputDate.year, 4, 0)}-${padLeft(+_inputDate.month + 1, 2, 0)}-${padLeft(_inputDate.day, 2, 0)}`
+      : null;
```
chrisolsen, line 288: What is the reason this code is being changed? One issue it introduces is that a `_change` event is now dispatched when the date is invalid, which is not how things should work.

As for the `value && +value`, that change is also not needed as `value` is defaulted to a string value.

libs/web-components/src/components/date-picker/DatePicker.svelte
chrisolsen: This is a helper method, which I am ok with, but it differs from the code that it replaced. This function just validates that the values are not blank, where as the old function ensured that the values combined to make a valid date.

libs/web-components/src/components/date-picker/DatePicker.svelte
```diff
@@ -132,21 +132,36 @@
...
+      return;
+    }
+
+    // else type === "input"
+    const [year = "", month = "", day = ""] = value.split("T")[0].split("-");
+
+    if (year === "" || month === "" || day === "") {
+      resetInputDate();
       return;
     }
```
chrisolsen, line 151: This seems to be the same as line 156:157, is there a reason the helper function is not being used here.

This also seems like non-required refactoring, which makes the intention of the change hard to understand.

chrisolsen (changes requested):

libs/web-components/src/components/date-picker/DatePicker.svelte
BumbleB2na: The reason it was changed is because in my opinion it should fire a `_change` event when any of the 3 fields change, so that developers can handle invalid input. You don't have that problem with calendar type date picker because every date is valid, but when using input type date picker you are depending on what the user types.

In the PR description Steps to test you can see input type date picker wrapped in form item and made to be required. The reason this was done is to demonstrate why I think we should fire a `_change` event on any change.

Fires `_change` event if date becomes null/invalid:
image

Does not fire `_change` event when date becomes null/invalid:
image

Let me know if you still  want to prevent firing change when date is invalid.

libs/web-components/src/components/date-picker/DatePicker.svelte
BumbleB2na: I was able to remove this redundant block of code:
```
    if (year === "" || month === "" || day === "") {
      resetInputDate();
      return;
    }
```

libs/web-components/src/components/date-picker/DatePicker.svelte
BumbleB2na: Okay, thank you. Oddly enough it was possible to remove the validation, but let's bring it back to how it was 4 days ago before resolving PR feedback. I've kept the new test coverage.

libs/web-components/src/components/date-picker/DatePicker.svelte
BumbleB2na: > As for the value && +value, that change is also not needed as value is defaulted to a string value.

Okay I see what you mean and it's cleaned up.

libs/web-components/src/components/date-picker/DatePicker.svelte
```diff
@@ -132,21 +132,31 @@
...
-    } else if (type === "calendar") {
-      _date = startOfDay(new Date(value));
+    // else type === "input"
+    const [year = "", month = "", day = ""] = value.split("T")[0].split("-");
+
+    // save without padded zeroes
+    _inputDate = { year: `${+year}`, month: `${+month - 1}`, day: `${+day}` };
+
+    if (!isInputDateValid()) {
+      resetInputDate();
```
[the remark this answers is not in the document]
chrisolsen: > The problem was that at the start it was deciding if datepicker value was invalid for both calendar and input types:

That isn't what that block of is doing. That block of code is just checking to whether any input exists and if it doesn't it ensures that the two date values (input/calendar) are reset. So this block of code is needed.

chrisolsen (approved):

Issue #2858, opened 2025-07-04
Public form - sub form: When add a second child/sub item, it shows null and cannot move forward


vanessatran-ddi:
Example: https://stackblitz.com/~/github.com/vanessatran-ddi/ui-components-angular-sandbox/tree/lts-public-form-sub-form-bug?file=src/app/public-form/SupportOrderDetails.html
I am using the angular lts + latest web-components version on prod and it doesn't work when I try to add the 2nd child

<img width="812" height="756" alt="Image" src="https://github.com/user-attachments/assets/b480a2f8-71ef-4ae6-a2e4-03f08aa76c16" />
 
<img width="582" height="874" alt="Image" src="https://github.com/user-attachments/assets/76791a22-b96d-4df0-b903-095a1c123ff0" />
## Acceptance Criteria:
* Should be able to add a second sub item

PR #2908, opened 2025-07-22
fix(#2858): re register state for 2nd or more subform item

vanessatran-ddi:
# Before (the change)
When we add a 2nd sub item, and no matter what we fill in the value, it returns as null for the below error message:

image

Also when we add the second item, the text field always has the value from the first sub item.

# After (the change)

- [ ] When we add the 2nd item of the sub-form, the text field isn't being filled with the previous value
- [ ] No error and we can see 2nd item in the summary and table/state

https://github.com/user-attachments/assets/2dec0d15-5e00-4e80-b957-7dfe4d2066fb

## Steps needed to test
https://stackblitz.com/~/github.com/vanessatran-ddi/ui-components-angular-sandbox/tree/lts-public-form-sub-form-bug?file=src/app/public-form/SupportOrderDetails.html

We can re-use the above code to test.

### So what is the problem:

When we first load the page, this is how the state being initialized:

1
1. `Input.svelte` sends `FormFieldMountMsg` (`relay`)
2. `FormItem.svelte` sends `FormItemMountMsg`
3. `Fieldset.svelte` listens to `FormItemMountMsg` and triggers `onFormFieldMount(data as FormFieldMountRelayDetail)` which init the state 

```
function onFormFieldMount(detail: FormFieldMountRelayDetail) {
    const { name, el } = detail;
    if (!name) return;

    _formFields[name] = el;
  }
```

### Why isn't the `FormItemMountMsg` being sent for the second sub form item?
1. The `FormItem` isn't remounted again.
2. The SubForm just changes the `mode` in order to show the fieldset. 
3. The state becomes null and it throws an error at function https://github.com/GovAlta/ui-components/blob/alpha/libs/web-components/src/components/form/Fieldset.svelte#L293 

### What I fixed:

Because there is a miss-communication between `Fieldset.svelte` to create/init state and `FormItem.svelte` when 2nd onward subform item added, so I fill the gap by:  SubForm when adding a 2nd sub item -> Form -> Fieldset(s) -> remount itself and reset the input value (so no more auto-fill value from the previous sub-form item)

1. I created a new message `SubFormNewItemMsg`
2. SubForm.svelte I relay this message when I add a 2nd sub form item
3. Form.svelte I listen to this message and I will relay this `SubFormNewItemMsg`
4. Fieldset.svelte, I listen to this new message `SubFormNewItemMsg` and then I remount myself and clear the value (`FieldsetSetValueMsg`) so `Input.svelte` or `Dropdown.svelte`... will reset its value.

### DIFF
--- a/libs/web-components/src/components/form/Fieldset.svelte
+++ b/libs/web-components/src/components/form/Fieldset.svelte
@@ -38,6 +38,8 @@
     FormToggleActiveRelayDetail,
     FormPageContinueMsg,
     FormPageContinueRelayDetail,
+    SubFormNewItemMsg,
+    SubFormNewItemRelayDetail,
   } from "../../types/relay-types";
 
   // ======
@@ -117,6 +119,10 @@
           resetFields(event);
           event.stopPropagation();
           break;
+        case SubFormNewItemMsg:
+          onReRegisterSubFormNewItem(data as SubFormNewItemRelayDetail);
+          event.stopPropagation();
+          break;
       }
     });
   }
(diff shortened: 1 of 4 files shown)


chrisolsen:
This was something that worked before and was included in demos, so some recent change seems to have broken this.

Issue #2852, opened 2025-06-30
Can't use space in a filterable Dropdown


ArakTaiRoth:
# Info

Raised by Fahad Mahmood

When using the Filterable dropdown, you can't use the space bar. This could impact being able to search items that have a space in them, as you can't replicate that space in the search.

This can be replicated with any filterable dropdown, including on our documentation site.

## Acceptance Criteria:

1. Filterable Dropdown should accept space as an input

PR #2910, opened 2025-07-22
fix(#2852): cant use space in a filterable Dropdown

syedszeeshan:
# Before (the change)

# After (the change)
-Now accepts SPACE key
-Also fixed the focus issue (this was a pre-existing issue in prod, as soon as the space is entered, the focus goes all the way to the bottom)

## Steps needed to test

**Angular Playground**
```
<div [formGroup]="fb">
  <div style="margin-bottom:30px">
    <goab-form-item label="Country">
      <goab-dropdown
        name="genderId"
        width="40em"
        placeholder="--Select Country--"
        formControlName="countryDropdown"
        [filterable]="true"
      >
        @for (country of countries; track country.code) {
          <goab-dropdown-item [value]="country.code" [label]="country.name">
          </goab-dropdown-item>
        }
      </goab-dropdown>
    </goab-form-item>
  </div>
```

```
import { Component } from "@angular/core";
import {
  GoabDropdown,
  GoabDropdownItem,
  GoabFormItem,
  GoabButton,
} from "@abgov/angular-components";
import { FormControl, FormGroup, ReactiveFormsModule } from "@angular/forms";
import { Countries } from "playground/angular/src/app/dropdown/countries.data";

@Component({
  selector: "abgov-dropdown",
  templateUrl: "./dropdown.component.html",
  standalone: true,
  imports: [
    GoabFormItem,
    GoabDropdown,
    GoabDropdownItem,
    GoabButton,
    ReactiveFormsModule,
  ],
})
export class DropdownComponent {
  countries = Countries;
  fb!: FormGroup;

  constructor() {
    this.fb = new FormGroup({
      countryDropdown: new FormControl(null),
    });
  }

  onClick() {
    console.log("RESET from Playground");
    this.fb.reset();
  }
}
```

### DIFF
--- a/libs/web-components/src/components/dropdown/Dropdown.svelte
+++ b/libs/web-components/src/components/dropdown/Dropdown.svelte
@@ -763,6 +763,7 @@
       open={_isMenuVisible}
       padded="false"
       tabindex="-1"
+      filterablecontext={fromBoolean(_filterable)}
       on:_open={showMenu}
       on:_close={hideMenu}
     >
--- a/libs/web-components/src/components/popover/Popover.svelte
+++ b/libs/web-components/src/components/popover/Popover.svelte
@@ -67,6 +67,8 @@
   // border radius of popover window
   export let borderradius = "var(--goa-border-radius-m)";
 
+  export let filterablecontext: string = "false";
+
   // Private
 
   let _rootEl: HTMLElement;
@@ -81,6 +83,7 @@
   $: _padded = toBoolean(padded);
   $: _open = toBoolean(open);
   $: _disabled = toBoolean(disabled);
+  $: _filterableContext = toBoolean(filterablecontext);
 
   $: (async () => _open && (await setPopoverPosition()))();
   $: (async () => _sectionHeight && (await setPopoverPosition()))();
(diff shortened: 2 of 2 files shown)


syedszeeshan:
@dustin-nielsen-goa 
The issue found by Vanessa is actually an existing production issue. When the value of the dropdown item has a space, then hitting the tab key will not select the item.

This is not something I introduced with the changes in my PR.

The intent of this PR was to allow space (and fix the issue of focus/jumping around when space key is entered).

Are you ok to takle this issue in a separate story?

ArakTaiRoth:
@vanessatran-ddi @syedszeeshan So I did some testing with this issue. The issue can only be replicated if you don't include a `label` in `goab-dropdown-item`, and if your `value` also includes a space. I don't see this as a probable issue a team will run into, because I've never seen a team using DropdownItem without a label, and I suspect it would be even less probable to have a value with a space in it. My suggestion is to keep this in the back of our heads in case someone reports it, but I highly doubt it's an issue. All that to be said, even if it is an issue, it's definitely not a new one, it's just one that we couldn't test for previously with the inability to use space.

So to sum up, ignore this and move on.

ArakTaiRoth (approved):

Issue #2922, opened 2025-07-24
Form Stepper no longer working vertically


ArakTaiRoth:
# Info

Reported by Burak

The Form Stepper is supposed to go into vertical orientation when inside a small container or inside a mobile viewport:

<img width="359" height="366" alt="Image" src="https://github.com/user-attachments/assets/423f96d3-c8b9-4112-8d79-18153ddcc2e7" />

But this is what it's currently doing inside a small container:

<img width="459" height="321" alt="Image" src="https://github.com/user-attachments/assets/0769c979-0505-406f-8177-0b86f9ed4038" />

Currently if you decrease the viewport size to mobile, it works, but it doesn't work inside a container that's less than mobile viewport in width.

## More Info

It seems that the function `calcStepDimensions` is not working correctly when the FormStepper is inside a small container, it's not giving the correct dimensions in this case.

## Acceptance Criteria:

1. When the Form Stepper is inside a container less than mobile viewport width, it should show as vertical

PR #2928, opened 2025-07-25
fix(#2922): properly size Form Steps on mobile

bdfranck:
This PR addresses https://github.com/GovAlta/ui-components/issues/2922.

I noticed two issues:
1. The Form Stepper uses the Grid component to size the steps. But the grid uses the viewport, not the container, to decide if the columns should be stacked. A Stepper in a mobile-sized container but a desktop-sized viewport will appear broken.
2. The Form Step has an `_isMobile` variable which is only defined when the Form Stepper is resized. The variable is `undefined` on mount so the Form Step defaults to Desktop.

I decided to fix the issues with two changes:
1. Replace the Grid component with a custom CSS grid which uses the container query.
2. Fire the `form-stepper:resized` event on mount so that the `_isMobile` variable is defined on every Step.

### Before

1. The Stepper layout appears broken on mobile on load for mobile.
3. The layout also appears broken when the container is mobile sized but the viewport is desktop sized.

stepper-broken-boo

### After

1. The Stepper layout appears correct on load for mobile.
2. The layout also appears correct when the container is mobile sized but the viewport is desktop sized.

stepper-fixed-with-gap-better-yeah

### Test code

```html
<div style="margin: 12px; padding: 24px;width:360px;border:1px solid #ccc;">
    <goa-form-stepper step="step">
    <goa-form-step text="Crisis response and safety planning" status="complete"></goa-form-step>
    <goa-form-step text="Basic needs and accomodation" status="incomplete"></goa-form-step>
    <goa-form-step text="Case management and system needs"></goa-form-step>
    <goa-form-step text="Child, youth, and family support"></goa-form-step>
    <goa-form-step text="Review and submit"></goa-form-step>
    </goa-form-stepper>
    <goa-pages current="step" mb="3xl">
        <div>Page 1 content</div>
        <div>Page 2 content</div>
        <div>Page 3 content</div>
        <div>Page 4 content</div>
    </goa-pages>
</div>

```

### DIFF
--- a/libs/web-components/src/components/form-stepper/FormStepper.svelte
+++ b/libs/web-components/src/components/form-stepper/FormStepper.svelte
@@ -125,6 +125,8 @@
       }
 
       _bindTimeoutId = setTimeout(() => {
+        const width = _gridEl?.offsetWidth ?? 0;
+        dispatchStepResize(width);
         bindChildren();
         calcStepDimensions();
         addClickListener();
@@ -181,19 +183,7 @@
       calcStepDimensions();
 
       const width = entries[0].contentRect.width;
-
-      for (const step of _steps) {
-        step.el.dispatchEvent(
-          new CustomEvent("form-stepper:resized", {
-            bubbles: true,
-            composed: true,
-            detail: {
-              testid,
-              mobile: width < MOBILE_BP,
-            },
-          }),
-        );
-      }
+      dispatchStepResize(width);
     });
   }
 
@@ -251,6 +241,23 @@
       }),
     );
   }
+
+  function dispatchStepResize(width: number) {
+    if (!_steps.length) return;
+
+    for (const step of _steps) {
+      step.el.dispatchEvent(
+        new CustomEvent("form-stepper:resized", {
+          bubbles: true,
+          composed: true,
+          detail: {
+            testid,
+            mobile: width < MOBILE_BP,
+          },
+        }),
+      );
+    }
+  }
 </script>
 
 <div id="container">
@@ -272,9 +279,7 @@
       <progress class="vertical" value={_progress} max="100"></progress>
     {/if}
     <div class="slots" bind:this={_gridEl}>
-      <goa-grid minchildwidth="10ch">
         <slot />
-      </goa-grid>
     </div>
   </div>
 </div>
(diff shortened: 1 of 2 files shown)


libs/web-components/src/components/form-stepper/FormStepper.svelte
```diff
@@ -321,8 +324,13 @@
     .form-stepper {
       position: relative;
     }
-  }
 
+    .slots {
+      display: grid;
+      grid-template-columns: repeat(auto-fit, minmax(10ch, 1fr));
```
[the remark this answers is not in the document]
bdfranck: The original code used an unchanging value for the `minchildwidth` property. A future enhancement could be to create a component design token for this value. But I can't think of a use case so I'll leave it hard coded for now.

libs/web-components/src/components/form-stepper/FormStepper.svelte
```diff
@@ -321,8 +326,13 @@
     .form-stepper {
       position: relative;
     }
-  }
 
+    .slots {
```
ArakTaiRoth, line 330: All of your CSS changes are literally just copying the CSS included in `Grid.svelte`. The only difference is that in `Grid.svelte` the CSS for `.goa-grid` aren't applied in a mobile context, which is probably why it didn't work for this use case. Could we not just change that in `Grid.svelte`?

bdfranck: I'm hesitant to outright change it in the Grid component itself because that would change how it behaves for existing team implementations (Assuming any delivery team uses this component).

It's one of those tiny code changes that could have a huge impact on existing layouts. I'm envisioning lots of "we updated our components and it broke our layout" complaints.

I'm curious if people want a version of the Grid that is calculated by a container query instead of a media query. If so, we could create a feature ticket to add this functionality. Then we could implement the new feature here.

syedszeeshan:
Passed.
https://github.com/user-attachments/assets/a8a09677-94a4-403b-8bf2-c1ab4a1043dc

libs/web-components/src/components/form-stepper/FormStepper.svelte
bdfranck: If we decide to change the Grid component to a container query then we can easily implement it here again.

PR #2934, opened 2025-07-29
fix: a dropdown list item and a radio item should be able to bind to a something other than a string

grantnicol:
Many projects are having to HACK this constraint.

# Angular
```html
      <goab-form-item label="Size">
        <goab-dropdown formControlName="parentDropdown" name="parent">
          <goab-dropdown-item *ngFor="let parent of parents" [value]="parent"/>
        </goab-dropdown>
      </goab-form-item>

      <!-- This doesn't work as expected -->
      <goab-checkbox name="num" text="Number" [formControl]="numericCheckbox"></goab-checkbox>
      <div>Binding: {{ numericCheckbox.value }}</div>


```

```ts
export class ComponentWrapperPageComponent {

  parents = [1, 2, 3, 4, 5, 6, 7];

  numericCheckbox = new FormControl<number>(1);
}

```

# Before (the change)

- You could only bind strings to dropdown list items
- You could only bind strings to radio list items

# After (the change)

- Now you can bind numbers to dropdown list items. example: courtroom ids
- Now could only bind booleans to radio list items. example Yes [true], No [false]

- [x] I have tested the functionality in Angular.

### DIFF
--- a/libs/react-components/src/lib/dropdown/dropdown-item.tsx
+++ b/libs/react-components/src/lib/dropdown/dropdown-item.tsx
@@ -2,7 +2,7 @@ import { useEffect } from "react";
 import { GoabDropdownItemMountType } from "@abgov/ui-components-common";
 
 interface WCProps {
-  value: string;
+  value: string | number;
   label?: string;
   filter?: string;
   mount?: GoabDropdownItemMountType;
--- a/libs/react-components/src/lib/radio-group/radio.tsx
+++ b/libs/react-components/src/lib/radio-group/radio.tsx
@@ -4,7 +4,7 @@ import type { JSX } from "react";
 
 interface WCProps extends Margins {
   name?: string;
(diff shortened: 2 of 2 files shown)


chrisolsen:
Seems to work well with numbers, but not so well with booleans as there is a blank value

https://github.com/user-attachments/assets/85c7edd2-f1db-4e9d-a210-10d4f2e0ca26

grantnicol:
@chrisolsen I have removed the boolean (it is ok not to populate a dropdown item with booleans -- could be overkill). I have also updated the events.

chrisolsen:
In the example code the numeric checkbox example doesn't work the way I would expect, in that the "value" returned after the first click is the `checked` status. 

https://github.com/user-attachments/assets/0a4d0ecf-cdf9-433c-ab51-639b2e57d7fb

Issue #2478, opened 2025-03-17
Workspace Side Menu:  New component


bdfranck:
Create a new workspace-specific side menu component based on  https://github.com/GovAlta/ui-components-design/issues/47.

### Acceptance criteria

- Aligns with [Figma design](https://www.figma.com/design/Jpy1Ea5qglwnp1SgGnagY9/%E2%9D%96-Component-library-BETA?node-id=55710-667994&t=y8EVuS0W47XCbVCB-4).

- Has a title for the service's name
- Has a URL for the service's home address
- Has account info for the current user:
  - Name
  - Email
- Has primary, secondary, and account menus
  - Accepts one or more menu items
  - Hidden if has no items
- Each menu item has:
  - A label
  - A URL
  - A mandatory icon* (The user may not supply an icon but then they will not be able to collapse the menu)
  - An optional badge
    - "Always visible" property to show the badge in the collapsed state
    - Normal, important, emergency, and success variants
- User profile menu
  - Slot for Display name (optional)
  - Slot for email address (optional)

- Meets accessibility requirements:
  - All interactive elements have the appropriate ARIA labels
  - All interactions can be done by keyboard
  - Has a logical tab order

Spark450:
This issue is blocked until we have further information about the direction being taken in design system evolution.

PR #2937, opened 2025-07-31
WIP feat(#2478): add work side menu component

bdfranck:
This PR adds the work side menu component.

https://github.com/user-attachments/assets/7406430a-e93c-47dc-bd86-86f00a3470a8

### To do
- [ ] Decide component name
- [ ] Decide property names
- [x] Create react wrapper
- [x] Create angular wrapper
- [ ] Ensure that angular and react align with v4 changes
- [x] Add initial component tokens
- [x] Add ARIA labels
- [ ] Write tests
- [ ] Write docs
- [x] Rename commit message to include issue number
- [ ] Test different variations
- [x] Cross-browser testing

### How to test

Add one of the following menus to your playground:

#### Empty menu
```svelte
<goa-work-side-menu
  url="#"
  title="Service name"
  username="Name"
  useremail="test@example.com"
>
</goa-work-side-menu>
``` 

#### Menu with primary, secondary, and account items
```svelte
<goa-work-side-menu
  url="#"
  title="Service name"
  username="Name"
  useremail="test@example.com"
>
  <div slot="primary">
    <goa-work-side-menu-item
      icon="triangle"
      label="Primary menu item"
      url="#primary"
    />
  </div>
  <div slot="secondary">
    <goa-work-side-menu-item
      icon="triangle"
      label="Secondary menu item"
      url="#secondary"
    />
  </div>
  <div slot="account">
    <goa-work-side-menu-item
      icon="triangle"
      label="Account menu item"
      url="#secondary"
    />
  </div>
</goa-work-side-menu>
``` 

#### Menu with primary , secondary, account items, and badges
```svelte
<goa-work-side-menu
  url="#"
  title="Service name"
  username="Name"
  useremail="test@example.com"
>
  <div slot="primary">
    <goa-work-side-menu-item
      icon="triangle"
      label="Primary menu item"
      url="#primary"
      badge="New
    />
  </div>
  <div slot="secondary">
    <goa-work-side-menu-item
      icon="triangle"
      label="Secondary menu item"
      url="#secondary"
      type="emergency"
      badge="12"
    />
  </div>
  <div slot="account">
    <goa-work-side-menu-item
      icon="triangle"
      label="Account menu item"
      url="#secondary"
    />
  </div>
</goa-work-side-menu>
``` 

#### A full example with a sub menu

```svelte
<goa-work-side-menu
  url="#client"
  title="Income and Employment Support (IES)"
  username="Edna Mode"
  useremail="edna.mode@example.com"
>
  <div slot="primary">
    <goa-work-side-menu-item
      icon="search"
      label="Search"
      badge="21"
      url="#search"
    >
      <goa-work-side-menu-item
        url="#sub1"
        variation="submenu"
        label="Sub menu item"
      >
      </goa-work-side-menu-item>

      <goa-work-side-menu-item
        url="#sub2"
        variation="submenu"
        label="Sub menu item"
      />

      <goa-work-side-menu-item
        url="#sub3"
        variation="submenu"
        label="Sub menu item"
      />
    </goa-work-side-menu-item>

    <goa-work-side-menu-item
      icon="list"
      label="Clients"
      type="success"
      badge="New"
      current="true"
      url="#clients"
    />

    <goa-work-side-menu-item
      icon="calendar"
      label="Schedule"
      type="emergency"
      badge="Urgent"
      url="#schedule"
    />
  </div>

  <div slot="secondary">
    <goa-work-side-menu-item
      icon="notifications"
      label="Notifications"
      type="success"
      badge="1"
      badgeLabelType="alwaysvisible"
      url="#notifications"
    />

    <goa-work-side-menu-item icon="help-circle" label="Support" url="#support" />

    <goa-work-side-menu-item icon="settings" label="Settings" url="#settings" />
  </div>

  <div slot="account">
    <goa-work-side-menu-item
      icon="person"
      label="Account management"
      url="#account"
    />

    <goa-work-side-menu-item icon="log-out" label="Log out" url="#logout" />
  </div>
</goa-work-side-menu>
```


libs/web-components/src/components/work-side-menu/WorkSideMenu.svelte
```diff
@@ -0,0 +1,691 @@
...
+  import { MOBILE_BP } from "../../common/breakpoints";
+  import { dispatch } from "../../common/utils";
+  import { isUrlMatch, getMatchedLink } from "../../common/urls";
+  import { onMount, onDestroy, tick } from "svelte";
+
+  // ******
+  // Public
+  // ******
+
+  export let title: string;
```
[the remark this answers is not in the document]
bdfranck: I've changed the property to `heading` to match the App Header Menu.

libs/react-components/src/lib/work-side-menu-item/work-side-menu-item.spec.tsx
```diff
@@ -0,0 +1,19 @@
+import { render } from "@testing-library/react";
```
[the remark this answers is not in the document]
chrisolsen: @bdfranck re: alignment => Yes, things would shift/padding would be reduced by a couple of pixels.

bdfranck:
@chrisolsen I addressed all your comments and pushed an amended commit. I've also added browser tests for testing events in the menu and menu item. Does the location of the Angular `experimental` folder make sense? Any further questions or concerns? Thanks!

libs/web-components/src/components/work-side-menu/WorkSideMenuWrapper.test.svelte
```diff
@@ -0,0 +1,33 @@
+<svelte:options customElement="test-work-side-menu-wrapper" />
```
chrisolsen, line 1: We don't need the *wrapper* tests any more now that we have the browser tests.

libs/web-components/src/components/work-side-menu/WorkSideMenuItem.svelte
```diff
@@ -0,0 +1,334 @@
...
+      position: absolute;
+      height: var(--goa-space-m);
+      width: var(--goa-space-m);
+      min-width: auto;
+      padding: 0;
+      top: var(--goa-space-3xs);
+      right: var(--goa-space-3xs);
+      text-align: center;
+      line-height: 1.2;
+      font-size: var(--goa-font-size-1, 12px);
```
chrisolsen, line 306: Is there a reason there is a fallback value for this one? If so, it should probably fall back to a `rem` value.

libs/web-components/src/components/work-side-menu/WorkSideMenuItem.svelte
```diff
@@ -0,0 +1,334 @@
...
+    gap: var(--goa-space-3xs);
+    border-left: var(--goa-space-2xs) solid var(--goa-color-greyscale-100);
+    margin: var(--goa-space-xs) 0 0 var(--goa-space-m);
+    padding: 0 0 0 var(--goa-space-s);
+  }
+
+  /* Badge */
+  .badge {
+    color: var(--goa-color-text-light);
+    height: 20px;
```
chrisolsen, line 254: We need to change a lot of these px values to rem.

libs/react-components/specs/work-side-menu-item.browser.spec.tsx
```diff
@@ -0,0 +1,63 @@
+import { render } from "vitest-browser-react";
```
chrisolsen, line 1: I am unsure we want to use browser tests for testing *child* components.

libs/web-components/src/components/work-side-menu/WorkSideMenu.svelte
```diff
@@ -0,0 +1,748 @@
...
+  import { onMount, onDestroy, tick } from "svelte";
+
+  // ******
+  // Public
+  // ******
+
+  export let heading: string;
+  export let url: string;
+  export let username: string;
+  export let useremail: string;
```
chrisolsen, line 20: As per some other designs I have seen we need something less specific here as in the other examples I believe I saw the users `role` being used.

libs/web-components/src/components/work-side-menu/WorkSideMenu.svelte
```diff
@@ -0,0 +1,748 @@
...
+  } from "../../common/urls";
+  import { onMount, onDestroy, tick } from "svelte";
+
+  // ******
+  // Public
+  // ******
+
+  export let heading: string;
+  export let url: string;
+  export let username: string;
```
chrisolsen, line 19: For the properties, we should probably start shifting over to using the kebab convention. This will require some changes to the `customElement` prop at the top of the page. Check out the Link.svelte file for an example on how to do this.

bdfranck: I've added kebab case for both user attributes.

libs/web-components/src/components/work-side-menu/WorkSideMenuItem.svelte
bdfranck: I've removed every `px` value that has a relevant token value.

twjeffery (changes requested):
@bdfranck This component is looking great. A few things for us to consider:

**1. Back Button / Secondary Navigation:** When the side menu is used as a secondary nav for something like a _detail view_, we add a _back_ menu button that's a bit spaced out away from the rest of the main menu items. Should we add this as a property for teams to be able to do this easily? Either as a type of menu item or as a prop on the menu itself (such as `has-back-button` or `is-secondary`)

image


**2. Submenu Expand/Collapse Behavior**
 - Currently submenus only show when parent item is active. (auto-revealed on navigation). It makes it a bit hard to know where the sub items are until you reveal that item. Should we add a chevron to toggle the menu open and closed? 
    - Perhaps something like this, but it's not fully resolved
    
      image

**3. Menu State Persistence During Navigation**
 - Should collapsed/expanded state persist when user navigates between pages? (Perhaps with the exception if there are sub items). Currently the menu opens whenever you navigate, which feels a little cumbersome if you've just collapsed the menu.

**4. The keyboard focus** does not enter into the Account/profile menu when it opens. And pressing `Escape` should close it

---
Here is a jam.dev of what I checked and adjusted while testing:
https://jam.dev/c/c212d2f9-e99e-4d1c-a72f-59fa5721e74e

libs/web-components/src/components/work-side-menu/WorkSideMenu.svelte
```diff
@@ -0,0 +1,800 @@
...
+    display: flex;
+    overflow: hidden;
+    align-items: center;
+    flex-shrink: 0;
+    border: var(
+      --goa-work-side-menu-border,
+      var(--goa-border-width-s) solid var(--goa-color-greyscale-200)
+    );
+    background-color: transparent;
+    padding: var(--goa-space-xs) var(--goa-space-2xs) var(--goa-space-xs)
```
twjeffery, line 580: Adjust to `padding: var(--goa-space-xs);` to match closed spacing to avoid shift on expand and collapse, and icons line up in both states.

image

image

twjeffery:
@bdfranck

**1. Back button** 
Sounds good

**2. Submenu items**
Sounds good

**3. Menu state persistence**
For sure, it might be a playground issue that wouldn't be an issue for teams. I am using multiple pages in my playground.

**4. Keyboard navigation for the account menu**
For sure. Maybe this should be pulled out to a separate issue for the popover?

twjeffery (changes requested):
Great work @bdfranck, I pulled in the changes and noticed a few more small things, added as comments

libs/angular-components/src/index.ts
```diff
@@ -1,3 +1,4 @@
 export * from "./lib/angular-components.module";
 export * from "./lib/components";
 export * from "@abgov/ui-components-common";
+export * from "./experimental";
```
chrisolsen, line 4: Having this line here will bundle the new experimental component with all the other components. In the `dist/libs/angular-components` folder is there an `experimental/index.js` file? If so this line should be able to be removed.

twjeffery (approved):
@bdfranck Just one small potential update, looks good

libs/react-components/src/experimental/work-side-menu/work-side-menu.tsx
```diff
@@ -0,0 +1,78 @@
...
+interface WCProps {
+  heading: string;
+  url: string;
+  "user-name": string;
+  "user-secondary-text": string;
+  testid?: string;
+  primaryContent?: ReactNode;
+  secondaryContent?: ReactNode;
+  accountContent?: ReactNode;
+  open?: boolean;
```
vanessatran-ddi: If we want to pass `open` as boolean, then under WorkSideMenu.svelte we need to assign it as type boolean, and by default `let open = false;` (you can refer Drawer.svelte), otherwise you can assign `open` as string, and follow the way `Modal.svelte` and `modal.tsx` doing, otherwise, the `open` is always `true` here. I need to add some code to make it works locally 
image

vanessatran-ddi:
> @vanessatran-ddi I've amended my commit with the following changes:
> 
> * Added `experimental` to the Angular imports
> * Added `aria-expaned` to the account menu
> * Improved keyboard navigation
>   
>   * Works on entire menu
>   * Loops when reaching the bottom or top
>   * Opens the account menu when navigating to it
>   * Closes the account menu when navigating from it
> * Fixed the border between the header and main menu when scrolling (an issue that @Spark450 noted)
> 
> keyboard-nav [ keyboard-nav ](https://private-user-images.githubusercontent.com/1479091/503270418-5a2ece2b-2d6c-4ebf-82c4-f75c2d3273ee.gif?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NjA5OTg0NjcsIm5iZiI6MTc2MDk5ODE2NywicGF0aCI6Ii8xNDc5MDkxLzUwMzI3MDQxOC01YTJlY2UyYi0yZDZjLTRlYmYtODJjNC1mNzVjMmQzMjczZWUuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI1MTAyMCUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNTEwMjBUMjIwOTI3WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9OTFiOGNmMjRkYjg4NGEwMWM0MDA3ZTQxMjE0ZDg4N2FkMzcwOGFmMzU5MTg1MThkNzI4Y2NjNDhhZGNiYjlkYSZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ._h200WseDoA41pZZaxuXclIKaUgVFnnz2isrQN7igVk) [ ](https://private-user-images.githubusercontent.com/1479091/503270418-5a2ece2b-2d6c-4ebf-82c4-f75c2d3273ee.gif?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NjA5OTg0NjcsIm5iZiI6MTc2MDk5ODE2NywicGF0aCI6Ii8xNDc5MDkxLzUwMzI3MDQxOC01YTJlY2UyYi0yZDZjLTRlYmYtODJjNC1mNzVjMmQzMjczZWUuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI1MTAyMCUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNTEwMjBUMjIwOTI3WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9OTFiOGNmMjRkYjg4NGEwMWM0MDA3ZTQxMjE0ZDg4N2FkMzcwOGFmMzU5MTg1MThkNzI4Y2NjNDhhZGNiYjlkYSZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ._h200WseDoA41pZZaxuXclIKaUgVFnnz2isrQN7igVk)
> 
> image

Hi @bdfranck , 
I have this comment to follow up vs you: 

Current behavior: Arrow Down from "Edna Mode" automatically expands the submenu and focuses on "Account management" → "Log out" before continuing to "Collapse Menu".

Expected behavior: Arrow Down from "Edna Mode" should go directly to "Collapse Menu" without auto-expanding the submenu. The user should need to press Enter/Space on "Edna Mode" to expand and access the submenu items.

libs/react-components/specs/work-side-menu.browser.spec.tsx
```diff
@@ -0,0 +1,189 @@
...
+            secondaryContent={<GoaxWorkSideMenuItem url="#item2" label="Item 2" />}
+            accountContent={<GoaxWorkSideMenuItem url="#item3" label="Item 3" />}
+            open={true}
+          />
+        );
+      };
+      const result = render(<Component />);
+
+      await vi.waitFor(() => {
+        expect(result.getByTestId("work-side-menu")).toBeTruthy();
```
chrisolsen: As per the previous comment this can be moved out of this waitFor block

bdfranck:
Here's a summary of some decisions I made for this component:
- I chatted with Tom about the issue where the tooltip and popover menu don't appear aligned if the menu is placed in an unconventional position. The only valid use of the component is if it's placed in the top left. We couldn't come up with a valid scenario of using it in an unconventional position. We decided to leave the issue and return to it later if a scenario arises.
- I removed the submenu items and created a separate ticket for a [side menu group](https://github.com/GovAlta/ui-components/issues/3137). I decided this because the existing submenu had accessibility and interaction issues. It was attempting to do the job of both a single menu item and a menu group at the same time. A dedicated menu group component felt like a better approach.

libs/react-components/specs/work-side-menu.browser.spec.tsx
```diff
@@ -0,0 +1,189 @@
...
+            open={true}
+          />
+        );
+      };
+      const result = render(<Component />);
+
+      // Wait for component to be fully rendered
+      await vi.waitFor(() => {
+        expect(result.getByTestId("work-side-menu")).toBeTruthy();
+      });
```
chrisolsen: No need for the waitFor and expect check. This applies to all other similar code blocks, so I will stop commenting on these :)

bdfranck:
@chrisolsen I've amended my commit with the following changes:
- Changed all instances of `goax` to `goabx`
- Removed all the unneeded `waitFor`s in the browser test
- Rebased from `alpha`

libs/angular-components/src/experimental/work-side-menu/work-side-menu.ts
```diff
@@ -0,0 +1,53 @@
+import {
```
vanessatran-ddi: QA:
I tested on angular with the following case:
1. I tab to focus on the workspace side menu
2. I press Arrow down
3. I expected to see my focusing menu item to be the first menu "Dashboard" but in fact it is "John Doe" and when I press Arrow down again, it comes to Collapse Menu. Arrow up and down only moving between John Doe & Collapse Menu

I added some console.log on my local machine, and I think  for angular, you will need to add a small delay (refer how other angular components doing such as drawer.ts), otherwise Svelte will `onMount` too quickly, that doesn't register all `goabx-work-side-menu-item` correctly to the `_focusItems`

image


https://github.com/user-attachments/assets/5cbe9e37-aa2b-446b-9780-c692fe41fdd2

vanessatran-ddi: Also for Angular, when I press Enter to open the User's popover:
image

In React, if I arrow up and down, in some moments, I can focus on the menu items inside the user's popover. For Angular, I press Arrow up and down, no matter how many times I press, I never have a chance to focus on the menu item inside the user's popover. I think the problem is similar to `missing a delay` for angular to be able to register vs Svelte's onMount

libs/angular-components/src/experimental/work-side-menu/work-side-menu.ts
```diff
@@ -0,0 +1,69 @@
...
+  TemplateRef,
+  EventEmitter,
+  OnInit,
+  ChangeDetectorRef,
+} from "@angular/core";
+import { NgTemplateOutlet } from "@angular/common";
+@Component({
+  standalone: true,
+  selector: "goabx-work-side-menu", // eslint-disable-line
+  imports: [NgTemplateOutlet],
```
vanessatran-ddi: I have this error: 
image


I think you are missing CommonModule here.

libs/react-components/src/experimental/work-side-menu/work-side-menu.tsx
```diff
@@ -0,0 +1,78 @@
...
+      "goa-work-side-menu": WCProps & React.HTMLAttributes<HTMLElement>;
+    }
+  }
+}
+
+/* eslint-disable-next-line */
+export interface GoabWorkSideMenuProps {
+  heading: string;
+  url: string;
+  userName: string;
```
vanessatran-ddi: `heading` in angular is optional, but `heading` under react is mandatory, same vs `url`, `username`
We should follow Angular for React, `url` or `userName` shouldn't be mandatory.

libs/web-components/src/components/work-side-menu/WorkSideMenu.svelte
```diff
@@ -0,0 +1,895 @@
...
+
+  function hideTooltip() {
+    clearTimeout(_mouseLeaveTimeoutId);
+    _mouseLeaveTimeoutId = setTimeout(() => {
+      _showTooltip = false;
+    }, 300);
+  }
+
+  // Account menu
+  function closeAccountMenu(e: Event) {
```
vanessatran-ddi: This function is being called multiple times
image

However, I can see we have duplicated code under Escape: 
image


My recommendation is: if we have multiple places call this function to remove the `click` listener, there will be cases that we call and remove the listener that even not exists. We should add a check like what you did under `Escape` and replace the duplicated code vs closeAccountMenu

```
function closeAccountMenu(e?: Event) {
    if (_showAccountMenu) {
      _showAccountMenu = false;
      window.removeEventListener("click", closeAccountMenu);
    }
  }

```

libs/web-components/src/components/work-side-menu/WorkSideMenu.svelte
```diff
@@ -0,0 +1,895 @@
...
+  let _tooltipEl: HTMLElement;
+  let _profileButtonEl: HTMLElement;
+  let _toggleButtonEl: HTMLElement;
+  let _tooltipLabel: string = "";
+
+  let _bindTimeoutId: any;
+  let _mouseEnterTimeoutId: any;
+  let _mouseLeaveTimeoutId: any;
+
+  let _windowWidth = window.innerWidth;
```
vanessatran-ddi: This variable is not being used, we should remove.

libs/web-components/src/components/work-side-menu/WorkSideMenu.svelte
```diff
@@ -0,0 +1,895 @@
...
+      (_menuEl?.scrollHeight || 0) > (_scrollEl?.clientHeight || 0);
+  }
+
+  function toggleMenu() {
+    open = !open;
+    _showTooltip = false;
+  }
+
+  function getShadowLinkEl(el: Element): HTMLAnchorElement | null | undefined {
+    if (el.tagName.includes("GOABX")) el = el.firstElementChild as Element;
```
vanessatran-ddi: if `el.firstElementChild` is null -> we will force Typescript to treat `null` as `Element`. 
Perhaps: 
```
function getShadowLinkEl(el: Element): HTMLAnchorElement | null | undefined {
    if (el.tagName.includes("GOABX")) {
      const firstChild = el.firstElementChild;
      if (!firstChild) {
        return null;  // Early return if no child exists
      }
      el = firstChild;
    }

    return el?.shadowRoot?.querySelector("a");
  }

```

libs/web-components/src/components/work-side-menu/WorkSideMenu.svelte
```diff
@@ -0,0 +1,895 @@
...
+    }
+
+
+    tick().then(() => {
+      focusOnMenuItem(newItem);
+    });
+  }
+
+  function setFocusedIndexToElement(el: HTMLElement) {
+    _focusedIndex = _focusItems.indexOf(el) || -1;
```
vanessatran-ddi: This logic will be wrong if `indexOf(e)` returns the first `index` which is `0`
image

libs/web-components/src/components/work-side-menu/WorkSideMenu.svelte
```diff
@@ -0,0 +1,895 @@
...
+      case "Escape":
+        if (_showAccountMenu) {
+          _showAccountMenu = false;
+          window.removeEventListener("click", closeAccountMenu);
+        }
+        break;
+    }
+  }
+
+  function onArrow(e: KeyboardEvent, direction: "up" | "down") {
```
vanessatran-ddi: We can remove this `e`
image

libs/web-components/src/components/work-side-menu/WorkSideMenuItem.svelte
```diff
@@ -0,0 +1,289 @@
+<svelte:options customElement="goa-work-side-menu-item" />
+
+<script lang="ts">
+  import { onMount, onDestroy, tick } from "svelte";
+  import { dispatch, getSlottedChildren } from "../../common/utils";
```
vanessatran-ddi: We can remove `getSlottedChildren` since it is not being used.

libs/web-components/src/components/work-side-menu/WorkSideMenuItem.svelte
```diff
@@ -0,0 +1,289 @@
...
+  class="root"
+  class:divider
+  role="none"
+  data-testid={testid}
+  on:mouseenter={handleMouseEnter}
+  bind:this={_rootEl}
+>
+  <a
+    class="menu-item"
+    class:current
```
vanessatran-ddi: Consider to use `aria-current="page"` when `current` is true. 
Source: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-current

libs/web-components/src/components/work-side-menu/WorkSideMenuItem.svelte
```diff
@@ -0,0 +1,289 @@
...
+    role="menuitem"
+    href={url}
+    bind:this={_linkEl}
+    on:click={handleClick}
+    tabindex="0"
+  >
+    <goa-icon size="small" theme={current ? "filled" : "outline"} type={icon} />
```
vanessatran-ddi: Same here, add `arialabel` so when the menu is collapsed, the screen reader is still able to announce the menu. I suggest to use `label` as the value.

libs/react-components/src/experimental/work-side-menu/work-side-menu.tsx
vanessatran-ddi: Same here, we should make `userName` and `userSecondaryText` optional (Example our new design system website DS 2.0 will use this workspace menu and won't need `userName`, `userSecondaryText`.

libs/web-components/src/components/work-side-menu/WorkSideMenu.svelte
```diff
@@ -0,0 +1,906 @@
...
+          </button>
+        {/if}
+        <div class="toggle-menu">
+          <button
+            class="toggle-button"
+            data-testid="toggle-menu"
+            on:click={handleToggleClick}
+            bind:this={_toggleButtonEl}
+          >
+            <goa-icon
```
vanessatran-ddi: Missing the `ariaLabel` for "Collapse Menu" or "Expand Menu" here, otherwise when we focus on Expand/Collapse button, we only hear that "image, menu, collapse button".

libs/web-components/src/components/work-side-menu/WorkSideMenu.svelte
```diff
@@ -0,0 +1,907 @@
...
+  function watchPathChanges(action: () => void): MutationObserver {
+    let currentLocation = document.location.href;
+    const observer = new MutationObserver((_mutationList) => {
+      if (isUrlMatch(document.location, currentLocation)) {
+        currentLocation = document.location.href;
+        action();
+      }
+    });
+    observer.observe(document.body, { childList: true, subtree: true });
+    return observer; // Return the created observer
```
vanessatran-ddi: we can remove the comment here.

vanessatran-ddi (commented):
Several comments left. If you fix them let me know and I am ready to approve.

chrisolsen (approved):

Issue #2547, opened 2025-04-01
Popover is hidden if it is near notification banner (in Safari), less than Desktop size


vanessatran-ddi:
# Info

<img width="1501" alt="Image" src="https://github.com/user-attachments/assets/e2c6c23e-06f4-4aa9-8e79-b34b7dda0890" />

I am using Safari. I cannot see the option to select/switch the language. The language switcher is using `GoabPopover`.

<img width="1524" alt="Image" src="https://github.com/user-attachments/assets/43f88165-c486-4279-b209-6d5c3384caa2" />

Screenshot from Desktop safari showing something like an "afterimage" once popover is closed.

![Image](https://github.com/user-attachments/assets/90f15c07-9d10-486b-aed9-b2644ad79885)

Issue reproduced On Safari in iOS

## Acceptance Criteria

1. Popover should go above the Notification component (in Safari)

dustin-nielsen-goa:
Can't be replicated in either Chrome, or Firefox. I believe this is a Safari exclusive issue

PR #2950, opened 2025-08-06
fix(#2547): microsite header slotted with popover for safari

vanessatran-ddi:
# Before (the change)
Using Safari, the pop-over under the `slot` `version` of microsite header is hidden with `callout` banner. 

image

# After (the change)
image

## Steps needed to test
```
import {
  GoabButton,
  GoabNotification,
  GoabPopover,
  GoabTooltip,
  GoabIcon, GoabMicrositeHeader,
} from "@abgov/react-components";

export const Issue2547 = () => {
  return (
    <>
      <GoabMicrositeHeader
        type={"live"}
        feedbackUrl="https://forms.microsoft.com/r/8Zp7zSJS6W"
        maxContentWidth={"1440px"}
        version={
          <>
            <GoabTooltip content="Framework">
              <GoabPopover
                target={
                  <a className="version-language-switcher__heading" id="language-switcher">
                    <GoabIcon type="chevron-down" size="small"></GoabIcon>Angular
                  </a>
                }
                padded={false}>
                <>
                  {["angular", "react"].map(lang => (
                    <a key={lang} style={{
                      display: 'block',
                      font: 'var(--goa-typography-body-s)',
                      padding: 'var(--goa-space-s) var(--goa-space-s) var(--goa-space-s) var(--goa-space-s)',
                      textDecoration: 'none',
                      color: 'var(--goa-color-text-default)!important',
                      whiteSpace: 'nowrap'
                    }}>
                      {lang}
                    </a>))
                  }
                </>
              </GoabPopover>
            </GoabTooltip>
            <GoabPopover target={<GoabButton type="secondary" size="compact">
              Click me
            </GoabButton>} >
              <p>This is a popover</p>
              It can be used for a number of different contexts.
            </GoabPopover>
          </>
        } />


      <GoabNotification type="information">
        Notification banner message
      </GoabNotification>
    </>
  )
}

```

### DIFF
--- a/libs/web-components/src/components/microsite-header/MicrositeHeader.svelte
+++ b/libs/web-components/src/components/microsite-header/MicrositeHeader.svelte
@@ -141,7 +141,6 @@
 }
 
   #container {
-    container: self / inline-size;
     background-color: var(--goa-microsite-header-color-bg);
   }
 
@@ -185,13 +184,13 @@
     margin: 0 auto;
   }
 
-  @container self (--not-mobile) {
+  @media (min-width: 624px) {
     .content-container {
       padding: var(--goa-microsite-header-padding-medium-screen);
     }
   }
 
-  @container self (--desktop) {
+  @media (min-width: 1024px) {
     .content-container {
       padding: var(--goa-microsite-header-padding-large-screen);
     }


ArakTaiRoth:
@vanessatran-ddi @twjeffery So I tried testing this in both Safari 18.4 (Sequoia) and Safari 26 (Tahoe Beta). I can't see a difference between alpha and this PR. This is what I see in 18.4, on both alpha and this PR:
image

Which doesn't look completely accurate, but it's the same between alpha and this PR. In 26, it looks exactly like what we expect, but again, same between alpha and this PR.

I'm not sure if Browserstack is lying to me, or maybe this issue was seen on an older version of Safari?

twjeffery (approved):
@vanessatran-ddi this looks like it fixes the issue for me on an older version of safari

### old version
image

### with your update (working as expected)
image

ArakTaiRoth:
@vanessatran-ddi @twjeffery So I retested with version 17 and 18. This does indeed fix the issue of it appearing behind the container. But it does have other issues that I'm not sure if we care enough about, or not?

This is how it looks for me on Chrome:
image
So, that's what I'm comparing to.

This is what the the popover inside a tooltip looks for me in 17 and 18:
image
Note the lack of shadow or border around the popover that's in the Chrome screenshot. So it's not completely accurate.

In addition there's an issue in version 18 (that I don't see in version 17):
image
This is just a Popover with a button as the target, when the popover disappears, a shadow of it is still there, and stays there until I click something else interactive on the page.

Let me know if these are issues we want fixed, or not.

Issue #2802, opened 2025-06-11
Support for Angular 20


ArakTaiRoth:
# Info

Angular v20 was released on May 29th. We should be testing with this, introducing support, and adding it to our list of accepted versions.

## Acceptance Criteria:

1. Test with Angular v20 (and v16 and v17 - to figure out if we should remove support or not)
2. Add support where necessary (assuming it doesn't break previous versions)
3. Add it to the `package.json` as a supported version

## More Info

There have been problems found with both Dropdown and Text Area at least, outside of this team.

Below are tasks:

- [x] Upgrade `ui-components` and its dependencies for angular 20
- [x] Upgrade angular playground template to be compatible with angular 20
- [x] Test angular 20 for all components (Doing now)

Below are issues from Angular 20 that we need to fix and test again:

- [x] Run `npm install` with our components will have this dependency issue. Expected after this story we shouldn't have this error when `install`

<img width="778" height="953" alt="Image" src="https://github.com/user-attachments/assets/d0e3e806-b329-4f03-a5cc-6f98a830b2bd" />

### Dropdown:

- [x] Dropdown doesn't render label and value 

<img width="305" height="174" alt="Image" src="https://github.com/user-attachments/assets/786f3650-e1e8-4c19-8ffc-b4206d2e7751" />

- [x]  Dropdown filterable doesn't work if dropdown options fetched from backend 

<img width="395" height="318" alt="Image" src="https://github.com/user-attachments/assets/06e1fe15-1691-4608-b6d6-a1a0ce92c42c" />

- [x] Dropdown example https://design.alberta.ca/components/v4.0.0+-angular/dropdown#tab-1#dynamically-change-items-in-a-dropdown doesn't work anymore:

https://github.com/user-attachments/assets/38c0cc0a-bb99-44f1-b557-d685599940af

### Circular Progress:
- [x] Full screen isn't displayed correctly

https://github.com/user-attachments/assets/2ff295f7-1877-4154-9981-f5c9bdbbaf70

### Public form: 
- [x] A delay between angular text input with public form state that makes values always empty/undefined

<img width="1128" height="738" alt="Image" src="https://github.com/user-attachments/assets/f9e80d19-ddd5-47f6-ab03-51e1fbc862de" />

- [x] Radio doesn't work with public form

<img width="1651" height="441" alt="Image" src="https://github.com/user-attachments/assets/08b49dec-c922-4980-bfc2-be0fe667955a" />

[...]

PR #2974, opened 2025-08-15
feat(#2802):upgrade the repo with angular 20

vanessatran-ddi:
# Before (the change)

# After (the change)

## Steps needed to test

You can use my example repo for angular 20:
https://stackblitz.com/~/github.com/vanessatran-ddi/ui-components-angular-sandbox/tree/angular-20?file=package.json

When using angular 20 with our production libraries, you will see those errors that this PR is trying to fix. Errors have been recorded https://github.com/GovAlta/ui-components/issues/2802 description. 

You should symlink the package json of the angular 20 repo with your local build. Like this: 
image

This is for angular 19:
https://stackblitz.com/~/github.com/vanessatran-ddi/ui-components-angular-sandbox/tree/angular-19?file=package.json

Angular 17 is https://stackblitz.com/~/github.com/GovAlta/ui-components-angular-sandbox?file=package.json

## ⚠️ Still Required Changes ✅ - this has been updated 

The Angular CircularProgress.ts has been updated to allow for the additional code required to get the components working with Angular to reside within the Angular component. The changes in this file must also be done in the other Angular components as well. 

In addition, the Svelte components will also require that the existing `await tick()` function call within the `onMount` hook can be removed, as well as the `setTimeout` calls also within the `onMount` hook.

## ‼️ Temporarily this PR will not support angular 16, I talked with Chris and would confirm if angular 16 is removed from our support:
image

* if we still want to support angular 16, I will need to downgrade this PR to angular 18 (if you see package.json, we are upgrading to angular 20).

### DIFF
--- a/libs/angular-components/src/lib/components/index.ts
+++ b/libs/angular-components/src/lib/components/index.ts
@@ -3,6 +3,7 @@ export * from "./badge/badge";
 export * from "./block/block";
 export * from "./button/button";
 export * from "./button-group/button-group";
+export * from "./calendar/calendar";
 export * from "./callout/callout";
 export * from "./card/card";
 export * from "./card-content/card-content";


chrisolsen:
Can you squash the non-drawer commits into one, then having just 2 commits in this PR?

libs/web-components/src/components/checkbox/Checkbox.svelte
```diff
@@ -81,12 +81,10 @@
...
 
   onMount(() => {
-    // hold on to the initial value to prevent losing it on check changes
-    _value = value;
-    _descriptionId = `description_${name}`;
-
     addRelayListener();
     addRevealSlotListener();
+    _value = value;
+    _descriptionId = `description_${name}`;
```
chrisolsen: Do these need to be moved? The two methods above don't have anything to do with these properties. If they do need to be moved here then a comment explaining why is required.

libs/web-components/src/components/drawer/Drawer.svelte
```diff
@@ -192,7 +192,6 @@
       class:drawer-open-right={position === "right" && open}
       class:drawer-open-left={position === "left" && open}
       bind:this={_contentEl}
-      data-first-focus="true"
```
chrisolsen: Should this have been removed?

libs/web-components/src/components/dropdown/Dropdown.spec.ts
```diff
@@ -823,8 +823,8 @@ describe("GoADropdown", () => {
 
       expect(dropdown).toBeTruthy();
       expect(dropdown).toHaveStyle("margin-top:var(--goa-space-s)");
-      expect(dropdown).toHaveStyle("margin-bottom:var(--goa-space-m)");
-      expect(dropdown).toHaveStyle("margin-right:var(--goa-space-l)");
+      expect(dropdown).toHaveStyle("margin-bottom:var(--goa-space-l)");
+      expect(dropdown).toHaveStyle("margin-right:var(--goa-space-m)");
```
chrisolsen: If the tests are changed where is the change to the Dropdown file?

libs/web-components/src/components/input/Input.svelte
```diff
@@ -147,20 +147,18 @@
...
     validateTextAlign(textalign);
     addRelayListener();
     showDeprecationWarnings();
     checkSlots();
-    sendMountedMessage();
 
     const { containerStyle, inputWidth } = handleWidth(width, type);
     _containerStyle = containerStyle;
     _inputWidth = inputWidth;
+    sendMountedMessage();
```
chrisolsen: I don't like the moving of a function down a couple lines. If there is a timing issue and it does fix things it's doing so in an extremely brittle way.

libs/web-components/src/components/tooltip/Tooltip.svelte
```diff
@@ -98,15 +98,12 @@
...
-      validator(position, halign);
-      maxwidth = validateMaxWidth(maxwidth);
-    }, 1);
-
+    validatePosition(position);
+    validateAlignment(halign);
+    validator(position, halign);
     _initialPosition = position;
     _tooltipInstanceId = Math.random().toString(36);
+    maxwidth = validateMaxWidth(maxwidth);
```
chrisolsen: Another "timing" thing here.

libs/web-components/src/components/dropdown/Dropdown.spec.ts
vanessatran-ddi: I have no idea why it didn't fail in the past, but the expect statements in alpha are wrong. The below screenshot is after I fixed. Without fixing it, it will fail on CI. 
image

libs/web-components/src/components/tooltip/Tooltip.svelte
vanessatran-ddi: I also addressed this.

Issue #1321, opened 2023-08-14
"Submit" button type


twjeffery:
## Change suggested - for discussion:
We have different buttons types: (primary, secondary, tertiary, start), that are all different buttons visually, with different intended uses in design.
Currently we gave a submit button “type” that is visually the primary button, but with a submit attribute. By doing this, we’re mixing up button types and functions. In the same way, we wouldn’t have a “reset” button type, and a “search” button type as well alongside primary, secondary, and tertiary button types.

<img width="993" alt="image" src="https://github.com/GovAlta/ui-components/assets/82968683/e2b6c8fd-a9a3-41de-bd8a-c76019f3929f">

I suggest that we include it as a property instead of a button type on the same level as primary, secondary, tertiary. This seems to be a more common approach within other design systems.

Spark450:
@chrisolsen Is there anything in Svelte 4 that allows the "Submit" type to work as expected, should we still be documenting this type?

syedszeeshan:
UI PR: https://github.com/GovAlta/ui-components/pull/2988
docs PR: https://github.com/GovAlta/ui-components-docs/pull/426

PR #2988, opened 2025-08-25
fix(#1321): deprecate goa-button submit type

syedszeeshan:
# Before (the change)

# After (the change)
image

docs PR: https://github.com/GovAlta/ui-components-docs/pull/426

### DIFF
--- a/libs/web-components/src/common/utils.ts
+++ b/libs/web-components/src/common/utils.ts
@@ -174,9 +174,12 @@ export function validateRequired(
 export function typeValidator(
   message: string,
   values: string[],
-  required = false,
+  opts: boolean | { required?: boolean; deprecated?: string[] } = {},
 ): [string[], (value: string | null) => void] {
   const validator = (value: string | null): void => {
+    const required = typeof opts === "boolean" ? opts : opts?.required;
+    const deprecated = typeof opts === "object" ? opts?.deprecated : null;
+
     if (!required && !value) {
       return;
     }
@@ -186,6 +189,9 @@ export function typeValidator(
     }
     if (!values.includes(value || "")) {
       console.error(`[${value}] is an invalid ${message.toLowerCase()}`);
+    }
+    if (deprecated?.includes(value as string)) {
+      console.warn(`[${value}] is deprecated`);
     }
   };
   return [values, validator];
--- a/libs/web-components/src/components/button/Button.spec.ts
+++ b/libs/web-components/src/components/button/Button.spec.ts
@@ -56,7 +56,7 @@ describe("GoAButtonComponent", () => {
   });
 
   describe("type", () => {
-    ["primary", "submit", "secondary", "tertiary", "start"].forEach((type) => {
+    ["primary", "secondary", "tertiary", "start"].forEach((type) => {
       it(`should render ${type} type`, async () => {
         const baseElement = render(GoAButton, { type });
         const button = await baseElement.findByRole("button");
--- a/libs/web-components/src/common/utils.spec.ts
+++ b/libs/web-components/src/common/utils.spec.ts
@@ -1,5 +1,5 @@
 import { waitFor } from "@testing-library/svelte";
-import { getTimestamp, performOnce, announceToScreenReader } from "./utils";
+import { getTimestamp, performOnce, announceToScreenReader, typeValidator } from "./utils";
 import { it, describe, expect, vi, beforeEach, afterEach } from "vitest";
 
 describe("getTimestamp", () => {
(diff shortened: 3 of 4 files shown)


libs/angular-components/src/lib/components/button/button.ts
```diff
@@ -55,6 +56,12 @@ export class GoabButton extends GoabBaseComponent {
 
   @Output() onClick = new EventEmitter();
 
+  ngOnInit() {
+    if ((this.type as string) === "submit") {
```
ArakTaiRoth, line 60: Is this how we want to share deprecated state? I had thought we were going another way that doesn't affect the console, such as what's seen in `date-picker.ts` on line 52.

@chrisolsen

libs/react-components/src/lib/button/button.tsx
```diff
@@ -68,6 +68,13 @@ export function GoabButton({
 }: GoabButtonProps): JSX.Element {
   const el = useRef<HTMLElement>(null);
 
+  // Warn about deprecated submit type
+  useEffect(() => {
+    if ((type as string) === "submit") {
```
ArakTaiRoth, line 73: Same question as above, but this one you can reference `date-picker.tsx` on line 35

libs/web-components/src/components/button/Button.spec.ts
```diff
@@ -56,7 +56,7 @@ describe("GoAButtonComponent", () => {
   });
```
ArakTaiRoth, line 56: This isn't for this file, but I notice we also didn't add a Deprecated message in `Button.,svelte`, something that can be seen in `DatePicker.svelte` on lines 40 and 78

libs/web-components/src/components/button/Button.spec.ts
```diff
@@ -56,7 +56,7 @@ describe("GoAButtonComponent", () => {
...
```
ArakTaiRoth, line 57: Another thing to add here, also not for this file. This work will require a PR in ui-components-docs as well to remove the `submit` type as an option in the properties table and the sandbox.

syedszeeshan: Here is the docs update PR https://github.com/GovAlta/ui-components-docs/pull/426

libs/angular-components/src/lib/components/button/button.ts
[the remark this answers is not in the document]
syedszeeshan: I wanted to avoid using the setTimeout, but its not a big deal. I have moved the console warning inside the Button.svlete now (removed from the wrappers)

syedszeeshan: wrapper changes are reverted, so marking it resolved.

libs/web-components/src/components/button/Button.spec.ts
syedszeeshan: @ArakTaiRoth 
In DatePicker.svelte, the property `relative` itself is deprecated, hence this message makes more sense.
```
/***
   * @deprecated This property has no effect and will be removed in a future version
   */
  export let relative: string = "";
```

But in Button.svelte
The `type` is not deprecated, only` type="submit"` is deprecated?
export let type: ButtonType = "primary";

**Do you still want me to add this style of message?**
```
/***
   * @deprecated The type submit is deprecated and will be removed in a future version
   */
```

libs/web-components/src/common/utils.ts
```diff
@@ -174,9 +174,12 @@ export function validateRequired(
 export function typeValidator(
   message: string,
   values: string[],
-  required = false,
+  opts: boolean | { required?: boolean; deprecated?: string[] } = {},
```
ArakTaiRoth, line 177: If we're making this change to `typeValidator`, then don't we need to update everywhere that uses the `required` field, not just `Button`?

- [ ] Badge
- [ ] Callout
- [ ] FormItem
- [ ] IconButton
- [ ] MicrositeHeader
- [ ] Notification
- [ ] Skeleton
- [ ] Table
- [ ] Tooltip

syedszeeshan: existing code will continue to work b/c of this line
`const required = typeof opts === "boolean" ? opts : opts?.required;`

You can also check this (resolved) comment above:
https://github.com/GovAlta/ui-components/pull/2988#discussion_r2304619714
image

ArakTaiRoth (approved):

libs/web-components/src/common/utils.ts
```diff
@@ -174,9 +174,12 @@ export function validateRequired(
 export function typeValidator(
```
chrisolsen, line 174: Even though there are no existing tests for this function, can you add some tests for this function?

Issue #3001, opened 2025-09-09
Checkbox for Angular doesn't return the value when checked


ArakTaiRoth:
# Info

Setting up a basic checkbox in Angular, with a value assigned to it. 

```html
<goab-form-item label="Basic" mb="l">
  <goab-checkbox
    name="item"
    text="Item"
    value="value1"
    (onChange)="checkboxOnChange($event)"
  />
</goab-form-item>
```
```javascript
checkboxOnChange(details: GoabCheckboxOnChangeDetail) {
  console.log(details.value);
}
```

The value does not get sent, instead the function above logs the default when no value is present, "checked". In React, the actual Checkbox value is sent.

## Acceptance Criteria:

1. Angular Checkbox should function the same way as it does in React, and should send the actual Checkbox value when checked.

chrisolsen:
@vanessatran-ddi made a good point in that when using a single checkbox a user is never using the value associated with the checkbox, but instead just whether it is checked or not.

PR #3009, opened 2025-09-11
fix(#3001): prevent angular from losing value when checkbox is toggled

chrisolsen:
# Before (the change)

Angular would lose the value when toggling the checkbox

# After (the change)

The value is retained


chrisolsen:
After talking with @vanessatran-ddi it doesn't seem like this is actually a problem on the Angular side as having a value associated with a checkbox is only really used when you have a group of checkboxes, and things currently work in that scenario.

Issue #3057, opened 2025-09-24
Accordion 2.0 Update


twjeffery:
#### Design Tokens 

- All design tokens are updated to match the new 2.0 designs
  - Has new & updated colour values
  - Has new & updated border radius values
  - Has new & updated shadow & elevation values
  - Has new & updated typography values

#### Web Component
- Has references to new design tokens
- Has old CSS values as fallbacks

#### Testing
- When using the new design tokens it looks and behaves like the updated v2 component
- When using the old design tokens it looks and behaves like the current component

### Resources

- [Figma: Updated Accordion design](https://www.figma.com/design/Jpy1Ea5qglwnp1SgGnagY9/%E2%9D%96-Component-library-BETA?node-id=15932-570754&t=BaDG1i5FDhfwdFqU-1)
- [Figma: Current Accordion component](https://www.figma.com/design/3pb2IK8s2QUqWieH79KdN7/%E2%9D%96-Component-library-%7C-DDD?m=auto&node-id=15932-570754&t=r7aLCk3jZ9F7W9Mx-1)
- [GitHub: Spike for v2 web component](https://github.com/GovAlta/ui-components/pull/3041)

PR #3078, opened 2025-09-29
feat(#3057): update accordion styling to match v2 designs

bdfranck:
This PR updates the appearance of the Accordion component based on our Design System Evolution work.

When using the new design tokens, the accordion looks the same as the [experimental v2 Figma component](https://www.figma.com/design/Jpy1Ea5qglwnp1SgGnagY9/%E2%9D%96-Component-library-BETA?node-id=15932-570754&t=BaDG1i5FDhfwdFqU-1).

When using the current design tokens, the accordion looks the same as the [current Figma component](https://www.figma.com/design/3pb2IK8s2QUqWieH79KdN7/%E2%9D%96-Component-library-%7C-DDD?m=auto&node-id=15932-570754&t=r7aLCk3jZ9F7W9Mx-1).

### DIFF
--- a/libs/web-components/src/components/accordion/Accordion.svelte
+++ b/libs/web-components/src/components/accordion/Accordion.svelte
@@ -58,7 +58,7 @@
     validateRequired("GoAAccordion", { heading });
     validateHeadingSize(headingsize);
     ensureSlotExists(_slotEl);
-    
+
     _headingSlotChildren = getHeadingChildren();
     _accordionId = `accordion-${generateRandomId()}`;
   });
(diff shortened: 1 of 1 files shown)


twjeffery (changes requested):
@bdfranck Looks really good, great to see how much we can change with just the design tokens. I noticed a couple things below: 

1. There is a small movement now on hover when the accordion is open.
   - https://jam.dev/c/f2dce09d-6722-45bf-84cc-68494f4ab6c8


2. The heading in the accordion is slightly higher than vertically centered with the new typography. We had dealt with this with one class throughout the system in v1 components, we can likely do something similar for this across all v2 components. 

    image

3. The border between the heading and the content is darker than the rest when the accordion is open
    
    image

PR #3097, opened 2025-10-07
--goa-border-width

bdfranck:
# Before (the change)

# After (the change)


libs/web-components/src/components/badge/Badge.svelte
```diff
@@ -94,10 +121,25 @@
...
+    sunset: "information-circle",
+    dawn: "information-circle",
+    default: "information-circle",
   }[type];
 
   onMount(() => {
-    setTimeout(() => validateType(type), 1);
+    setTimeout(() => {
+      if (version === "2" && !type) {
+        type = "default";
```
bdfranck, line 136: `Type` is required in v1 but optional in v2. I'm not sure if this is the best approach. Maybe we could handle it in the wrappers or rethink how to have a default option.

libs/web-components/src/components/badge/Badge.svelte
```diff
@@ -308,4 +357,141 @@
...
+    box-shadow: inset 0 0 0 var(--goa-badge-border-width)
+      var(--goa-color-extended-sunset-default);
+    color: var(--goa-color-text-default-sunset-text);
+  }
+
+  .v2.badge-subtle.badge-dawn {
+    background-color: var(--goa-color-extended-dawn-light);
+    box-shadow: inset 0 0 0 var(--goa-badge-border-width)
+      var(--goa-color-extended-dawn-default);
+    color: var(--goa-color-text-default-dawn-text);
```
twjeffery: This should reference the extended colour token, not text-default.
```suggestion
    color: var(--goa-color-extended-dawn-text);
```

libs/web-components/src/components/badge/Badge.svelte
twjeffery: should we keep it required in both? I know we have a default badge type, but I don't know if that's expected.

twjeffery (dismissed):
@bdfranck Looks great!

## Version 2 styles all look correct

Image

Image

Image

Image

## While Version 1 styles maintained

Image

Image

Image

Image

twjeffery (dismissed):

twjeffery (dismissed):
The only changes I see are from the rebase to dev, otherwise looks the same to me

twjeffery (approved):

chrisolsen (approved):

Issue #2915, opened 2025-07-23
Callout 2.0 update


twjeffery:
## Component update for design system 2.0: Callout

Update includes new emphasis levels, visual refinements, component tokens, and deprecation of the `size` property and `event` type.  
[Design in Figma](https://www.figma.com/design/Kkd4vrWhbQ9Oqv0WufXbXC/v2.0?node-id=65517-62144)

---

#### Props

- Add new prop: `emphasis: "high" | "medium" | "low"`
  - Emphasis level controls visual styling and layout spacing
- Deprecate (but keep in code): `size: "medium" | "large"`
  - Undocumented going forward
- Keep existing `type` prop:
  - `success`, `important`, `information`, `emergency`
- Deprecate (but keep in code): `type="event"`
  - Remove from documentation but allow continued use
- Keep all other existing props:
  - `heading`, `maxWidth`, `ariaLive`, margin props, `testId`

#### Visual & Layout Changes

- Visual style is determined by `type` and `emphasis`
- If `heading` is omitted, layout adjusts automatically
  - This mostly affects the `low` emphasis variant, where spacing and alignment are adjusted
  - High and medium emphasis variants are largely unaffected by heading presence
- No new variants are introduced — heading presence or absence simply affects layout within the existing `emphasis` levels
- Design examples will demonstrate how to use emphasis levels with or without headings

#### Token Updates

- Use component tokens for all visual styles:
  - Background, icon, heading text, body text
  - Padding, margin, border-radius
- Tokens vary by:
  - `type` and `emphasis`
  - Presence/absence of `heading` if needed

#### Accessibility

- No changes to current behavior
- `ariaLive` remains optional
- No additional ARIA attributes or roles added

#### Component Structure

- No major markup changes expected
- Layout adjustments for heading presence may require minor structural refinements (to be confirmed in development)
- All styling handled via CSS classes and tokens where possible

PR #3100, opened 2025-10-09
feat(#2915): callout v2 layout and emphasis levels

bdfranck:
This PR updates the callout to matches the new v2 Callout component. It makes the following changes:
- Updates the callout layout
- Uses the [v2 design tokens](https://github.com/GovAlta/design-tokens/pull/96)
- Adds emphasis levels

### DIFF
--- a/libs/web-components/src/components/callout/Callout.svelte
+++ b/libs/web-components/src/components/callout/Callout.svelte
@@ -19,16 +19,23 @@
     "medium",
     "large",
   ]);
+  const [CalloutEmphasis, validateCalloutEmphasis] = typeValidator(
+    "Callout emphasis",
+    ["high", "medium", "low"],
+  );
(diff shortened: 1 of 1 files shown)


twjeffery (changes requested):
Looks great @bdfranck , just a couple adjustments suggested below.

---
Also, it looks like the heading font-weight is showing up as bold, not semi-bold. Now that I look at it, I think it's the same for the Modal heading where it uses semi-bold. It is showing font-weight 600, but visually looks off. Maybe we can connect to figure this one out. 

Here is a screenshot of the coded component (left) beside the figma component (right):

image

twjeffery (approved):

twjeffery (approved):

twjeffery (approved):
The changes look the same as before you rebased

Issue #3072, opened 2025-09-25
Value binding required for Reactive Forms


ArakTaiRoth:
# Info

Reported by Walter and Song Jiang

This was discovered while trying to reset values in a form. This can all be duplicated with the following code in Angular (this is an Angular only problem):
```html
<form [formGroup]="form">
  <goab-form-item label="Last Name" mb="xl">
    <goab-input formControlName="lastName" [trailingContent]="inputTrailing">
      <ng-template #inputTrailing> $ </ng-template>
    </goab-input>
  </goab-form-item>
  <goab-form-item label="First Name" mb="xl">
    <goab-input formControlName="firstName" [value]="this.form.get('firstName')?.value" />
  </goab-form-item>
  <goab-form-item label="Member ID" mb="xl">
    <goab-input-number
      formControlName="memberId"
      [value]="this.form.get('memberId')?.value"
    />
  </goab-form-item>
  <goab-form-item label="Gender" mb="xl">
    <goab-dropdown formControlName="genderId">
      <goab-dropdown-item label="Male" value="1" />
      <goab-dropdown-item label="Female" value="2" />
      <goab-dropdown-item label="Other" value="3" />
    </goab-dropdown>
  </goab-form-item>
  <goab-form-item label="Birth Date" mb="xl">
    <goab-date-picker formControlName="birthDate" />
  </goab-form-item>
</form>
<goab-button (onClick)="resetForm()" mt="2xl">Reset Form</goab-button>
<br />
{{ form.valueChanges | async | json }}
```

```javascript
form!: FormGroup;

constructor(private fb: FormBuilder) {
  this.form = this.fb.group({
    lastName: [""],
    firstName: [""],
    memberId: [""],
    genderId: [""],
    birthDate: [""],
  });
}

resetForm() {
  this.form.reset();
}
```
The issue is, the form reset works fine the first time, but everytime after that, it no longer works on most input types (it works fine for DatePicker). That being said, it does actually reset the value property of the form control, but it doesn't reset the UI, so it looks like the value is still entered (it's not).

This can be alleviated if you bind the `[value]` property to the value of the associated formControl. But accordingly, binding the value to the input shouldn't be required, as linking the input to a formControl should be enough.

## More Info

This is actually really easy to fix, the reason it works properly for the DatePicker is this Angular code:
```javascript
override writeValue(value: Date | null): void {
  this.value = value;

  const datePickerEl = this.elementRef?.nativeElement?.querySelector("goa-date-picker");

[...]

PR #3107, opened 2025-10-15
fix(#3072): angular reset and value binding issue

syedszeeshan:
# Before (the change)

# After (the change)

### DIFF
--- a/libs/angular-components/src/lib/components/base.component.ts
+++ b/libs/angular-components/src/lib/components/base.component.ts
@@ -1,6 +1,13 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import { Spacing } from "@abgov/ui-components-common";
-import { booleanAttribute, Component, Input } from "@angular/core";
+import {
+  booleanAttribute,
+  Component,
+  Input,
+  ElementRef,
+  ViewChild,
+  Renderer2,
+} from "@angular/core";
 import { ControlValueAccessor } from "@angular/forms";
 
 @Component({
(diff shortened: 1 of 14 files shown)


libs/angular-components/src/lib/components/input/input.ts
```diff
@@ -139,13 +140,31 @@ export class GoabInput extends GoabControlValueAccessor implements OnInit {
...
+  }
+
   ngOnInit() {
     this.handleTrailingIconClick = this.onTrailingIconClick.observed;
     if (typeof this.value === "number") {
       console.warn("For numeric values use goab-input-number.");
     }
   }
 
+  override writeValue(value: unknown): void {
```
vanessatran-ddi: Input/Textarea/RadioGroup `writeValue` are quite identical, the Dropdown has the function slightly different: check `!value` , and Checkbox is a unique case, I am thinking if we have a way to elimiate the duplication?

libs/angular-components/src/lib/components/base.component.ts
```diff
@@ -87,12 +87,71 @@ export abstract class GoabControlValueAccessor
     }
   }
 
+  /**
```
vanessatran-ddi: We introduced 5 protected methods, so we end up having child classes to implement 3 or 4 methods to make `writeValue` work. 

Here I will suggest as the below: 
* add goaComponentRef that pointed to the `goa-` component in HTML
* constructor
image

And modify `writeValue` as below:
image
The rest 4 functions we can remove.

Then for the child component you can refer to my comment under `input.ts`

libs/angular-components/src/lib/components/input/input.ts
```diff
@@ -142,7 +143,10 @@ export class GoabInput extends GoabControlValueAccessor implements OnInit {
   isReady = false;
   handleTrailingIconClick = false;
 
-  constructor(private cdr: ChangeDetectorRef) {
+  constructor(
```
vanessatran-ddi: Here, you just assign `#goaComponentRef` vs the `goa` in HTML

image

and modify the constructor:
image

The rest you can remove:
image


Doing so, by handling the common case, 90% of `input` components will just set `value` attribute in the base class, we allow a special case like `checkbox` which is using `checked`, so we can override the writeValue.

vanessatran-ddi (dismissed):

ArakTaiRoth (dismissed):

ArakTaiRoth (approved):

Issue #1813, opened 2024-04-29
Date Picker: min-width and width property


ArakTaiRoth:
# Info

We need to add a `width` property to the Datepicker component, so it can function like our other text based input components. In order to support this, we also need to add a static minimum width to the component, so that it can always be at least the width required to show the data.

## Acceptance Criteria:

1. Add a `width` property to the Datepicker

PR #3127, opened 2025-10-20
fix(#1813): date picker width prop

syedszeeshan:
# Before (the change)

# After (the change)

### DIFF
--- a/libs/react-components/specs/datepicker.browser.spec.tsx
+++ b/libs/react-components/specs/datepicker.browser.spec.tsx
@@ -144,6 +144,137 @@ describe("DatePicker", () => {
       expect(inputEl.disabled).toBe(true);
     })
   });
+
+  describe("Width property", () => {
+    it("applies custom width with px units", async () => {
+      const Component = () => {
+        return <GoabDatePicker testId="date-picker" width="400px" />;
+      };
+
+      const result = render(<Component />);
+      const input = result.getByTestId("calendar-input");
+
+      await vi.waitFor(() => {
+        // Check the input element's computed style
+        const computedStyle = window.getComputedStyle(input.element());
+        const inputWidth = parseFloat(computedStyle.width);
+
+        // The width should be close to 400px (the underlying goa-input component handles the width)
+        expect(inputWidth).toBeGreaterThan(300);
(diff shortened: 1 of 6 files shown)


libs/web-components/src/components/date-picker/DatePicker.svelte
```diff
@@ -301,41 +302,80 @@
...
+        {min}
+        {max}
+        bordered="false"
+        on:_change={onCalendarChange}
+      />
+    </goa-popover>
+  {/if}
 {:else if type === "input"}
-  <goa-form-item error={_error && error} bind:this={_rootEl}>
+  <goa-form-item error={_error && error} bind:this={_rootEl} style={width ? `width: ${width};` : ""}>
```
vanessatran-ddi: `width` has no effect on type="input" date picker.
If you see the Month dropdown, no width specified so it uses default value from `input`, then day input hardcoded its width with 4ch, year input with 6ch. 
image

libs/react-components/specs/datepicker.browser.spec.tsx
```diff
@@ -144,6 +144,128 @@ describe("DatePicker", () => {
...
+    it("applies width to input type datepicker container", async () => {
+      const Component = () => {
+        return <GoabDatePicker type="input" width="500px" />;
+      };
+
+      const result = render(<Component />);
+
+      await vi.waitFor(() => {
+        // For input type, width is applied to the form-item container
+        const monthInput = result.getByTestId("input-month");
```
vanessatran-ddi: this test asserts nothing related about width property

libs/web-components/src/components/date-picker/DatePicker.svelte
ArakTaiRoth: @chrisolsen I don't think we want the `width` property to have an effect on `type="input"` do we? If we do, do we set all components of that type to that width? Or how would that work?

vanessatran-ddi (dismissed):

ArakTaiRoth:
@twjeffery Only issue with this, is that if they put any width below the default "16ch", it will work and thus become unreadable. Do we want to enforce a `min-width`?

ArakTaiRoth (approved):

Issue #2664, opened 2025-04-24
Calendar / Datepicker date value


chrisolsen:
Currently the calendar and datepicker return a `Date` typed value within the _change event. This is resulting in timezone conversions having to be handled to prevent incorrect dates to be displayed to the user at a later time. Since the calendar does not include any time (hour, minute) selection there is no advantage to using the `Date` type.

We need to remove the the `Date` type being used for the `value` and related props replacing it with a custom Date type that only contains the day, month and year.

To prevent breaking changes the `_change` event will need to return the current `value` value as a `Date` type, but the `_change` event will also include a `valueStr` string property.

All the logic within the calendar component dealing with dates can be removed and the `value` with the `_change` event can be calculated on the fly from the `valueStr` value.

---

Use the `getLocalDAteValues()` utils function within the `onMount` to immediately parse the value prop. No `Date` types should be used from this point on.

---

Before taking this on chat with @chrisolsen first.

## Acceptance Criteria

1. Function returns data as is
2. Function also needs to return date in "y-m-d" string

chrisolsen:
Use the `getLocalDAteValues()` utils function within the `onMount` to immediately parse the `value` prop. No `Date` types should be used from this point on.

PR #3148, opened 2025-10-29
fix(#2664): move away from Date value within the calendar/datepicker

chrisolsen:
Since time is never wanted when selecting dates it only adds to complexity and adds issues regarding timezones. These changes replace the Date with a custom type and minimizes the use of the Date type to prevent breaking the current component binding.

# Before (the change)

# After (the change)

### DIFF
--- a/libs/web-components/src/components/date-picker/DatePicker.svelte
+++ b/libs/web-components/src/components/date-picker/DatePicker.svelte
@@ -9,9 +9,8 @@
 
 <script lang="ts">
   import { onMount, tick } from "svelte";
-  import { addDays, addMonths, addYears, format, startOfDay } from "date-fns";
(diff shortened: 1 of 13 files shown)


libs/web-components/src/components/date-picker/DatePicker.svelte
```diff
@@ -217,69 +177,47 @@
 
     switch (e.key) {
       case "ArrowLeft":
-        _date ||= addDays(new Date(), 1);
-        _date = addDays(_date, -1);
+        _date ||= new CalendarDate().nextDay;
```
willcodeforcoffee: I like that you added this next/previous API. The `date ||=` bit here makes it feel complicated, but these methods really clarify what each `case` is doing.

Issue #3123, opened 2025-10-18
`icon` property doesn't default to false


ArakTaiRoth:
Issue reported by: andrzej.zadorozny@gov.ab.ca
Web component version: 1.38.0
Angular/React version: N/A

## Expected Behaviour

As per documentation, the default value of the icon attribute for goa-badge should be false.

## Actual Behaviour

The goa-badge icon is enabled by default if no icon attribute is specified.
Observed in web-components v1.38.0.

## Replication Steps

Add a goa-badge, and do not set the icon attribute.

## More Info

This is an Angular only issue. React functions as expected in this manner.

## Acceptance Criteria

1. Fix Angular Badge to use default value of `false` for the `icon` property
2. Update Badge documentation to include `iconType` property for newer versions.
    * Add property to Sandbox as well

PR #3158, opened 2025-11-04
fix(#3123): set badge icon prop default to false

syedszeeshan:
# Before (the change)

# After (the change)

### DIFF
--- a/libs/angular-components/src/lib/components/badge/badge.ts
+++ b/libs/angular-components/src/lib/components/badge/badge.ts
@@ -17,7 +17,7 @@ import { GoabBaseComponent } from "../base.component";
     <goa-badge
       *ngIf="isReady"
       [attr.type]="type"
-      [attr.icon]="icon"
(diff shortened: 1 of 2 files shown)


libs/angular-components/src/lib/components/badge/badge.ts
```diff
@@ -17,7 +17,7 @@ import { GoabBaseComponent } from "../base.component";
     <goa-badge
       *ngIf="isReady"
       [attr.type]="type"
-      [attr.icon]="icon"
+      [attr.icon]="icon ? 'true' : null"
```
ArakTaiRoth: @chrisolsen Just curious on your thoughts. If we actually made `icon` default to "false" in `Badge.svelte`, it would remove the need to do this, and allow us to remove code from `badge.tsx` as well. It feels like that would be more inline with what we've done everywhere else.

ArakTaiRoth (approved):

ArakTaiRoth (approved):

Issue #2977, opened 2025-08-19
OnChangeDetails common components don't have the event


ArakTaiRoth:
# Info

Reported by Nick Yang, Jim Wen, and Inoka Silva

For any of our `OnChangeDetails` components located in `common.ts`, none of them supply the actual event, they all only supply `name` and `value`. This makes it impossible to do anything with the event itself, such as stopping it from bubbling up.

This is further exacerbated if you put any kind of input or accordion inside a Tabs component, which then fires the Tab onChange event.

## Acceptance Criteria:

1. All of our `OnChangeDetails` components need to have the event added to them as output

PR #3179, opened 2025-11-12
DONT REVIEW - fix(#2977): add event to common components

syedszeeshan:
# Before (the change)

# After (the change)


chrisolsen (changes requested):
This PR contains non-required changes in web-components and passes the wrong event ref in the React/Angular components. The initial change event is not what is needed to be passed into the initial _change event, rather it is the CustomEvent that is received by the React/Angular components that need need to be added to the *OnChangeDetail.

PR #3181, opened 2025-11-13
feat(#1908): linear progress indicator

willcodeforcoffee:
[Design specs](https://www.figma.com/design/3pb2IK8s2QUqWieH79KdN7/%E2%9D%96-Component-library-|-DDD?node-id=622-13604&p=f&m=dev)

# Before (the change)
No linear progress indicator. 🥺

# After (the change)
Linear progress indicator! 💥

image

## Make sure that you've checked the boxes below before you submit the PR

- [x] I have read and followed the [setup steps](https://goa-dio.atlassian.net/wiki/spaces/DS/pages/3086385489/Contribution+-+web+content#%5BinlineExtension%5DCode-contributions)
- [x] I have created necessary unit tests
- [x] I have tested the functionality in both React and Angular.

## Steps needed to test

You just need to add the component to your page.

See `doc,md` or `apps/prs/react/src/app/all.tsx` for some examples.

Issue #1908, opened 2024-06-10
Linear Progress indicator: new component


Spark450:
## Acceptance Criteria:
Create linear progress indicator as outline in the [Figma design](https://www.figma.com/design/3pb2IK8s2QUqWieH79KdN7/%E2%9D%96-Component-library-%7C-DDD?m=auto&node-id=41625-1999968&t=lb3sVM9XKfCcGXmE-1)

PR #3210, opened 2025-11-25
feat(#1908): linear progress bar (take 2)

willcodeforcoffee:
[Figma Designs are here.](https://www.figma.com/design/3pb2IK8s2QUqWieH79KdN7/%E2%9D%96-Component-library-%7C-DDD?node-id=622-13604&p=f&m=dev)

This PR replaces https://github.com/GovAlta/ui-components/pull/3181 which I'm closing to keep things a little simpler.

# Before (the change)

No progress bar.

# After (the change)

Two kinds of linear progress bars.

image

image

## Steps needed to test

PR files have been added for Angular and React under feat1908.

animated gif because Eric

### DIFF
--- a/libs/web-components/src/components/linear-progress/LinearProgress.spec.ts
+++ b/libs/web-components/src/components/linear-progress/LinearProgress.spec.ts
@@ -1,0 +1,148 @@
+import { render, waitFor } from "@testing-library/svelte";
+import GoALinearProgress from "./LinearProgress.svelte";
+import { it, describe } from "vitest";
+
+describe("GoALinearProgress", () => {
+  const testid = "progressbar";
+
+  [-1, 0, 20, 100, 101].forEach((progress: number) => {
+    it(`renders the linear progress with progress value ${progress}%`, async () => {
+      const { queryByTestId } = render(GoALinearProgress, {
+        testid,
+        progress,
+      });
+
+      await waitFor(() => {
+        const expectedPercentage = Math.max(0, Math.min(progress || 0, 100));
+        expect(queryByTestId(`${testid}-percentage`)?.innerHTML).toContain(
+          `${expectedPercentage}%`,
+        );
+        expect(
+          queryByTestId(`${testid}-progressbar-container`)?.getAttribute(
+            "aria-valuenow",
+          ),
+        ).toBe(`${expectedPercentage}`);
+      });
+    });
+  });
+
+  it("renders the ARIA attributes correctly", async () => {
+    const { queryByTestId } = render(GoALinearProgress, {
+      testid,
+      progress: 50,
+      ariaLabel: "Loading progress",
+      ariaLabelledby: "progress-label",
+    });
+
+    await waitFor(() => {
+      const progressBar = queryByTestId(
+        `${testid}-progressbar-container`,
+      ) as HTMLElement;
+      expect(progressBar.getAttribute("role")).toBe("progressbar");
+      expect(progressBar.getAttribute("aria-label")).toBe("Loading progress");
+      expect(progressBar.getAttribute("aria-labelledby")).toBe(
+        "progress-label",
+      );
+      expect(progressBar.getAttribute("aria-valuemin")).toBe("0");
+      expect(progressBar.getAttribute("aria-valuemax")).toBe("100");
+      expect(progressBar.getAttribute("aria-valuenow")).toBe("50");
+      expect(progressBar.getAttribute("aria-valuetext")).toBeNull();
+    });
+  });
+
+  it("rounds the percentage down when progress is a decimal", async () => {
+    const { queryByTestId } = render(GoALinearProgress, {
+      testid,
+      progress: 50.6,
+    });
+
+    await waitFor(() => {
+      expect(queryByTestId(`${testid}-percentage`)?.innerHTML).toContain("51%");
+    });
+  });
+
+  it("rounds the percentage up when progress is a decimal", async () => {
+    const { queryByTestId } = render(GoALinearProgress, {
+      testid,
+      progress: 50.4,
+    });
+
+    await waitFor(() => {
+      expect(queryByTestId(`${testid}-percentage`)?.innerHTML).toContain("50%");
+    });
+  });
+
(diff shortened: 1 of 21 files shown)


libs/web-components/src/components/linear-progress/LinearProgress.svelte
```diff
@@ -0,0 +1,109 @@
...
+  <div
+    class="progressbar-container"
+    role="progressbar"
+    aria-label={arialabel}
+    aria-labelledby={arialabelledby}
+    aria-valuenow={progress === undefined || progress === null
+      ? undefined
+      : Math.round(Math.max(0, Math.min(progress, 100)))}
+    aria-valuemin="0"
+    aria-valuemax="100"
```
vanessatran-ddi: We need to add 
`aria-live="polite"` so the screen reader will announce.
And consider adding aria-valuetext for better screen reader support, reason is: when we add `aria-live="polite"` for the screen reader to announce it, then we need to have valuetext because the screen reader only be notified if the text is changed. 

```
aria-valuetext={progress === undefined || progress === null
    ? undefined
    : `${Math.round(Math.max(0, Math.min(progress, 100)))} percent`}
```

I find this reference is covering everything https://www.patternfly.org/components/progress/accessibility/

libs/web-components/src/components/linear-progress/LinearProgress.svelte
```diff
@@ -0,0 +1,121 @@
+<svelte:options
+  customElement={{
+    tag: "goa-linear-progress",
+    props: {
+      testid: { type: "String", attribute: "testid", reflect: true },
+      progress: { type: "Number", attribute: "progress" },
+      // showpercentage is left as a string because we want it to be truthy by default (even for undefined or null), but
+      // React and Angular will ignore attributes that are left falsey.
+      showpercentage: { type: "String", attribute: "show-percentage" },
```
chrisolsen: One of the benefits to creating the props is allow our internal props to use the camelCase convention, which isn't allowed on html elements. So `showpercentage` can be changed to `showPercentage`

libs/web-components/src/components/linear-progress/LinearProgress.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+describe("GoALinearProgress", () => {
+  [-1, 0, 20, 100, 101].forEach((progress: number) => {
+    it(`renders the linear progress with progress value ${progress}%`, async () => {
+      const { container } = render(GoALinearProgress, {
+        progress,
+      });
+
+      await waitFor(() => {
+        const expectedPercentage = Math.max(0, Math.min(progress || 0, 100));
+        expect(container.querySelector(".percentage")?.innerHTML).toContain(
```
chrisolsen: Try to use `data-testid` attributes for queries within tests, to reduce breaking tests if someone decides to rename classes. It also allow for easier queries within tests as `getByTestId` is one of the functions that makes browser tests easier.

libs/web-components/src/components/linear-progress/LinearProgress.svelte
```diff
@@ -0,0 +1,121 @@
...
+    aria-valuemax="100"
+    aria-live="polite"
+    aria-valuenow={progress === undefined || progress === null
+      ? undefined
+      : Math.round(Math.max(0, Math.min(progress, 100)))}
+    aria-valuetext={progress === undefined || progress === null
+      ? undefined
+      : `${Math.round(Math.max(0, Math.min(progress, 100)))} percent`}
+  >
+    {#if progress === undefined || progress === null}
```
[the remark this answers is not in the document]
willcodeforcoffee: I thought I had tried the `$:` and it didn't work, but I read up and tried it again.
So much better.

libs/web-components/src/components/linear-progress/LinearProgress.spec.ts
willcodeforcoffee: I've been thinking about this one. To be thorough I'm trying check values inside different elements in the controls. I think using the `testid` is going to actually complicate the tests more that simplify.

willcodeforcoffee: @chrisolsen I checked how some other components use `testid`, like the Accordion, and copied how they do it. LMK if you approve?

vanessatran-ddi (dismissed):

libs/web-components/src/components/linear-progress/LinearProgress.svelte
```diff
@@ -0,0 +1,128 @@
+<svelte:options
+  customElement={{
+    tag: "goa-linear-progress",
+    props: {
+      testid: { type: "String", attribute: "testid", reflect: true },
+      progress: { type: "Number", attribute: "progress" },
+      // showPercentage is left as a string because we want it to be truthy by default (even for undefined or null), but
+      // React and Angular will ignore attributes that are left falsey.
+      showPercentage: { type: "String", attribute: "show-percentage" },
```
chrisolsen: In this case booleans don't quite work. So there are a few options
1. Use strings as have done here. The downfall is that it is a little hacky, hence the need for a comment to explain things. I would also like to get away from using the toBoolean function
2. Invert the prop i.e. `hidePercentage` which allows the default falsey value to allow things to work as designed. 
3. Use a union type ex `percentProgress`: "visible" | "hidden;" which defaults to visible <= best option in my opinion

twjeffery (approved):
@willcodeforcoffee All looks good to me now!

Image

libs/web-components/src/components/linear-progress/LinearProgress.svelte
```diff
@@ -0,0 +1,128 @@
...
+/>
+
+<script lang="ts">
+  function isLabelVisible(visibility: string | undefined): boolean {
+    return visibility === undefined || visibility.toLowerCase() === "visible";
+  }
+
+  export let testid: string | undefined = undefined;
+  export let progress: number | undefined = undefined;
+  export let percentVisibility: "visible" | "hidden" | undefined = undefined;
```
chrisolsen: Could you just default this to `visible`? This would allow for the removal of the `undefined`

PR #3223, opened 2025-11-27
fix(#2977): Added event properties to all input components

ArakTaiRoth:
# Before (the change)

All Input components (Input, Radio, Checkbox, Checkbox List, Text Area, File Upload, Dropdown, Date Picker) send the value and name of component for all of their events.

# After (the change)

Now they send the actual event itself as well. So you can use that to stop propagation if needed.

## Steps needed to test

Use `npm run serve:prs:angular` and `npm run serve:prs:react`, and use Bug 2977. Use the console to see what's sent for all the events for all components. And at the bottom is an area to test stopping propagation, the first tab it should stop and thus the Tab onChange won't run, the second and third tab will show you what the problem is this was meant to fix.

### DIFF
--- a/libs/angular-components/src/lib/components/checkbox-list/checkbox-list.spec.ts
+++ b/libs/angular-components/src/lib/components/checkbox-list/checkbox-list.spec.ts
@@ -124,16 +124,19 @@ describe("GoabCheckboxList", () => {
       By.css("goa-checkbox-list"),
     ).nativeElement;
 
+    const changeEvent = new Event("change");
     const detail: GoabCheckboxListOnChangeDetail = {
       name: "fruits",
       value: ["apple", "banana"],
+      event: changeEvent,
     };
 
-    const changeEvent = new CustomEvent("_change", {
-      detail,
-    }) as Event;
-
-    fireEvent(el, changeEvent);
+    fireEvent(
+      el,
+      new CustomEvent("_change", {
+        detail,
+      }),
+    );
 
     expect(fcChangeSpy).toHaveBeenCalledWith(["apple", "banana"]);
     expect(fcTouchedSpy).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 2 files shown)


libs/react-components/specs/input.browser.spec.tsx
```diff
@@ -0,0 +1,81 @@
...
+        />
+        ,
+        <GoabInput
+          name="blur-target"
+          testId="blur-target"
+          // eslint-disable-next-line @typescript-eslint/no-empty-function
+          onChange={() => {}}
+        />
+        ,
+      </div>,
```
chrisolsen: The tests should have a wrapping element that is also listening for `_change` events. A change then needs to be made on the inner component, a test could be added to ensure that the _change event is received on the inner element and the other element needs to be tested that it doesn't receive the event.

ArakTaiRoth: @chrisolsen I updated the test, if you can take another look

libs/react-components/src/lib/file-upload-card/file-upload-card.tsx
```diff
@@ -51,8 +51,8 @@ export function GoabFileUploadCard({
     if (!el.current) return;
 
     const current = el.current;
-    const deleteHandler = () => onDelete?.({ filename });
-    const cancelHandler = () => onCancel?.({ filename });
+    const deleteHandler = (event: Event) => onDelete?.({ filename, event });
```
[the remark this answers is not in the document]
ArakTaiRoth: @vanessatran-ddi I believe this is fixed

vanessatran-ddi (approved):

chrisolsen:
Looks good. Just need you to squash the commits down to one.

ArakTaiRoth:
Closing, as this PR had a lot of issues with merges and rebases and conflict changes, you can see the new PR here, https://github.com/GovAlta/ui-components/pull/3316

PR #3225, opened 2025-11-27
feat(#3137): add work side menu group

### LINKED ISSUE #3137: Workspace Side Menu: Menu Group
The work side menu needs a way to group menu items like the `goab-side-menu-group` component. You can see an example of a menu group in the [Design System website v2 exploration](https://www.figma.com/design/uHgovIyK7mANSwgr84TVjB/DS-2.0-Website-Design?node-id=162-88237&t=rICV7KBMvaznE88m-1).

## Acceptance Criteria:
- Has a heading with...
  - Heading text
  - Icon (optional)
- Can accept one or more menu items
- Can select the heading to toggle the menu items

## Questions
- Do we include a property for controlling if the group is open?
    - No

bdfranck:
This PR adds a Work Side Menu Group to use with the Workspace Side Menu. Features include...

- Toggling the group open and closed
- Using design tokens to style the Work Side Menu Items

work-side-menu-group

### Sample code
```svelte
<goa-work-side-menu
  url="/"
  heading="Income and Employment Support (IES)"
  user-name="Edna Mode"
  user-secondary-text="edna.mode@example.com"
  open="true"
>
  <div slot="primary">
    <goa-work-side-menu-item
      icon="search"
      label="Search"
      url="/search"
    />

    <goa-work-side-menu-item
      icon="list"
      label="Clients"
      url="/clients"
    />

    <goa-work-side-menu-item
      icon="calendar"
      label="Schedule"
      url="/schedule"
    />

    <goa-work-side-menu-group icon="trending-up" heading="Reports">
      <goa-work-side-menu-item
        icon="file-text"
        label="Monthly Report"
        url="/reports/monthly"
      />
      <goa-work-side-menu-item
        icon="bar-chart-2"
        label="Annual Report"
        url="/reports/annual"
      />
    </goa-work-side-menu-group>
  </div>

  <div slot="secondary">
    <goa-work-side-menu-item
      icon="notifications"
      label="Notifications"
      type="success"
      badge="1"
      url="/notifications"
    />

    <goa-work-side-menu-item
      icon="help-circle"
      label="Support"
      url="/support"
    />

    <goa-work-side-menu-item icon="settings" label="Settings" url="/settings" />
  </div>

  <div slot="account">
    <goa-work-side-menu-item
      icon="person"
      label="Account management"
      url="/account"
    />

    <goa-work-side-menu-item icon="log-out" label="Log out" url="/logout" />
  </div>
</goa-work-side-menu>
```

### Known issues

This component currently doesn't work with my brittle keyboard navigation in the Work Side Menu. I've created a bug to address it separately: https://github.com/GovAlta/ui-components/issues/3279

### DIFF
--- a/apps/prs/angular/src/app/app.component.html
+++ b/apps/prs/angular/src/app/app.component.html
@@ -75,6 +75,8 @@
           <a href="/features/1908">1908</a>
           <a href="/features/2609">2609</a>
           <a href="/features/v2-icons">v2 header icons</a>
+          <a href="/features/3137">3137</a>
+          <a href="/features/1908">1908</a>
         </goab-side-menu-group>
       </goab-side-menu>
     </section>
--- a/apps/prs/angular/src/app/app.routes.ts
+++ b/apps/prs/angular/src/app/app.routes.ts
@@ -58,6 +58,7 @@ import { Feat3102Component } from "../routes/features/feat3102/feat3102.componen
 import { Feat1908Component } from "../routes/features/feat1908/feat1908.component";
 import { Feat2609Component } from "../routes/features/feat2609/feat2609.component";
 import { FeatV2IconsComponent } from "../routes/features/featV2Icons/feat-v2-icons.component";
+import { Feat3137Component } from "../routes/features/feat3137/feat3137.component";
 
 export const appRoutes: Route[] = [
   { path: "everything", component: EverythingComponent },
@@ -119,4 +120,6 @@ export const appRoutes: Route[] = [
   { path: "features/3102", component: Feat3102Component },
   { path: "features/2609", component: Feat2609Component },
   { path: "features/v2-icons", component: FeatV2IconsComponent },
+  { path: "features/3137", component: Feat3137Component },
+  { path: "features/1908", component: Feat1908Component },
 ];
--- a/apps/prs/angular/src/routes/features/feat3137/feat3137.component.html
+++ b/apps/prs/angular/src/routes/features/feat3137/feat3137.component.html
@@ -1,0 +1,15 @@
+<h1>Feature 3137: Work Side Menu Group</h1>
+
+<p>
+  This demonstrates a working Work Side Menu Group component with some child items. You should be able to toggle the menu.
+</p>
+
+<goabx-work-side-menu heading="Menu heading" url="#" [primaryContent]="primaryTemplate" [open]="true">
+</goabx-work-side-menu>
+<ng-template #primaryTemplate>
+  <goabx-work-side-menu-group heading="Side menu group" icon="star">
+    <goabx-work-side-menu-item label="Menu item 1" icon="star" url="/item-1" />
+    <goabx-work-side-menu-item label="Menu item 2" icon="star" url="/item-2" />
+    <goabx-work-side-menu-item label="Menu item 3" icon="star" url="/item-3" />
+  </goabx-work-side-menu-group>
+</ng-template>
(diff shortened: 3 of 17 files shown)


bdfranck:
@chrisolsen I've resolved the conflicts... for now. 🤪

twjeffery (commented):
Looks good @bdfranck, just a few minor things I see:

1. the text label alignment issue that Chris mentioned still seems to be there for me (text is slightly down of vertically centering with icons. I like the flex-start, so multi line will align properly a the start, but the line height seems to be shifting the label down too much.

Image

2. maybe related to the align-start from #1, the icons all shift slightly up when the menu expands/collapses. removing       270 -      align-items: center;    seems to fix it.                                       

https://github.com/user-attachments/assets/f9345de9-72f8-4e5d-9fca-e7ff5215c600

**I think this fixes both #1 and #2 above:**                                         
                                                                               
  A. In WorkSideMenuItem.svelte, remove align-items: center from the collapsed 
  container query:                                                             
 ```
 @container (max-width: 160px) {                                              
    .menu-item {                                                               
      height: 36px;                                                            
      margin: 0;                                                               
      /* remove: align-items: center; */                                       
      padding-left: calc(var(--goa-space-xs) + 2px);                           
    }                                                                          
  }   
```                                                                         
                                                                               
  B. In WorkSideMenuItem.svelte, add margin-top to the icon:                   
```
  goa-icon {                                                                   
    display: var(--goa-work-side-menu-item-icon-display, flex);                
    margin-top: 2px;                                                           
  }   
```                                                                         
                                                                               
  C. In WorkSideMenuGroup.svelte, add margin-top to the trailing chevron:      
```
  .trailing-icon {                                                             
    display: flex;                                                             
    margin-top: 2px;                                                           
  }     
```                                                                       
                                                                               
  This keeps icon positioning consistent between states (no shift) and aligns  
  the icons with the text's first baseline. 

Image

---

3. The line underneath the leading icon that shows the group is slight right of horizontal center from that icon. 

Image

4. in Figma we have the parent menu item showing an active state when expanded. I don't know if we need that state to persist, but having the background helps the child items feel closer and related to the parent above them. If we reduce the gap between the parent and the group below it, that could achieve the same thing? I'm curious what you think would be a better design.

Image

5. One last minor thing I noticed: in Figma we have the sub items closer to the vertical line (12px), slightly inset from where the parent label starts. Would we want to change this to be closer? Up to you for what the design should be. It feels a bit spaced out when I see it in something like the account menu (see below).

Image

^Figma^

Image

^Code^

Let me know if you have any questions and/or want to discuss anything.

twjeffery (commented):
Looks great to me now @bdfranck! Thanks for making those changes. 

One more thing comes up now that might be a bit confusing: 
- when a parent menu item is expanded and the side menu is collapsed, the menu item looks active because of that background. Also, when a user clicks on the parent group when the menu is collapsed, it opens and closes, but the user can't see it because the main side menu is collapsed. See video below: 

https://github.com/user-attachments/assets/ebe6212f-e42b-472a-bce9-71ada0dae60f

In my latest version of the design system website, I made it so if a side menu group is clicked when the side menu is collapsed, the menu opens automatically. thoughts on adding that approach as applicable here? https://goa-website-generation-v2.netlify.app/

twjeffery (approved):
> @twjeffery Good catch! I've updated my commit to remove the background when the menu is collapsed. 👍
> 
> Instead of expanding the scope of this ticket, I could create a new one for improving the menu collapsed state. Here's all the ideas that I've captured so far:
> 
> * Expand the menu when selecting a group
> * Expand the menu on load if a sub-menu item is the current item
> * Remember the expanded state between pages (and also between sessions?)
> * Collapse the menu when reducing the viewport to tablet width

Good idea Benji, approved this version

chrisolsen (approved):

PR #3226, opened 2025-11-28
feat(#3195): update date picker to v2

bdfranck:
This PR updates the Date Picker and related Calendar components to match the latest Design Evolution changes.

It includes the following changes:
- Uses the correct version for child components
- Uses compact inputs for v2
- Adds an offset for the focus ring on a selected day
image

PR #3230, opened 2025-12-01
fix(#3231): restore focus state for v2 read-only Input

bdfranck:
Our v2 changes to the read-only state of an Input in https://github.com/GovAlta/ui-components/pull/3082 assumed that a user doesn't interact with a read-only input. But we discovered in https://github.com/GovAlta/ui-components/pull/3226 that we use an interactive read-only Input in our Date Picker. 

[MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/readonly#attribute_interactions) on `readonly` writes the following:

> The difference between [disabled](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/disabled) and readonly is that read-only controls can still function and are still focusable, whereas disabled controls can not receive focus and are not submitted with the form and generally do not function as controls until they are enabled.

This PR restores the focus state to a read-only v2 Input.

### Before

Date Picker doesn't have a hover or focus state

readonly-no-focus

### After

Date Picker has a hover  and focus state

read-only-input-focused

### DIFF
--- a/libs/web-components/src/components/input/Input.svelte
+++ b/libs/web-components/src/components/input/Input.svelte
@@ -500,12 +500,12 @@
     --goa-text-input-space-btw-icon-text: var(--goa-text-input-space-btw-icon-text-compact);
   }
 
-  .goa-input:not(.error):not(.input--disabled):not(:has(input:read-only)):hover:not(:has(input:focus-visible)) {
+  .goa-input:not(.error):not(.input--disabled):hover:not(:has(input:focus-visible)) {
     /* hover border */
     box-shadow: var(--goa-text-input-border-hover);
   }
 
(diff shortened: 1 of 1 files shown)


libs/web-components/src/components/input/Input.svelte
```diff
@@ -572,7 +572,7 @@
   }
 
   input:read-only {
-    cursor: default;
+    cursor: var(--goa-text-input-cursor-readonly, default);
```
bdfranck: I've kept the default cursor but added a token name to allow the Date Picker to override it.

chrisolsen (approved):

PR #3233, opened 2025-12-01
fix(#3232): GoabText tag size fix

### LINKED ISSUE #3232: GoabText: tag prop should automatically apply corresponding heading size
When using GoabText with a heading tag (h1-h5) but without explicitly setting the size prop, the component renders the semantic HTML element but doesn't apply the design system typography styles. Instead, it falls back to browser defaults, which don't match our design tokens.

### Expected behavior
`<GoabText tag="h1">Title</GoabText>` should render with the same typography as `<GoabText tag="h1" size="heading-xl">Title</GoabText>`

### Current Behavior
When text includes a tag prop without the size prop, the heading gets incorrect sizing.
https://jam.dev/c/30cffcdd-0095-45c2-8c9c-e35f42753a85

### Cause
In Text.svelte, the size prop is applied as a CSS class (line 103), but when size is undefined, no class is applied. There's no mapping from tag to a default size.

### Suggested fix
Add automatic size derivation when tag is a heading and size isn't explicitly set:
```
  function getDefaultSizeForTag(tag: string): Size | undefined {
    switch(tag) {
      case 'h1': return 'heading-xl';
      case 'h2': return 'heading-l';
      case 'h3': return 'heading-m';
      case 'h4': return 'heading-s';
      case 'h5': return 'heading-xs';
      default: return undefined;
    }
  }
```

twjeffery:
When using `GoabText` with a heading tag (h1-h5) without explicitly setting the `size` prop, the component now correctly applies the corresponding design system typography instead of incorrect sizes.

# Before (the change)
When text includes a tag prop without the size prop, the heading gets incorrect sizing.
Video of mismatched text size: [jam.dev/c/30cffcdd-0095-45c2-8c9c-e35f42753a85](https://jam.dev/c/30cffcdd-0095-45c2-8c9c-e35f42753a85)

# After (the change)

**Mapping:**
   - `tag="h1"` → `heading-xl`
   - `tag="h2"` → `heading-l`
   - `tag="h3"` → `heading-m`
   - `tag="h4"` → `heading-s`
   - `tag="h5"` → `heading-xs`

image

## Steps needed to test
- [ ] Navigate to playground `/bugs/3232` to see comparison examples
- [ ] Verify `<GoabText tag="h1">` matches `<GoabText tag="h1" size="heading-xl">`
- [ ] Verify non-heading tags (p, span, div) remain unstyled without explicit size
- [ ] Verify explicit size prop still overrides tag-base default"

### DIFF
--- a/libs/web-components/src/components/text/Text.svelte
+++ b/libs/web-components/src/components/text/Text.svelte
@@ -1,11 +1,4 @@
-<svelte:options
-  customElement={{
-    tag: "goa-text",
-    props: {
-      id: { attribute: "id", type: "String", reflect: true },
-    },
-  }}
-/>
+<svelte:options customElement="goa-text" />
 
 <script lang="ts" context="module">
   export type HeadingElement = "h1" | "h2" | "h3" | "h4" | "h5";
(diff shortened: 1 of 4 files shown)


libs/web-components/src/components/text/Text.svelte
```diff
@@ -1,11 +1,4 @@
-<svelte:options
-  customElement={{
-    tag: "goa-text",
-    props: {
-      id: { attribute: "id", type: "String", reflect: true },
-    },
-  }}
-/>
```
chrisolsen: This can be removed, but the `id` then needs to be added to the main html element ex h1, h2, ...

chrisolsen (approved):

bdfranck (approved):
I tested the changes...

- ✅ They appear visually consistent if either `as` or `size` is set. 

image

Looks good to me! 👍

Issue #3215, opened 2025-11-25
Initial open state exceeds viewport height and hides footer actions


Spark450:
Reported by Thang Lam

## Summary

When the drawer is initially opened, its height can exceed the viewport height. In these cases, the drawer extends past the bottom of the viewport and the footer area (including action buttons) is partially cut off and not fully visible.

<img width="322" height="939" alt="Image" src="https://github.com/user-attachments/assets/0f29e270-cc90-420b-be15-46b0688ba930" />

## More Info

It should be noted as well, this only happens the first time you open the drawer. If you close and then re-open the drawer on your page load, it'll look how it's supposed to. You have to refresh the page to see the issue again.

## Replication Code

```typescript
<goab-button type="primary" (onClick)="openDrawer()">Open Drawer</goab-button>

<ng-template #drawerActions>
  <goab-button-group alignment="end" gap="relaxed">
    <goab-button type="secondary" (onClick)="closeDrawer()">Cancel</goab-button>
    <goab-button type="primary" (onClick)="closeDrawer()">Save</goab-button>
  </goab-button-group>
</ng-template>

<goab-drawer
  [open]="open"
  position="right"
  heading="Project Review Panel"
  [actions]="drawerActions"
  (onClose)="close()"
>
  <p>
    Use the buttons below to close this right-positioned drawer. The content area
    stretches to fill the available vertical space for layout testing.
  </p>
  <h4 style="margin: 0 0 8px 0">Review checklist</h4>
  <ul style="margin: 0; padding-left: 18px; line-height: 1.5">
    <li>Confirm scope alignment</li>
    <li>Validate stakeholder sign-offs</li>
    <li>Verify accessibility and QA coverage</li>
  </ul>
  <p style="margin-top: 0">
    This panel expands to the full viewport height so you can test scrolling and layout
    behavior. Add or remove content here to simulate different drawer payloads.
  </p>
  <p>
    Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed non risus. Suspendisse
    lectus tortor, dignissim sit amet, adipiscing nec, ultricies sed, dolor.
  </p>
  <p>
    Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed non risus. Suspendisse
    lectus tortor, dignissim sit amet, adipiscing nec, ultricies sed, dolor.
  </p>
</goab-drawer>
```

```typescript
open = false;

openDrawer() {
  this.open = true;
}

close() {
  this.open = false;
}
[...]

PR #3261, opened 2025-12-10
fix(#3215): Fixed actions slot position for Drawer

ArakTaiRoth:
# Before (the change)

This is what a Drawer with full content and an action slot would like:
image

# After (the change)

Now, the action slot should appear normally and not be cut off at the bottom

I didn't add any unit tests, as I'm not sure there are any that would catch something like this.

## Steps needed to test

Manual test files are located under Bug 3215


willcodeforcoffee (dismissed):

ArakTaiRoth:
@chrisolsen This is good to be merged

twjeffery (approved):
Looks good to me

Issue #3201, opened 2025-11-20
Modify Input component events


ArakTaiRoth:
# Info

Reported by Gurpreet Nijjar

Currently an HTML input component `<input>` fires `onChange` when something in the input changes, and `onBlur` when the user switches focus away from the component.

Our Input component `<goab-input>` fires `onChange` when something in the input changes, but fires `onBlur` and `onChange` when the user switches focus away from the component.

This is an incorrect model on when to fire events based on the general understanding of how an HTML input functions, and will cause issues if there are competing requirements in the `onChange` function and the `onBlur` function. Potentially what happens in `onChange` can and will overwrite what happened in `onBlur`.

## Acceptance Criteria:

1. `onChange` shouldn't fire when the user switches focus from the Input component
2. Check TextArea as well, and fix if needed there

PR #3284, opened 2025-12-17
fix(#3201): Input component onChange doesn't fire onBlur

ArakTaiRoth:
# Before (the change)

When you tabbed out of an Input or TextArea component, the onChange event would fire.

# After (the change)

When you tab out of an Input or TextArea component, the onChange event will not fire

**NOTE:** I didn't add any additional tests, as I figured the current tests that have been changed are sufficient

## Steps needed to test

PR test Bug 3201

### DIFF
--- a/apps/prs/angular/src/app/app.component.html
+++ b/apps/prs/angular/src/app/app.component.html
@@ -47,6 +47,7 @@
           <a href="/bugs/3072">3072</a>
           <a href="/bugs/3118">3118</a>
           <a href="/bugs/3156">3156</a>
+          <a href="/bugs/3201">3201</a>
           <a href="/bugs/3215">3215</a>
           <a href="/bugs/3248">3248</a>
         </goab-side-menu-group>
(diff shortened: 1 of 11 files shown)


apps/prs/angular/src/app/app.component.html
```diff
@@ -47,6 +47,7 @@
           <a href="/bugs/3072">3072</a>
```
chrisolsen: The commit message doesn't make sense and needs a small update.

libs/web-components/src/components/input/Input.svelte
```diff
@@ -224,14 +224,17 @@
...
   function setValue(detail: FieldsetSetValueRelayDetail) {
     // @ts-expect-error
     value = detail.value;
     dispatchOnChange(value);
   }
 
-  function dispatchOnChange(value: string) {
-    dispatch(_rootEl, "_change", { name, value }, { bubbles: true });
+  function dispatchOnChange(nextValue: string) {
+    _lastEmittedValue = nextValue ?? "";
```
chrisolsen: This value doesn't seem to be used anywhere.

willcodeforcoffee: Can we not just attach to the change event for the input itself?

willcodeforcoffee (approved):
LGTM

Issue #2873, opened 2025-07-09
If Drawer has scrolling - clicking an interactive element focuses but doesn't activate


ArakTaiRoth:
# Info

Reported by Jenna Granitto

If you create a drawer component with enough content to activate scrolling in it, and you scatter interactive elements throughout it. When you click an interactive element near the top or near the bottom, it will focus on that item (moving the scrollbar as well), which is good, but it doesn't activate that element, you would have to click it again to activate it.

Seen in [jam](https://jam.dev/c/be802c61-88ff-40fc-adb1-bc9df1c90240)

## Acceptance Criteria:

1. Clicking an interactive element will always focus and activate that element inside the Drawer component

PR #3307, opened 2026-01-06
fix(#2873): Updated Drawer to not auto-scroll when an element is clicked

ArakTaiRoth:
# Before (the change)

Using a Details or Accordion component inside a Drawer, would cause the Drawer to scroll to that component but not activate the component, so you ended up needing to click the component twice.

# After (the change)

Now the Drawer will not auto-scroll when you click Details or Accordion, and will activate the component. Scrolling to interactive elements is retained though if using keyboard navigation.

**NOTE** No specific tests were added for this, as this is a purely visual issue

## Steps needed to test

Use the included PR files for both Angular and React. Bug 2873.


ArakTaiRoth:
@chrisolsen This is good to be merged

Issue #3323, opened 2026-01-12
Tooltip inside Callout gets cut off


ArakTaiRoth:
# Info

Reported by Dylan Hall

If you use a tooltip inside a callout component, the tooltip will get cutoff by the borders of the Callout.
<img width="358" height="105" alt="Image" src="https://github.com/user-attachments/assets/82825d15-b376-47f6-be9f-0cfded7b90c3" />

## More Info

It appears this also happens inside a Container

## Possible Workaround

You can use the `position` property to change where the tooltip shows up, but this all depends on how much content you have in your callout, and in your tooltip, and where something can be placed to work.

## Acceptance Criteria:

1. The Callout should allow a Tooltip to go beyond it's area

vanessatran-ddi:
This also happens if inside a container: 
https://design.alberta.ca/components/container/#example-show-full-date-in-a-tooltip 
<img width="1042" height="380" alt="Image" src="https://github.com/user-attachments/assets/4ea5b156-460a-42e7-8b46-2c907041b500" />

PR #3324, opened 2026-01-12
fix(#3323): Modified Callout to use overflow:visible

ArakTaiRoth:
# Before (the change)

Callout wouldn't allow a Tooltip component to expand beyond its borders

# After (the change)

Callout should now allow anything like a pop-up (Tooltip) to expand beyond its borders

**NOTE:** no new tests have been added as this is a purely visual change

## Steps needed to test

1. Run `npm run serve:prs:angular` and `npm run serve:prs:react`. Test using Bug 3323.

### DIFF
--- a/libs/web-components/src/components/callout/Callout.svelte
+++ b/libs/web-components/src/components/callout/Callout.svelte
@@ -136,7 +136,7 @@
   .notification {
     display: flex;
     align-items: stretch;
-    overflow: hidden;
+    overflow: visible;
     font: var(--goa-callout-l-text-size);
     border: var(--goa-callout-l-border-width) solid;
     border-radius: var(--goa-callout-border-radius);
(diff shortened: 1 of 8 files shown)


libs/web-components/src/components/callout/Callout.svelte
```diff
@@ -136,7 +136,7 @@
   .notification {
     display: flex;
     align-items: stretch;
-    overflow: hidden;
+    overflow: visible;
```
chrisolsen: I believe that the reason the overflow was initially hidden was to prevent ugly color "bleeding" when border radius was initially set. Can you verify that changing the `overflow` prop does not result in this. BTW I believe that the `overflow` attr could just be removed.

ArakTaiRoth:
Closing, as we don't want this fix any longer

PR #3409, opened 2026-02-10
feat(#3344): add multi-column sorting to Table and TableSortHeader

### LINKED ISSUE #3344: Add optional multi-column sorting (primary + secondary) to the table component
## Summary
Add support for multi-column sorting (primary + secondary) in `goa-table`, enabled via a component property. This should be opt-in so teams can continue using single-column sorting when that meets their requirements.

## Reference implementation
multi column sort implemented as desired on [this example](https://workspace-demo-v4.netlify.app/search)

UX / interaction requirements
- Single sort (default)
  - Clicking a sortable column header applies that column as the only active sort (current behavior).
- Multi sort (opt-in)
  - Define user interaction to set secondary sort (Click second column header to add/modify secondary sort).
  - Define how to remove secondary sort and revert to single sort
    - 1st click - Sorted (arrow up) - asc
    - 2nd click - Sorted (arrow down) - desc
    - 3rd click - Sort removed

Accessibility considerations
- Sort state must be conveyed to assistive tech.
- keyboard Interaction for setting sort primary and secondary must be maintained.
- Provide clear announcements/labels for primary vs secondary sort state.

Acceptance criteria
- table component has a new property to enable multi-column sorting (primary + secondary).
- Single sort configuration remains supported and documented.
- When multi-sort is enabled:
  - Users can set a primary and a secondary sort (each with asc/desc).
  - Sorting is applied as: primary first, then secondary.
  - UI indicates primary vs secondary sort order and direction.
  - Keyboard interaction supports configuring secondary sort.
  - Sort state is exposed to screen readers appropriately.

twjeffery:
## Summary

Adds built-in multi-column sorting to Table and TableSortHeader components.

- **sortMode** prop: `"single"` (default) or `"multi"` (up to 2 columns)
- **initialSort** prop: Pre-configure sort state
- **onSortChange** callback: Receives array of current sorts
- **sortOrder** prop on TableSortHeader: Shows "1", "2" priority badges
- Table manages sort state internally, headers update automatically
- New icons: `chevron-expand` (unsorted), `arrow-up`/`arrow-down` (sorted)
- V2 focus ring on icon only

### Multi sort
image

### Single sort
image

## Test plan

- [x] Visit React playground at `/features/3344`
- [x] Test single-column sort (Test 1) - click headers, verify only one sorts at a time
- [x] Test multi-column sort (Test 2) - click multiple headers, verify "1" and "2" badges appear
- [x] Test initial sort (Test 3) - verify Department starts sorted
- [x] Test V2 experimental wrappers (Test 4) - verify focus ring on icon only
- [x] Visit Angular playground at `/features/3344` and repeat tests

Closes #3344

### DIFF
--- a/libs/web-components/src/components/table/Table.svelte
+++ b/libs/web-components/src/components/table/Table.svelte
@@ -1,18 +1,24 @@
-<svelte:options customElement={{
-  tag: "goa-table",
-  props: {
-    variant: { reflect: true },
-    version: { reflect: true },
-    striped: { reflect: true }
-  }
-}} />
+<svelte:options
+  customElement={{
+    tag: "goa-table",
+    props: {
+      variant: { reflect: true },
+      version: { reflect: true },
+      striped: { reflect: true },
+      sortMode: { attribute: "sort-mode", reflect: true },
+    },
+  }}
+/>
 
 <script lang="ts">
-  import { onMount, tick } from "svelte";
+  import { onMount } from "svelte";
   import { calculateMargin } from "../../common/styling";
-  import { typeValidator, toBoolean } from "../../common/utils";
-  import type { GoATableSortDirection } from "./TableSortHeader.svelte";
+  import { typeValidator, toBoolean, dispatch } from "../../common/utils";
   import type { Spacing } from "../../common/styling";
+
+  // Types
+  type SortDirection = "asc" | "desc";
+  type SortEntry = { column: string; direction: SortDirection };
 
   // Validators
   const [Variants, validateVariant] = typeValidator(
@@ -24,6 +30,13 @@
 
   const [Version, validateVersion] = typeValidator("Version", ["1", "2"]);
   type VersionType = (typeof Version)[number];
+
+  const [SortModes, validateSortMode] = typeValidator(
+    "Sort mode",
+    ["single", "multi"],
+    { required: true },
+  );
+  type SortMode = (typeof SortModes)[number];
 
   // Public
 
@@ -39,6 +52,8 @@
   export let version: VersionType = "1";
   /** Sets a data-testid attribute for automated testing. */
   export let testid: string = "";
+  /** Sort mode: "single" allows one column, "multi" allows up to 2 columns. */
+  export let sortMode: SortMode = "single";
 
   /** Top margin. */
   export let mt: Spacing = null;
@@ -53,6 +68,9 @@
 
   let _rootEl: HTMLElement;
   let _isTableRoot: boolean = false;
+  let _sorts: SortEntry[] = [];
+  let _headings: NodeListOf<Element> | undefined;
+  const MAX_SORTS = 2;
 
   // Reactive
 
@@ -64,9 +82,7 @@
   onMount(() => {
     validateVariant(variant);
     validateVersion(version);
-
-    // without setTimeout it won't properly sort in Safari
-    setTimeout(attachSortEventHandling, 0);
+    validateSortMode(sortMode);
 
     // exit here if when running tests (tests don't have assignedElements)
     const slot = _rootEl.querySelector("slot") as HTMLSlotElement;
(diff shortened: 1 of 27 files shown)


libs/web-components/src/components/table/Table.svelte
```diff
@@ -83,56 +103,134 @@
...
+    });
+
+    // Apply initial state to headers and dispatch initial event
+    updateHeaderAttributes();
+    if (_sorts.length > 0) {
+      dispatchSortEvent();
+    }
+  }
+
+  function initializeSorts() {
```
bdfranck: You can remove this function when you switch to using an array of `SortEntries` for the sorts.

libs/react-components/src/lib/table/table.tsx
```diff
@@ -1,14 +1,26 @@
...
 
+// Note: JSX.IntrinsicElements for "goa-table" is declared in experimental/table/table.tsx
+
 /* eslint-disable-next-line */
 export interface GoabTableProps extends Margins {
   width?: string;
+  /** @deprecated Use onSortChange for new implementations */
   onSort?: (detail: GoabTableOnSortDetail) => void;
+  /** Called when sort state changes. Receives array of current sorts. */
+  onSortChange?: (sorts: GoabTableSortEntry[]) => void;
```
bdfranck: Instead of building a parallel function that does the same thing as `onSort`, why not expand `onSort` to support either a single item or an array?

`GoabTableOnSortDetail[] | GoabTableOnSortDetail`

This would allow you to add new functionality while preserving the original structure.

bdfranck (commented):
The sorting interactions are looking good! I've left some comments that can be grouped into three categories:

1. Try to re-use and expand on existing variables and functions.
2. Avoid using JSON because it can lead to increased errors. Use Types instead.
3. Make sure that you're giving devs the data they need for their sort functions.

libs/web-components/src/components/table/TableSortHeader.svelte
```diff
@@ -53,56 +68,81 @@
...
-    scale: 0.8;
+  /* V2: Icon color on focus for unsorted state */
+  button.v2:not(.sorted):focus-visible goa-icon {
+    color: var(--goa-color-interactive-hover);
   }
 
-  .direction--none goa-icon {
-    height: 0.625rem;
+  /* Sort order number badge */
+  .sort-order {
```
vanessatran-ddi: I tested and see there is a small shift in a column:
https://jam.dev/c/8bafde76-cd5d-4d17-8e1e-68912f31a9b1

I think we should add a hidden class for this .sort-order, for example:
`<span class='sort-order' class:hidden={!sortorder || direction === 'none'>`

Then we can make sure `width:0` and `margin:0` to prevent the shifting.

libs/web-components/src/components/table/TableSortHeader.svelte
```diff
@@ -25,24 +37,27 @@
...
-  {:else if direction === "asc"}
-    <goa-icon type="caret-up" size="small" />
-  {:else}
-    <div class="direction--none">
-      <goa-icon type="caret-up" size="small" />
-      <goa-icon type="caret-down" size="small" />
-    </div>
-  {/if}
+  <span class="icon-wrapper">
+    {#if sortorder && direction !== "none"}
```
vanessatran-ddi: This is a bit difficult to read, because there are 2 `if` blocks here with different conditions. 

Suggested: (so else here means no direction)
```
<span class="icon-wrapper">
    {#if direction === "desc"}
      {#if sortorder}<span class="sort-order">{sortorder}</span>{/if}
      <goa-icon type="arrow-down" size="3" />
    {:else if direction === "asc"}
      {#if sortorder}<span class="sort-order">{sortorder}</span>{/if}
      <goa-icon type="arrow-up" size="3" />
    {:else}
      <goa-icon type="chevron-expand" size="3" />
    {/if}
  </span>
```

libs/web-components/src/components/table/Table.svelte
```diff
@@ -39,6 +51,10 @@
   export let version: VersionType = "1";
   /** Sets a data-testid attribute for automated testing. */
   export let testid: string = "";
+  /** Sort mode: "single" allows one column, "multi" allows up to 2 columns. */
+  export let sortmode: SortMode = "single";
+  /** Initial sort configuration as JSON string: [{"column":"name","direction":"asc"}] */
+  export let initialsort: string = "";
```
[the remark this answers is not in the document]
vanessatran-ddi: I agree vs the point that we shouldn' use json. 
I think instead of using `initialsort` as a JSON string on the table  (which is a new property), We could declare the initial sort state directly on each `goa-table-sort-header` (which we do nowadays)
For example:
```
<goa-table sortmode="multi">
... <goa-table-sort-header name="name" direction="asc" sortorder="1">Name</goa-table-sort-header>
... <goa-table-sort-header name="department" direction="desc" sortorder="2">Department</goa-table-sort-header>
...<goa-table-sort-header name="salary">Salary</goa-table-sort-header>
</goa-table>
```

So we only add a new property `sortorder` to the `goa-table-sort-header`. This approach will be consistent with how the old single-sort already worked. The table just scans headers `onMount` to build its internal sort state from `direction` + `sortorder` attribute.

libs/web-components/src/components/table/Table.svelte
```diff
@@ -14,6 +15,10 @@
   import type { GoATableSortDirection } from "./TableSortHeader.svelte";
   import type { Spacing } from "../../common/styling";
 
+  // Types
+  type SortDirection = "asc" | "desc";
```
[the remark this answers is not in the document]
vanessatran-ddi: It is on purpose. SortDirection here only have `asc` and `desc`, to handle active sorts only, while `GoabTableSortDirection` includes the unsorted state.

willcodeforcoffee (approved):
👍  Looks good!

bdfranck (approved):
- ✅ I see the JSON was removed
- ✅ The PR playground examples work as expected in Angular and React
- ✅ All tests pass

Looks good to me! 👍

libs/web-components/src/components/table/TableSortHeader.svelte
```diff
@@ -1,48 +1,85 @@
...
         _rootEl.style.setProperty("--header-align", "right");
       }
+
+      dispatch<GoATableSortHeaderProps>(
+        _rootEl,
+        "table-sort-header:mounted",
+        { el: host, name, direction, sortOrder },
+        { bubbles: true, timeout: 10 },
+      );
     }
```
chrisolsen: What is the reason for adding the `dispatch` call? Since all the props are reflected all the values are accessible via the attributes.

libs/web-components/src/components/table/Table.svelte
```diff
@@ -80,60 +112,136 @@
 
   // Functions
```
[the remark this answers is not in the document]
twjeffery: It's now separated into `handleSingleSort` and `handleMultiSort`. The shared work (updating header attributes and dispatching the event) stays in `handleSortClick` so it's not duplicated across both functions.

libs/web-components/src/components/table/Table.svelte
```diff
@@ -53,6 +68,9 @@
 
   let _rootEl: HTMLElement;
   let _isTableRoot: boolean = false;
+  let _sorts: SortEntry[] = [];
+  let _headings: NodeListOf<Element> | undefined;
+  const _maxSorts = 2;
```
chrisolsen: For clarity, since this is a const value this should be renamed to `MAX_SORTS`

chrisolsen (commented):
Just one more small thing, then I am good merging it. @bdfranck @willcodeforcoffee do you guys need to re-review?

willcodeforcoffee (approved):
Testing works, looks good!
Still! 😉 😆

PR #3410, opened 2026-02-10
chore: Fixed the intermittent failing browser tests

ArakTaiRoth:
# Before (the change)

We had multiple intermittent failures in various browser tests.

Components affected:
* Date Picker
* Menu Button
* Radio
* Text Area
* Temporary Notification

# After (the change)

Hopefully no more intermittent failures. Though I did notice some failures in `datagrid.browser.spec.tsx`, but I never modified that file as it's quite different from the rest of our browser tests.

## Steps needed to test

Explanation of changes made:

* `datepicker.browser.spec.tsx`
    * Modified `renders with props` to use an exact date as a value instead of basing it off today's date (which cares about timezone)
    * Modified waitFor under `browser event` to detect the element only, and then click it outside of the waitFor, this should avoid a race condition
    * Removed checking the month dropdown in the `disabled` test, I couldn't get that to work reliable, if you have a better way let me know. Also changed it to check the actual elements, this seemed to work better.

* `textarea.browser.spec.tsx`
    * Modified a couple waitFor statements to only check if an object exits, clicking the object is done outside. Same reason as above

* `radio.browser.spec.tsx`
    * Modified it to always re-get the elements involved in the enable/disable test. This should help due to React re-rendering wonkiness

* `menu-button.browser.spec.tsx`
    * Same waitFor changes as above made in this file as well

* `temporary-notification-ctrl.browser.spec.tsx`
    * Same waitFor changes as above

Also updated the `vitest` package to 4.0.13 (actually 4.0.18 I think), so it's the same as the other vitest packages (`@vitest/browser` for example). We were getting version mismatch errors, and that could potentially cause other issues with tests.

### DIFF
--- a/libs/react-components/specs/datepicker.browser.spec.tsx
+++ b/libs/react-components/specs/datepicker.browser.spec.tsx
@@ -2,7 +2,7 @@ import { render } from "vitest-browser-react";
 import { GoabDatePicker } from "../src";
 import { expect, describe, it, vi } from "vitest";
 import { userEvent } from "@vitest/browser/context";
-import { addDays, format, addMonths, addYears } from "date-fns";
+import { format } from "date-fns";
 
 describe("DatePicker", () => {
   it("renders", async () => {
@@ -19,7 +19,7 @@ describe("DatePicker", () => {
   });
 
   it("renders with props", async () => {
-    const value = addDays(new Date(), -10);
+    const value = "2026-01-30";
     const Component = () => {
       return <GoabDatePicker testId="date-picker" value={value} />;
     };
@@ -29,7 +29,7 @@ describe("DatePicker", () => {
 
     await vi.waitFor(() => {
       const inputEl = input.element() as HTMLInputElement;
-      expect(inputEl.value).toBe(format(value, "MMMM d, yyyy"));
+      expect(inputEl.value).toBe("January 30, 2026");
     });
   });
 
@@ -65,11 +65,11 @@ describe("DatePicker", () => {
 
     await userEvent.click(input);
 
-    await vi.waitFor(async () => {
-      const dateEl = dateToSelect.element() as HTMLElement;
-      expect(dateEl).toBeTruthy();
-      await userEvent.click(dateEl);
+    await vi.waitFor(() => {
+      expect(dateToSelect.element()).toBeTruthy();
     });
+
+    await dateToSelect.click();
 
     await vi.waitFor(() => {
       expect(handleChange).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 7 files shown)


libs/react-components/specs/datepicker.browser.spec.tsx
```diff
@@ -402,14 +402,13 @@ describe("Date Picker input type", () => {
...
 
     await vi.waitFor(() => {
-      expect(monthInput).toBeDisabled();
-      expect(datePickerDay).toBeDisabled();
-      expect(datePickerYear).toBeDisabled();
+      const datePickerDay = result.getByTestId("input-day").element() as HTMLInputElement;
+      const datePickerYear = result.getByTestId("input-year").element() as HTMLInputElement;
+
+      expect(datePickerDay.disabled).toBe(true);
+      expect(datePickerYear.disabled).toBe(true);
```
chrisolsen: How to move the locators out of the waitFor and use the locators to perform the test on
```
    const result = render(<Component />);

    const datePickerDayLoc = result.getByTestId("input-day");
    const datePickerYearLoc = result.getByTestId("input-year");

    await vi.waitFor(() => {
      expect(datePickerDayLoc).toBeDisabled();
      expect(datePickerYearLoc).toBeDisabled();
    });

```

libs/react-components/specs/menu-button.browser.spec.tsx
```diff
@@ -93,46 +86,59 @@ describe("MenuButton", () => {
 
     const result = render(<Component />);
     const menuButton = result.getByTestId("menu-button");
+    const firstAction = result.getByTestId("first-action");
+    const secondAction = result.getByTestId("second-action");
 
-    // Click first action
```
chrisolsen: These comments also make things much easier to understand why things are being done the way they are.

libs/react-components/specs/menu-button.browser.spec.tsx
```diff
@@ -19,61 +24,49 @@ describe("MenuButton", () => {
...
-
-    // Test each menu action
-    for (let i = 1; i <= 3; i++) {
-      await vi.waitFor(async () => {
-        // open menu
-        await menuButton.click();
-
-        const menuAction = result.getByTestId(`menu-action-${i}`);
-        expect(menuAction).toBeDefined();
+      const menuAction = result.getByTestId(`menu-action-${i}`);
```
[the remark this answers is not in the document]
ArakTaiRoth: Didn't know about the `nth()` function, thanks.

vanessatran-ddi (dismissed):
I tested on my local, it passed all the test after I clean nx and dist and re-build.

vanessatran-ddi (approved):

chrisolsen (approved):

PR #3414, opened 2026-02-11
feat(#3229): allow text optional & size to MenuButton

### LINKED ISSUE #3229: Allow menu button vs `icon` only
Right now, the menu button doesn't allow `text` optional.
However, in a design such as Workspace table, the action column, we want ... only, instead of text

Image

## Acceptance Criteria:
* Allow menu button to have only icon
* Add `variant`, `size`, and `theme`

vanessatran-ddi:
# Before (the change)

image

# After (the change)

image

### DIFF
--- a/libs/react-components/specs/menu-button.browser.spec.tsx
+++ b/libs/react-components/specs/menu-button.browser.spec.tsx
@@ -1,6 +1,7 @@
 import { render } from "vitest-browser-react";
 import { vi } from "vitest";
 import { GoabMenuAction, GoabMenuButton } from "../src";
+import { GoabxMenuButton, GoabxMenuAction } from "../src/experimental";
 
 describe("MenuButton", () => {
   it("should render and trigger action when clicked", async () => {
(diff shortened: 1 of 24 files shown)


twjeffery (commented):
Looks good overall @vanessatran-ddi, just a few things I noticed that we could improve:

**1. Size prop in icon-only mode**

I noticed the size prop gets passed to goa-button but not goa-icon-button. We'd want to use different icon button sizes in icon-only mode too. Also IconButton uses numeric sizes ("1", "2", etc.) while MenuButton uses "normal" | "compact", so the values would need mapping.

**2. Default icon fallback**

When no leadingIcon is provided in icon-only mode, it falls back to the chevron. For a standalone action menu, could we change it to ellipsis-horizontal as the default?

**3. Accessibility for icon-only**

In icon-only mode, the goa-icon-button doesn't get an arialabel passed to it, so screen readers would just announce "button" with no context. Should MenuButton set a default label like "Open menu" or add a prop for this?

libs/react-components/specs/menu-button.browser.spec.tsx
```diff
@@ -146,3 +147,83 @@ describe("MenuButton", () => {
...
+    };
+
+    const result = render(<Component />);
+    const menuButton = result.getByTestId("icon-compact-menu");
+
+    await vi.waitFor(async () => {
+      await menuButton.click();
+
+      const assignAction = result.getByTestId("action-assign");
+      expect(assignAction).toBeDefined();
```
chrisolsen: `assignAction` is a selector and will always be truthy

twjeffery (approved):
Looks good to me now @vanessatran-ddi , approved

ArakTaiRoth (approved):

PR #3422, opened 2026-02-12
fix(#3273): keep Side Menu Group open

### LINKED ISSUE #3273: SideMenuGroup doesn't stay open
Issue reported by: andrzej.zadorozny@gov.ab.ca
Web component version: Dev
Angular/React version: Dev

Also reported by Thang Lam

**NOTE:** This has been heavily modified as the original issue was reported in Vue. I have tested in Angular and fully confirmed the issue and thus changed all of the code, and included new info.

## Expected Behaviour

Side menu should stay open (i.e. expanded) and highlight the current page the user is on.

## Actual Behaviour

When the side menu contains any side menu group inside another side menu group, and the user navigates to a page / link, the side menu tree collapses and does not show the current page the user is on.

I have noticed it only works correctly for the last nested group, but not for the others.

## Replication Steps

Here is code (Angular)

```html
<goab-side-menu>
  <goab-side-menu-group heading="Example 1">
    <a href="testing">Item 1</a>
    <a href="testing/2">Item 2</a>

    <goab-side-menu-group heading="Item 3">
      <a href="testing/31">Item 3.1</a>
      <a href="testing/32">Item 3.2</a>
      <a href="testing/33">Item 3.3</a>
    </goab-side-menu-group>

    <goab-side-menu-group heading="Item 4">
      <a href="testing/41">Item 4.1</a>
      <goab-side-menu-group heading="Item 4.2">
        <a href="testing/421">Item 4.2.1</a>
        <a href="testing/422">Item 4.2.2</a>
        <a href="testing/423">Item 4.2.3</a>
      </goab-side-menu-group>
      <a href="testing/43">Item 4.3</a>
    </goab-side-menu-group>

    <goab-side-menu-group heading="Item 5">
      <a href="testing/51">Item 5.1</a>
      <a href="testing/52">Item 5.2</a>
      <a href="testing/53">Item 5.3</a>
    </goab-side-menu-group>
    <a href="testing/6">Item 6</a>
    <a href="testing/7">Item 7</a>
  </goab-side-menu-group>
</goab-side-menu>
```

Even simpler example:

```html
<goab-side-menu>
  <goab-side-menu-group heading="Example 1">
    <a href="testing">Item 1</a>
    <a href="testing/2">Item 2</a>

    <goab-side-menu-group heading="Item 3">
      <a href="testing/31">Item 3.1</a>
      <a href="testing/32">Item 3.2</a>
      <a href="testing/33">Item 3.3</a>
    </goab-side-menu-group>
  </goab-side-menu-group>
</goab-side-menu>
```

[jam.dev for example 1](https://jam.dev/c/2dfb6994-8804-4aa1-a81e-335858960f07)
[jam.dev for example 2](https://jam.dev/c/a7371373-9a4a-491e-88d2-60123e8375ee)

bdfranck:
# Before 

When a parent Side Menu Group is current and open,
But a child Group is not current or open,
The parent group will be closed.

menu-group-broken

# After 

When a parent Side Menu Group is current and open,
But a child Side Menu Group is not current or open,
The parent Group is still current and open.

menu-group-fixed

### DIFF
--- a/apps/prs/react/src/main.tsx
+++ b/apps/prs/react/src/main.tsx
@@ -40,6 +40,8 @@ import { Bug3215Route } from "./routes/bugs/bug3215";
 import { Bug3232Route } from "./routes/bugs/bug3232";
 import { Bug3279Route } from "./routes/bugs/bug3279";
 import { Bug3248Route } from "./routes/bugs/bug3248";
+import { Bug3273Route } from "./routes/bugs/bug3273";
+import { Bug3273Page2Route } from "./routes/bugs/bug3273-2";
 import { Bug3275Route } from "./routes/bugs/bug3275";
 import { Bug3322Route } from "./routes/bugs/bug3322";
 import { Bug3281Route } from "./routes/bugs/bug3281";
@@ -125,6 +127,8 @@ root.render(
           <Route path="bugs/3215" element={<Bug3215Route />} />
           <Route path="bugs/3232" element={<Bug3232Route />} />
           <Route path="bugs/3248" element={<Bug3248Route />} />
+          <Route path="bugs/3273" element={<Bug3273Route />} key={"bugs/3273"} />
+          <Route path="bugs/3273-2" element={<Bug3273Page2Route key={"bugs/3273-2"} />} />
           <Route path="bugs/3275" element={<Bug3275Route />} />
           <Route path="bugs/3322" element={<Bug3322Route />} />
           <Route path="bugs/3281" element={<Bug3281Route />} />
--- a/libs/react-components/specs/side-menu.browser.spec.tsx
+++ b/libs/react-components/specs/side-menu.browser.spec.tsx
@@ -1,0 +1,115 @@
+import { render } from "vitest-browser-react";
+import { GoabSideMenu, GoabSideMenuGroup } from "../src";
+import { describe, it, expect, vi } from "vitest";
+
+describe("SideMenu", () => {
+  const Component = () => {
+    return (
+      <GoabSideMenu testId="side-menu">
+        <GoabSideMenuGroup heading="Parent Group" testId="parent-group">
+          <a href="/parent-item">Parent Item</a>
+
+          <GoabSideMenuGroup heading="Child Group" testId="child-group">
+            <a href="/child-item" data-testid="child-item-link">
+              Child Item
+            </a>
+          </GoabSideMenuGroup>
+        </GoabSideMenuGroup>
+      </GoabSideMenu>
+    );
+  };
+
+  const isGroupOpen = (groupHost: HTMLElement | null) => {
+    if (!groupHost) return;
+    const container = groupHost.shadowRoot?.querySelector('[data-testid="group"]');
+    return !!container && !container.classList.contains("hidden");
+  };
+
+  const updateURL = async (url: string) => {
+    window.history.pushState({}, "", url);
+    window.dispatchEvent(new PopStateEvent("popstate"));
+  };
+
+  beforeEach(() => {
+    window.history.pushState({}, "", "/");
+  });
+
+  it("closes parent and child groups when no items are current", async () => {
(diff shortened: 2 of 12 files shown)


libs/web-components/src/components/side-menu-group/SideMenuGroup.svelte
```diff
@@ -78,7 +78,7 @@
 
     // listen to events by children (if child is open the parent also has to be open)
     _rootEl.addEventListener("_open", (e: Event) => {
-      _open = _current = (e as CustomEvent).detail.current;
+      if ((e as CustomEvent).detail.current) _current = _open = true;
```
ArakTaiRoth: Should probably have a test for this. Testing if a parent item is clicked, and then a child item is clicked, that the parent item stays expanded.

apps/prs/react/src/main.tsx
```diff
@@ -120,6 +122,8 @@ root.render(
           <Route path="bugs/3215" element={<Bug3215Route />} />
           <Route path="bugs/3232" element={<Bug3232Route />} />
           <Route path="bugs/3248" element={<Bug3248Route />} />
+          <Route path="bugs/3273" element={<Bug3273Route />} key={"bugs/3273"} />
+          <Route path="bugs/3273-2" element={<Bug3273Route key={"bugs/3273-2"} />} />
```
ArakTaiRoth: You're pointing to the wrong route on this line. Should be using `Bug3273Page2Route` I think.

libs/react-components/specs/side-menu.browser.spec.tsx
```diff
@@ -0,0 +1,111 @@
...
+            <a href="/child-item" data-testid="child-item-link">
+              Child Item
+            </a>
+          </GoabSideMenuGroup>
+        </GoabSideMenuGroup>
+      </GoabSideMenu>
+    );
+  };
+
+  const handler = vi.fn();
```
ArakTaiRoth: You're declaring `handler` globally for all tests to use. This could cause `toHaveBeenCalled` functions to pass only because it was called in a previous test, not because it was called in the current test.

apps/prs/react/src/main.tsx
bdfranck: Good catch! I've updated the route.

ArakTaiRoth (approved):

willcodeforcoffee (approved):
It seems to be working correctly now. LGTM 👍

libs/web-components/src/components/side-menu-group/SideMenuGroup.svelte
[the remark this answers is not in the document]
bdfranck: @chrisolsen Here's how the bug was happening. If we have an embedded group like the one below...

```
<GoabSideMenu>
  <GoabSideMenuGroup heading="Parent Group">
    <Link to="/parent">Parent Link</Link>
    <GoabSideMenuGroup heading="Child Group">
      <Link to="/child">Child Link</Link>
    </GoabSideMenuGroup>
  </GoabSideMenuGroup>
</GoabSideMenu>
```

...the Parent Group receives two `_open` events. So when the Parent Link is `current`, the following happens oh load:

1. The Parent Group receives an `_open` event from Parent Link. The Group opens because `detail.current` is `true`.
2. The Parent Group receives an `_open` event from the Child Link. The Group closes again because `detail.current` is `false`.

But we never need to close a group because it's closed by default. My change only opens a group if `details.current` is `true`. It does nothing when it is `false`.

Is there a better way to address this? Have I neglected to consider other scenarios? I'm happy to chat about it if you have further questions.

chrisolsen (approved):

chrisolsen:
@ArakTaiRoth @willcodeforcoffee changes were made, so you guys might want to test things again; at least one of you.

willcodeforcoffee (approved):
✅ LGTM!

PR #3496, opened 2026-02-25
Security updates

chrisolsen:


vanessatran-ddi (approved):
I tested again vs angular 21.1.5 and angular playground works as normal.

Image

PR #3515, opened 2026-03-03
fix(#3397): use existing design tokens in Astro docs

### LINKED ISSUE #3397: V2 container body text styles not showing V2 text styling in Astro doc site
### Expected

Content inside `goa-container` should render with V2 text styling (not semi condensed).

Image

### Actual

Content shows as v1 text styling (semi-condensed)

Image

bdfranck:
# Before (the change)

The Astro layouts and components for our v2 Docs used custom design tokens that overwrote some existing ones.

# After (the change)

The Astro layouts and components for our v2 Docs uses existing design tokens.

### DIFF
--- a/docs/src/components/CodeSnippet.tsx
+++ b/docs/src/components/CodeSnippet.tsx
@@ -540,7 +540,7 @@ function CodeSnippetStyles() {
         border-radius: var(--goa-border-radius-m, 4px);
         overflow: hidden;
         background: var(--goa-color-greyscale-100, #f1f1f1);
-        font-size: 0.875rem;
+        font-size: var(--goa-font-size-2);
       }
 
       .framework-switcher .code-snippet {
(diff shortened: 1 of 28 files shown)


twjeffery (commented):
error

twjeffery (commented):
Looks good @bdfranck . A few things I noticed that we should address:

1. Other files still use the removed custom properties

Most of the removed :root properties were duplicates of real V2 tokens (--goa-color-text-secondary, --goa-space-*, etc.) so those will keep working fine. But a few were not real V2 tokens, and other files still reference them:

- ExampleCard.astro uses --goa-color-border, --goa-color-background-light, --goa-color-brand, --goa-radius-s/m
- ComponentSandbox.tsx uses --goa-color-border, --goa-color-brand, --goa-radius-m (these have inline fallbacks so they won't break visually, but still worth migrating)
- GuidanceGrid.astro uses --goa-color-background-light, --goa-radius-m

We should either keep those properties in :root for now, or update those files too. We can create a separate issue for migrating them if we want to keep this PR focused.

2. Inline code loses its background

code background went from --goa-color-background-light (#f5f5f5, light grey) to --goa-color-greyscale-white (#fff). Since the page background is also white, inline code blocks won't have any visual distinction. Maybe something like --goa-color-greyscale-100 to keep the contrast?

3. Mobile header hardcoded values

font-size: 18px, margin-left: -8px, background: white. Since we're already in this file, these should move to tokens too.

vanessatran-ddi (commented):
I have some comments about using a token instead of hard-code values

twjeffery (approved):
@bdfranck Looks good, thanks for migrating the other files too. All the token names check out.

bdfranck:
@vanessatran-ddi I've replaced the instances you mentioned and did a find and replace of similar ones.

vanessatran-ddi (approved):

PR #3534, opened 2026-03-05
Chris/npm audit x

chrisolsen:
Libs are now specific versions, which ensures everyone is using the same version. 

A new scripts was added to auto update libs to the latest minor version, with a 5 day wait, to avoid any releases that are followed by a patch the next day (or five)


willcodeforcoffee (commented):
@chrisolsen there are still a lot of deprecation warnings and 28 vulnerabilities

This seems like a good opportunity to fix some of those, too?

** Oops I accidentally editted your response when trying to quote you **

willcodeforcoffee (approved):

vanessatran-ddi (approved):

Issue #3511, opened 2026-03-03
Design and build cross-version navigation between DS 1.0 and DS 2.0 docs


Spark450:
## Summary
As DS 2.0 becomes the primary documentation at **design.alberta.ca**, we still need to support teams using **DS 1.0**.

This issue covers the design and implementation of an experience that helps users:
- quickly understand whether they’re viewing DS 1.0 or DS 2.0 content
- quickly switch (“off-ramp”) to the other version when needed

## Goals
- Make the current docs version obvious (DS 1.0 vs DS 2.0)
- Provide a fast, low-friction way to switch versions
- Reduce user confusion and support questions during the migration period
- Preserve deep linking and allow users to keep their place where possible

## Experience requirements
The final experience must include:
- Clear version identification so users can immediately tell which version they are reading
- A visible switch / off-ramp in both directions:
  - DS 1.0 → DS 2.0 (for current guidance and recommendations)
  - DS 2.0 → DS 1.0 (for teams maintaining older implementations)

## Open design questions (to resolve during design/dev work)
- What design pattern works best?
  - Banner, version switcher, persistent header element, interstitial landing page, etc.

## Suggested solution directions (examples only)
Pick one (or combine) based on what’s simplest and clearest:
- A persistent version badge + switch link
- A dismissible notification banner at the top of every page
- A version landing page that routes users to DS 1.0 or DS 2.0
- Contextual links like “Looking for the DS 1.0 version of this component?”

## Acceptance criteria
- DS 1.0 pages clearly indicate “Design System 1.0” and include a prominent link to DS 2.0
- DS 2.0 pages clearly indicate “Design System 2.0” and include less-prominent linking to DS 1.0
- The switch/off-ramp is visible without scrolling on desktop, and remains easy to find on mobile
- Switching works reliably
- If page-to-page matching is implemented, unmapped pages fall back gracefully (no 404s; clear fallback)

## Implementation notes
- Coordinate with the “[DS 1.0 hosting and URL](https://github.com/GovAlta/ui-components/issues/3510)” issue so the links and routing match the final decision:
  - Path-based (`/v1`) or subdomain (`v1.design.alberta.ca`)
- Prefer a solution that is easy to maintain and adjust during the migration period.

## Deliverables
- Proposed UX pattern (Figma) with copy (optional)
- Implemented UI in the docs site(s)

PR #3554, opened 2026-03-11
feat(#3511): add redirects from DS 1 docs

bdfranck:
This PR prepares our DS 2 docs for links from DS 1 docs with the following changes...

- Adds DS 1 docs links to the bottom of individual component and example pages
- Adds redirects from DS 1 docs URLs including...
  - Components with different paths (ex: `/components/circular-progress-indicator` ➡️ 	`/components/circular-progress`)
  - ~Paths with version segments (ex: `/components/v6.0.0+-react/button`)~ (I removed this b/c it caused a deploy issue)
- Changes the side menu heading to title case: "Design System"

image

image

The related DS 1 docs changes: https://github.com/GovAlta/ui-components-docs/pull/484

### DIFF
--- a/docs/astro.config.mjs
+++ b/docs/astro.config.mjs
@@ -12,6 +12,19 @@ const workspaceRoot = path.resolve(__dirname, "..");
 export default defineConfig({
   root: ".",
   outDir: "../dist/docs",
+  redirects: {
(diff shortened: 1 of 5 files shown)


Spark450:
@bdfranck are we able to get a deploy preview running for this?

PR #3556, opened 2026-03-11
feat(#3300, docs): addition of motion foundations page

### LINKED ISSUE #3300: Create motion guidelines page on the DS 2.0 site
## Summary
Publish motion guidance on the new design system website, including overview and tokens

## Acceptance criteria

- Page includes content (drawn from [Figma docs "overview"](https://www.figma.com/design/wZn6HSRBFCXiEt3e5buvPl/Design-System-2.0-Update?node-id=7033-217843&t=tfQMjQaiODQzglHp-1) and token values
- Includes reduced motion guidance (suggested location: Foundations > Style guide > Motion)

mxsoco:
# Summary

- Added Motion page to Foundations section
- For Tables and Grids, added ResponsiveDisplay.tsx file to differentiate the mobile and desktop views. Uses the useCompactToolbar hook.
- CodeCopy.tsx is added for all of the "copy to dashboard" buttons.
- Images have been added for the curve types.

## Test plan

- [ ] Navigate to /foundations, go to /foundations/motion.
- [ ] Responsiveness works and views switch from Table (desktop) to Grid (mobile) and vice-versa.
- [ ] Verify the "copy to dashboard" buttons work.


docs/src/pages/foundations/motion.astro
```diff
@@ -0,0 +1,477 @@
...
+  </goa-text>
+
+  <h2 id="easing">Easing</h2>
+  <goa-text size="body-m" mt="l" mb="l">
+    Easing controls how motion accelerates and decelerates during a transition. It affects how clear and predictable state changes feel. Use the curves below to keep motion consistent across products:
+  </goa-text>
+  <goa-text size="body-m" mt="l" mb="l">
+    <ul>
+      <li>
+        <strong>expressive easing</strong> is the default for larger page-level transitions
```
Spark450: Capitalize "Expressive"

docs/src/pages/foundations/motion.astro
```diff
@@ -0,0 +1,477 @@
...
+
+  <h2 id="accessibility">Accessibility</h2>
+  <goa-text size="body-m" mt="l" mb="l">
+    While animations can make a product eye-catching and interesting, it has the potential to cause real issues for users. Some people might feel dizzy or nauseous when confronted with certain types of motion. Take the following accessibility guidelines into account when working with motion to provide a great user experience:
+  </goa-text>
+  <goa-text size="body-m" mt="l" mb="2xl">
+    <ul>
+      <li>
+        <strong>Minimize flashing</strong>: Web pages should not contain anything that flashes more than three times in any one-second period, as it can trigger
+        epileptic physical reactions especially for users who have vestibular disorders. If you do need to include flashing content, ensure that it is below the general flash and red flash thresholds.
```
Spark450: Can you put "general flash" and "red flash" in quotes?

docs/src/pages/foundations/motion.astro
```diff
@@ -0,0 +1,477 @@
...
+      </li>
+      <li>
+        <strong>Design alternatives for reduced motion</strong>: Users should have control on how they view animations. They should have the ability to disable animations unless they are essential to the functionality or information being conveyed. Use the <code>prefers-reduced-motion</code> CSS media query to check if someone has asked for less animation. If they have, make sure that your product respects those settings.
+      </li>
+    </ul>
+  </goa-text>
+</DocumentationPageLayout>
```
Spark450: I think we have an updated "bottom of the page callout". You can see it at the bottom of this page:
https://deploy-preview-3546--goa-design-2.netlify.app/get-started/

Spark450 (commented):
Looking good Miguel, just a few changes to make.

Also it looks like you and might be using different page margins. Would you be able to connect with him to make sure we are aligned in our approach.

Spark450 (approved):
Thanks for the changes 👍

twjeffery (commented):
@mxsoco It looks great overall. A few things I noticed:

1. **File/export naming mismatch** The file is `CodeCopy.tsx` but the export is `TokenSnippet`. Should we rename one to match the other? Since it's specifically for copying token names, maybe `TokenCopy` for both?

2. **Copy value** The component prepends `--` to the token name before copying, which gives the CSS custom property format. But teams using SCSS would need the `$` prefix instead. Should we just copy the plain token name and let teams add their own prefix?

3. **Inline styles in TokenSnippet** The component includes a `<style>` tag in JSX, so the same CSS gets added to the page for every copy button. Could we move those styles into the `<style>` block at the bottom of motion.astro? They're already page-specific and there's a style block there for the other motion page styles.

4. **Comment cleanup in ParentMenu.tsx** There's a leftover comment line that says `* - : direct navigation (when content exists)` where "Foundations" was removed but the line wasn't cleaned up.

twjeffery (approved):

PR #3590, opened 2026-03-16
feat(#3589): add breadcrumb navigation to docs website

### LINKED ISSUE #3589: Add breadcrumb navigation to docs website
The docs website doesn't have breadcrumbs, so there's no way to see where you are in the site hierarchy or navigate back up.

## What this adds

An Astro breadcrumb component that auto-generates the ancestor trail from the URL path. Shows at least the section name on every page (Components, Get Started, Tokens, etc).

For component pages, the category group (e.g. "Inputs and actions") is included in the trail and links back to All Components with that category pre-filtered.

Current page is visually hidden but available to screen readers via aria-current.

## Approach

Building this as an Astro component first, specific to our docs website. This gives us a proof of concept we can use and learn from before designing a shared Svelte breadcrumb component for the UI component library.

## Pages affected
- Documentation pages (Get Started, Foundations)
- Component pages (with category in trail)
- All Components, Examples, Tokens index pages
- Example detail pages (replaces "Back to all examples" link)

twjeffery:
## Summary
- Adds an auto-generated breadcrumb component that derives the navigation trail from the URL path
- Shows at least the section name on every page (Components, Get Started, Tokens, etc)
- Component pages include the category group (e.g. "Inputs and actions") linking back to All Components with that category pre-filtered
- Replaces category badge on component pages and "Back to all examples" link on example pages
- Keyboard-only focus styles, visually-hidden current page for screen readers (aria-current)

Closes #3589

## Approach
Built as an Astro component first for our docs site as a proof of concept. We can learn from this before designing a shared Svelte breadcrumb component for the UI component library.

## File-by-file walkthrough

**Breadcrumbs.astro** (new, 142 lines)
The core component. Derives the breadcrumb trail from the URL path automatically. Each segment gets a label from a lookup map (or falls back to formatting the slug). Always shows at least the section name. Has an optional `parentGroup` prop for when the logical hierarchy isn't in the URL (only used by component pages for the category). Current page is visually hidden but available to screen readers via `aria-current="page"`. Uses design tokens for all styling, focus-visible for keyboard-only focus rings.

**ComponentsGrid.tsx** (+16 lines)
Adds URL param reading on mount so `?category=inputs-and-actions` pre-filters the grid. This is what makes the breadcrumb category link work. Same pattern ExamplesGrid already uses.

**DocumentationPageLayout.astro** (+10 lines)
Adds breadcrumbs inside the content card. `breadcrumb-row` has a 960px max-width to match the content grid centering.

**examples/index.astro** (+2 lines) — Adds breadcrumbs to the All 
  Examples page.

**TokensLayout.astro** (+9 lines)
Same as above, 1408px max-width.

**components/[slug].astro** (+5/-4 lines)
Passes the component's category as `parentGroup` with a link to `/components?category=...`. Removes the old category badge that was above the title (breadcrumbs replace that context).

**components/index.astro** (+2 lines)
Adds breadcrumbs to the All Components page.

**examples/[slug].astro** (-10 lines)
Adds breadcrumbs inside the page template (not the layout) so they align with the narrower content width. Removes the "Back to all examples" link and its CSS since breadcrumbs handle that navigation now.

image
image
image

## Test plan
- [ ] Navigate to /components/button and verify breadcrumb shows Design System / Components / Inputs and actions /
- [ ] Click "Inputs and actions" in the breadcrumb and confirm it goes to /components with the category filter applied
- [ ] Navigate to /get-started/designers and verify breadcrumb shows Design System / Get Started /
- [ ] Check section landing pages (/components, /tokens, /examples) show the section name
- [ ] Tab through breadcrumb links to verify focus ring only appears on keyboard nav, not on click
- [ ] Test with a screen reader to confirm aria-current="page" is announced

### DIFF
--- a/docs/src/components/Breadcrumbs.astro
+++ b/docs/src/components/Breadcrumbs.astro
@@ -1,0 +1,159 @@
+---
+/**
+ * Breadcrumbs.astro
+ *
+ * Auto-generates ancestor breadcrumb trail from the current URL path.
+ * Always shows at least the section name (e.g. "Components", "Get Started").
+ * Use parentGroup for logical parents not reflected in the URL (e.g. component category).
+ */
+
+interface Props {
+  /** Optional logical parent not reflected in the URL (e.g. component category) */
+  parentGroup?: { label: string; url?: string };
+}
+
+const { parentGroup } = Astro.props;
+const pathname = Astro.url.pathname.replace(/\/$/, '');
+const segments = pathname.split('/').filter(Boolean);
+
+// Show on any page within a section
+const showBreadcrumbs = segments.length >= 1;
+
+// Build set of valid page routes from the pages directory
+const pageFiles = import.meta.glob('/src/pages/**/*.{astro,md,mdx}', { eager: false });
+const validPaths = new Set(
+  Object.keys(pageFiles).map(file =>
+    file
+      .replace('/src/pages', '')
+      .replace(/\/index\.(astro|md|mdx)$/, '')
+      .replace(/\.(astro|md|mdx)$/, '')
+      .replace(/\[.*?\]/g, '*') // dynamic routes aren't valid intermediate targets
+  )
+);
+
+// Known labels for path segments
+const SEGMENT_LABELS: Record<string, string> = {
+  "get-started": "Get Started",
+  "foundations": "Foundations",
+  "components": "Components",
+  "tokens": "Tokens",
+  "examples": "Examples",
+  "designers": "Designers",
+  "developers": "Developers",
+};
+
+function segmentToLabel(segment: string): string {
+  if (SEGMENT_LABELS[segment]) return SEGMENT_LABELS[segment];
+  // Fallback: kebab-case to sentence case
+  return segment.split('-').map((w, i) =>
+    i === 0 ? w.charAt(0).toUpperCase() + w.slice(1) : w
+  ).join(' ');
+}
+
+// Build ancestor crumbs: all segments except the last, but always include the section
+const crumbs: Array<{ label: string; url: string; hasPage: boolean }> = [];
+let path = '';
+const crumbCount = Math.max(1, segments.length - 1);
+for (let i = 0; i < crumbCount; i++) {
+  path += '/' + segments[i];
+  crumbs.push({ label: segmentToLabel(segments[i]), url: path, hasPage: validPaths.has(path) });
+}
+---
+
+{showBreadcrumbs && (
+  <nav class="breadcrumbs" aria-label="Breadcrumb">
+    <ol>
+      <li>
+        <a href="/">Design System</a>
+        <span class="separator" aria-hidden="true">/</span>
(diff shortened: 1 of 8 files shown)


docs/src/components/Breadcrumbs.astro
```diff
@@ -0,0 +1,142 @@
...
+  ).join(' ');
+}
+
+// Build ancestor crumbs: all segments except the last, but always include the section
+const crumbs: Array<{ label: string; url: string }> = [];
+let path = '';
+const crumbCount = Math.max(1, segments.length - 1);
+for (let i = 0; i < crumbCount; i++) {
+  path += '/' + segments[i];
+  crumbs.push({ label: segmentToLabel(segments[i]), url: path });
```
ArakTaiRoth: This is usually great, but there are situations where a generated portion of the URL doesn't actually generate a workable page. Take the URL `http://localhost:4203/foundations/style-guide/colour` for example, takes you to Colour, and generates this as the breadcrumb:

Image

The problem is, "Style guide" in that breadcrumb creates a link to `/foundations/style-guide`, which isn't an actual page.

I think we need some checking in here to verify that a created link is actually valid.

docs/src/components/Breadcrumbs.astro
```diff
@@ -0,0 +1,142 @@
...
+          {parentGroup.url ? (
+            <a href={parentGroup.url}>{parentGroup.label}</a>
+          ) : (
+            <span>{parentGroup.label}</span>
+          )}
+          <span class="separator" aria-hidden="true">/</span>
+        </li>
+      )}
+      <li class="sr-only" aria-current="page">
+        {segmentToLabel(segments[segments.length - 1])}
```
ArakTaiRoth, line 92: This is likely pretty minor, but maybe an issue (I'm not sure). What the screen reader reads here doesn't line up with the header of the page in all situations:
1. App Header - reads App Header - title reads Header
2. Radio Group - reads Radio Group - title reads Radio
3. File Upload Input - reads File Upload Input - title reads File uploader
4. Linear Progress - reads Linear Progress - title reads Linear Progress Indicator

docs/src/components/Breadcrumbs.astro
twjeffery: I added a check that collects all valid page routes at build time. If an intermediate segment doesn't have a page, it shows as plain text instead of a link. So on the colour page you'd see "Style guide" as text and "Foundations" as a link.

docs/src/components/Breadcrumbs.astro
twjeffery: Yeah, the sr-only text comes from the URL slug, so it won't always match the h1 exactly. I think it's close enough for now since it's only for screen readers and the context is still clear. We could add a prop to pass the real title if it becomes an issue.

ArakTaiRoth (approved):

chrisolsen (approved):

PR #3615, opened 2026-03-20
spike(#3347): sticky header utility classes

### LINKED ISSUE #3347: Create reusable sticky header/footer component
## Summary
_Investigate if creating a component for this capability is the best approach to provide this capability._ 

Create reusable components (or a single component with header/footer positioning capability) to support “sticky” UI regions within scrollable pages.

## Context / reference
- Workspace demo shows:
  - a search toolbar that remains visible at the top while scrolling
  - a bulk action footer that remains visible at the bottom while scrolling (often conditional on having selected items)

Reference implemented version in the [workspace demo](https://workspace-demo-v4.netlify.app/cases#tab-0) for UX/UI

Accessibility requirements
- Keyboard focus must never be hidden behind sticky regions.
- Sticky regions should not trap focus; tab order remains natural.

Responsive requirements
- Works across supported breakpoints.
- Sticky header/footer should remain usable on smaller screens (elements wrap or truncate appropriately without breaking the layout).

Acceptance criteria
- A reusable component(s) exists for sticky header and sticky footer regions.
- Implemented behaviour matches the workspace demo intent (search toolbar at top; bulk actions at bottom).
- Works inside a scroll container without requiring consumers to write custom CSS.
- Create component docs page

## Open questions
- Should we provide an optional “offset” prop (e.g., if there’s an existing fixed app header inside the scroll region)?
- Can the scope include swapping the component into workspace template?
- Should we [consider using an HTML "section" to define a "region" landmark.](https://www.w3.org/WAI/ARIA/apg/patterns/landmarks/examples/region.html) This is to enable screen reader users to be able to quick navigate to these sections.

bdfranck:
This spike explores one sticky header idea: what if we used CSS utility classes?

```
<div class="scroll-container">
  <div class="sticky">
    <goab-container>This is sticky</goab-container>
  </div>
  <!-- tons of content goes here -->
</div>
```

### Benefits

- Lightweight & high performance
- Only uses CSS
- SImple and flexible

### Drawbacks

- Goes against our current Svelte component pattern
- Teams may have trouble implementing it if they don't know how `position: sticky` works
- Still needs JS to add shadows

Would more simply utility classes like this be useful for simple layout assistance?

### DIFF
--- a/libs/web-components/src/assets/css/utils.css
+++ b/libs/web-components/src/assets/css/utils.css
@@ -1,0 +1,31 @@
+/**
+ * CSS Utilities
+ * Common utility classes used across components
+ */
+
+:root {
+  /* Sticky utility custom properties */
+  --sticky-top: 0;
+  --sticky-z-index: 10;
+  --scroll-container-height: auto;
+}
+
+/**
+ * Sticky Utility Classes
+ * Classes for creating sticky positioned elements
+ */
+
+/* Base sticky class - makes element stick to top of its scrollable parent */
+.sticky {
+  position: sticky;
+  top: var(--sticky-top);
+  z-index: var(--sticky-z-index);
+  width: 100%;
+}
+
+/* Base scroll container - ensures proper setup for sticky children */
+.scroll-container {
+  overflow-y: auto;
+  position: relative;
+  height: var(--scroll-container-height);
+}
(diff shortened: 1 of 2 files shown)


bdfranck:
I'm closing this PR as it is an exploratory branch just for reference.

PR #3666, opened 2026-03-24
fix(#3541) - adjust modal so left side of Checkbox is not cut off

### LINKED ISSUE #3541: Checkbox inside a Modal is cut off on the left side
# Info

Implementing a GoabxCheckbox inside a GoabModal, sees the left side of the checkbox when focused or giving an error get cut off:

Image

This is only an issue with v2 components and styling.

## Replication Code (Angular)

```html
<goabx-form-item label="Select interests" mb="l">
  <goabx-checkbox name="sports" text="Sports" />
  <goabx-checkbox name="music" text="Music" />
  <goabx-checkbox name="travel" text="Travel" />
  <goabx-checkbox name="reading" text="Reading" />
</goabx-form-item>
```

## Acceptance Criteria:

1. Checkbox should not be cut off inside a Modal

mxsoco:
# Before (the change)
On focus, left side of the checkbox gets cut off by the modal container.
image

# After (the change)
After adjusting the padding of the Modal and Scrollable container, the left side now displays on modal.
image

## Steps needed to test
- [ ] Ensure that v1 styles for modal remains the same
- [ ] Confirm that checkbox left side does not get cut off by the side of the "scrollable" container.
- [ ] Check if checkbox left side still appears when modal has a scrollbar.

### DIFF
--- a/apps/prs/angular/src/app/app.component.html
+++ b/apps/prs/angular/src/app/app.component.html
@@ -233,10 +233,18 @@
           url="/bugs/3498"
         ></goabx-work-side-menu-item>
         <goabx-work-side-menu-item
+          label="3541 Checkbox cut off on the left side in modal"
+          url="/bugs/3541"
+        ></goabx-work-side-menu-item>
+        <goabx-work-side-menu-item
           label="3607 Radio and Checkbox Interaction Area"
           url="/bugs/3607"
         ></goabx-work-side-menu-item>
         <goabx-work-side-menu-item label="3505 Link Icon Click" url="/bugs/3505"></goabx-work-side-menu-item>
+        <goabx-work-side-menu-item
+          label="3541 Checkbox cut off on the left side in modal"
+          url="/bugs/3541"
+        ></goabx-work-side-menu-item>
       </goabx-work-side-menu-group>
       <goabx-work-side-menu-group icon="star" heading="Features">
         <goabx-work-side-menu-item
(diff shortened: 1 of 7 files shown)


libs/web-components/src/components/modal/Modal.svelte
```diff
@@ -289,7 +289,8 @@
           <div data-testid="modal-content" class="modal-content">
             <goa-scrollable
               direction="vertical"
-              hpadding="var(--scrollable-padding)"
+              hpadding={version === "2" ? "var(--goa-space-2xs)" : "var(--scrollable-padding)"}
```
chrisolsen: Is this override required? If things are version 2 then the --scrollable-padding should be overridden already.

bdfranck (commented):
Instead of modifying the component itself, I suggest that we update the [Modal design tokens](https://github.com/GovAlta/design-tokens/blob/a7079480e87525876eeefd959d89956a8dfe7cbd/data/component-design-tokens/modal-design-tokens.json#L113-L157).


When we move the left/right padding out of `--goa-modal-content-padding`...

```
  "modal-content-padding": {
    "value": "{space.l} 0 {space.xl} 0",
    "type": "spacing",
    "description": "Content section padding"
  },
```

...and into `--goa-modal-scrollable-padding-desktop`...

```
  "modal-scrollable-padding-desktop": {
    "value": "{space.l}",
    "type": "spacing",
    "description": "Scrollable component horizontal padding for desktop"
  },
```

...and do the same for the mobile versions...

```
  "modal-content-padding-mobile": {
    "value": "{space.l} 0 {space.xl} 0",
    "type": "spacing",
    "description": "Content section padding for mobile (24px top, 16px sides, 32px bottom)"
  },
```

```
  "modal-scrollable-padding-mobile": {
    "value": "{space.m}",
    "type": "spacing",
    "description": "Scrollable component horizontal padding for mobile"
  },
```

...then we can achieve the same result without altering the component itself.

Image

PR #3697, opened 2026-03-26
fix(#3635): add in new color token for input leading icon

### LINKED ISSUE #3635: Input: Leading icon color
## Description

The leading icon in `GoabxInput` uses the default icon color, which is too dark. It should use a lighter color so it reads as secondary to the input text.

## Expected

- The leading icon should use `--goa-color-greyscale-600` (and/or a new component token is created: `--goa-text-input-colour-icon`) so it feels like a visual hint rather than competing with the input text.

## Screenshots
Current

Image

Design

Image

mxsoco:
# Before (the change)
The colour of the leading icon shared the same colour as the input text.

image

# After (the change)
With the new token added, the leading icon is lighter than the input text.

image

## Steps needed to test
- [ ] Confirm that v2 input, both default and compact sizes, are both using the  `--goa-text-input-colour-icon` design token.

### DIFF
--- a/libs/web-components/src/components/input/Input.svelte
+++ b/libs/web-components/src/components/input/Input.svelte
@@ -591,6 +591,11 @@
     margin-right: var(--goa-text-input-padding-lr);
   }
 
+  .leading-icon,
+  .trailing-icon {
+    color: var(--goa-text-input-color-icon);
+  }
+
   .trailing-icon-button {
     margin-right: var(--goa-text-input-padding-lr);
   }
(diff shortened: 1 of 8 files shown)


bdfranck (commented):
Almost there! I left a couple comments.

bdfranck (approved):

The leading and trailing icons in a v2 input are `greyscale.600`. ✅  

Image

The leading and trailing icons in a v1 input (with v1 tokens) are the default colour. ✅ 

Image

Looks good to me! 👍

libs/web-components/src/components/input/Input.svelte
```diff
@@ -651,6 +651,11 @@
     cursor: default !important;
   }
 
+  .leading-icon,
+  .trailing-icon {
+    color: var(--goa-text-input-color-icon);
+  }
+
   /* Adjust the leading icon style when input is disabled */
```
chrisolsen: Can this be moved up to line ~590 to keep it close the other matching selectors?

twjeffery (approved):

PR #3702, opened 2026-03-27
feat(docs): add inline search with slash commands to grid pages

### LINKED ISSUE #3710: Add inline search with slash commands to docs grid pages
Bring the site-wide FlexSearch into the grid pages (Components, Examples, Tokens) as inline search, scoped to each section. Replaces the basic string-match inputs that were there before.

**What it does:**
- Inline search uses the same FlexSearch index as the global Cmd+K search, scoped to the relevant content type
- Type `/` to open a slash command menu for quick filter access (category, status, scale, user type) without opening the filter drawer
- Keyboard navigation throughout (arrow keys, enter, escape)
- Consistent UX across all three grid pages

PR: #3702

twjeffery:
## Summary

- Replace basic string-match search inputs on Components, Examples, and Tokens grids with FlexSearch-powered inline search
- Add contextual slash commands (type `/`) that toggle grid filters with a stepped dimension picker
- Consistent search UX across all three grid pages

## What it does
https://github.com/user-attachments/assets/40a80315-808b-48d1-958c-e193b429f576

Type in the search input to filter the grid by relevance. Type `/` to open a slash command menu that lets you quickly apply filters (category, status, scale, user type) without opening the filter drawer. Multi-dimension pages (Components, Examples) show a dimension picker first, single-dimension pages (Tokens) go straight to values.

Keyboard navigation: arrow keys to navigate, right arrow to drill into a dimension, left/backspace to go back, escape to close, enter to toggle a filter.

## Test plan

- [ ] `/components` - search narrows grid, `/` shows Category + Status dimensions
- [ ] `/examples` - search narrows grid, `/` shows Scale, Category, User Type dimensions
- [ ] `/tokens` - search narrows grid, `/` shows category values directly
- [ ] Arrow key navigation works (up/down, right to drill in, left to go back)
- [ ] Slash command filters stay in sync with filter drawer checkboxes
- [ ] Clear all resets search + filters + sort
- [ ] Cmd+K global search modal still works independently
- [ ] No flash of unstyled content on page load

### DIFF
--- a/docs/src/components/ComponentsGrid.tsx
+++ b/docs/src/components/ComponentsGrid.tsx
@@ -23,13 +23,14 @@ import {
   GoabFormItem,
   GoabIcon,
   GoabIconButton,
-  GoabInput,
   GoabPushDrawer,
 } from "@abgov/react-components";
 
 import { useTwoLevelSort } from "../hooks/useTwoLevelSort";
 import { useContainerNarrow } from "../hooks/useContainerWidth";
 import { useViewSettings } from "../hooks/useViewSettings";
+import { InlineSearch, type SlashCommand } from "./search/InlineSearch";
+import { useSearch } from "./search/useSearch";
 
 // Type for component data (matches Astro content collection)
 export interface Component {
@@ -102,10 +103,8 @@ function getCategoryBadgeType(
 
 // Format category for display
 function formatCategory(category: string): string {
-  return category
-    .split("-")
-    .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
-    .join(" ");
+  const words = category.replace(/-/g, " ");
+  return words.charAt(0).toUpperCase() + words.slice(1);
 }
 
 // Format status for display
@@ -137,10 +136,9 @@ function getThumbnailPath(slug: string): string {
 
 export function ComponentsGrid({ components }: ComponentsGridProps) {
   // State
-  const [searchValue, setSearchValue] = useState("");
-  const [searchChips, setSearchChips] = useState<string[]>([]);
   const [filterDrawerOpen, setFilterDrawerOpen] = useState(false);
   const [isSticky, setIsSticky] = useState(false);
+  const [searchQuery, setSearchQuery] = useState("");
 
   // Ref for grid container (used by ResizeObserver for container-aware breakpoints)
   const gridRef = useRef<HTMLDivElement>(null);
@@ -182,6 +180,7 @@ export function ComponentsGrid({ components }: ComponentsGridProps) {
   const { sortConfig, setSortConfig, sortByKey, clearSort, handleTableSort } =
     useTwoLevelSort();
   const isContainerNarrow = useContainerNarrow(gridRef, 624);
+  const { search, isLoading, error } = useSearch();
   const { viewSettings, setLayout } = useViewSettings({
     pageKey: "components",
     defaultLayout: "card", // 'card' = grid view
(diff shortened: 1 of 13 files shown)


twjeffery:
> @twjeffery needs a rebase

@chrisolsen Done

docs/src/components/ComponentsGrid.tsx
```diff
@@ -182,6 +180,7 @@ export function ComponentsGrid({ components }: ComponentsGridProps) {
   const { sortConfig, setSortConfig, sortByKey, clearSort, handleTableSort } =
     useTwoLevelSort();
   const isContainerNarrow = useContainerNarrow(gridRef, 624);
+  const { search } = useSearch();
```
ArakTaiRoth: Maybe not an issue, I'm not sure. But because we're using `useSearch` here, any search results that are filtered on inside this file, are reduced to seeing only 50 results before any filters here are even run. This could be fine, but nowhere in our UX do we state "the top 50 results" or something like that, so people might be left wondering. Basically, the `/` filters run on the 50 results, not all of the potential results, so the numbers given might be misleading.

twjeffery: I see what you mean. The limit was set to 50, I bumped it to 500. Also made it configurable in case we need to adjust it later.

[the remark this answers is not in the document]
twjeffery: Fixed. The search input now shows "Loading search..." while it's loading and "Search unavailable" if something goes wrong.

docs/src/components/search/InlineSearch.tsx
```diff
@@ -0,0 +1,382 @@
...
+  );
+
+  const displayValue = commandMode ? commandInput : value;
+  const showClose = commandMode || !!value;
+  const hasDropdownItems = showDimensions
+    ? filteredDimensions.length > 0
+    : filteredCommands.length > 0;
+
+  const effectivePlaceholder = error
+    ? "Search unavailable"
```
ArakTaiRoth: This doesn't work I don't think. In this screenshot I've blocked requests to search-index to simulate an error:

Image

This just modifies the placeholder of the search input. But they've already typed something the search input, so a placeholder change does nothing because there's already text in the input. Handling it like the modal I think would be preferable:

Image

docs/src/components/TokensGrid.tsx
```diff
@@ -151,6 +142,7 @@ export function TokensGrid({ tokens, filterGroups }: TokensGridProps) {
   const { sortConfig, setSortConfig, sortByKey, clearSort, handleTableSort } =
     useTwoLevelSort();
   const isContainerNarrow = useContainerNarrow(gridRef, 840);
+  const { search, isLoading, error } = useSearch();
```
ArakTaiRoth: We're coupling this page to `useSearch`, but it isn't actually used anywhere in this page, the page seems to use a local page search instead.

docs/src/components/search/InlineSearch.tsx
twjeffery: I switched it to show an error dropdown below the input, similar to how the modal handles it.

docs/src/components/TokensGrid.tsx
twjeffery: Good call, removed it. The tokens page uses its own substring matching so it doesn't need the search hook at all.

chrisolsen (approved):

docs/src/components/ComponentsGrid.tsx
```diff
@@ -182,6 +180,7 @@ export function ComponentsGrid({ components }: ComponentsGridProps) {
   const { sortConfig, setSortConfig, sortByKey, clearSort, handleTableSort } =
     useTwoLevelSort();
   const isContainerNarrow = useContainerNarrow(gridRef, 624);
+  const { search, error } = useSearch();
```
ArakTaiRoth: You're still not using `isLoading`, in this file and ExamplesGrid. This is different from the Modal Search and the Search page. Potentially on slower connections this could result in 0 results shown instead of a loading message.

twjeffery: Good catch, I added isLoading back. The search input should now show a "Loading search..." message below it while the index is loading, but only when someone has actually started typing.

ArakTaiRoth (approved):

ArakTaiRoth:
@chrisolsen This is ready to be merged

PR #3709, opened 2026-03-30
fix(#3548): pin secondary menu at bottom of Work Side Menu during scroll

### LINKED ISSUE #3548: Work Side Menu: Secondary menu should stay pinned at bottom during scroll
## Description

When the primary nav content in the Work Side Menu is long enough to require scrolling, the entire area below the logo scrolls, including the secondary menu items (Search, Get support, Release notes, Collapse menu).

The secondary menu should stay pinned at the bottom of the sidebar, with only the primary nav area scrolling.

## Current behavior

The scroll region includes everything below the logo/header. When you scroll down, the secondary menu items scroll off-screen. When you scroll to the bottom, the primary nav items scroll behind the logo.

## Expected behavior

- Logo/header: fixed at top
- Primary nav: scrollable
- Secondary menu (Search, Get support, Release notes, Collapse menu): fixed at bottom

### Current (dev)

https://github.com/user-attachments/assets/d1550b80-0943-4994-87b9-cd8778cc0919

### Expected (Figma)

Image

Image

twjeffery:
## Summary

- Restructures the Work Side Menu so only the primary nav area scrolls, while the secondary menu (Search, Get support, Release notes), account section, and collapse toggle stay pinned at the bottom
- Adds scroll-aware border and shadow indicators that fade in/out (300ms) based on scroll position, only showing when there is hidden content in that direction
- Adds browser test verifying secondary menu remains visible when primary nav overflows

https://github.com/user-attachments/assets/98aa5743-a78f-479f-9f52-000c4d284d02

### Top of scroll
image
### Middle of scroll
image
### Bottom of scroll
image
### No scroll
image

Closes #3548

## Test plan

- [ ] Open the PR playground at `/bugs/3548` which has enough primary items to overflow
- [ ] Verify primary nav scrolls independently while secondary menu stays pinned at bottom
- [ ] Verify header stays fixed at top, items don't scroll behind it
- [ ] Verify border and shadow appear/fade when scrolled away from top or bottom
- [ ] Verify border and shadow disappear when scrolled fully to either end
- [ ] Collapse the menu and verify layout still works at 72px width
- [ ] Resize the viewport height and confirm the layout adapts
- [ ] Test on mobile viewport (menu opens as full-screen modal)

### DIFF
--- a/libs/react-components/specs/work-side-menu.browser.spec.tsx
+++ b/libs/react-components/specs/work-side-menu.browser.spec.tsx
@@ -1,11 +1,7 @@
 import { render } from "vitest-browser-react";
 import { useState } from "react";
 import { GoabButton } from "../src";
-import {
-  GoabWorkSideMenu,
-  GoabWorkSideMenuItem,
-  GoabWorkSideMenuGroup,
-} from "../src";
+import { GoabWorkSideMenu, GoabWorkSideMenuItem, GoabWorkSideMenuGroup } from "../src";
 import { expect, describe, it, vi } from "vitest";
 import { page } from "@vitest/browser/context";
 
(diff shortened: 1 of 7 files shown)


libs/web-components/src/components/work-side-menu/WorkSideMenu.svelte
ArakTaiRoth, line 132: This function doesn't recalculate `_isScrolling`, neither does the function `handleWindowResize()`. This creates the issue shown in this [jam](https://jam.dev/c/e08a4e5f-640a-4c5a-a724-ffb9ebf9e0c1). Specifically the CSS visual style of the overflow that was created in this PR is only shown on page load if it exists. If it's created dynamically by expanding a menu or by resizing the window, it doesn't get shown.

libs/react-components/specs/work-side-menu.browser.spec.tsx
```diff
@@ -240,6 +236,52 @@ describe("WorkSideMenu", () => {
     });
   });
 
+  describe("Scroll pinning (#3548)", () => {
```
ArakTaiRoth, line 239: This test doesn't accurately replicate what happens when the overflow is triggered dynamically, such as by opening menus or resizing the window. So the issue I stated above isn't caught.

libs/web-components/src/components/work-side-menu/WorkSideMenu.svelte
twjeffery: Good catch. The scroll state wasn't being recalculated when content height changed dynamically. I added a ResizeObserver on the primary menu element so `setMenuScrolling()` fires whenever the content size changes, whether from a group expanding, collapsing, or a window resize. Also added `updateScrollPosition()` so the border/shadow state stays in sync when overflow is first detected.

libs/react-components/specs/work-side-menu.browser.spec.tsx
twjeffery: Added a second test that starts with a collapsed group that fits without overflow, expands it to trigger overflow dynamically, then asserts the scroll indicators appear.

bdfranck (commented):
Looks good! I have a couple suggestions that I added in a commit. You can squash them if you agree. Changes include:

- Give the top section the class name `.top-section` for consistency
- Remove bottom padding on primary menu to get rid of the extra space
- Use tokens for new values (https://github.com/GovAlta/design-tokens/pull/144)

libs/web-components/src/components/work-side-menu/WorkSideMenu.svelte
```diff
@@ -74,21 +77,49 @@
     await tick();
     addEventListeners();
     observer = watchPathChanges(setCurrentUrl);
+
+    if (typeof ResizeObserver !== "undefined") {
+      _resizeObserver = new ResizeObserver(() => setMenuScrolling());
```
ArakTaiRoth: Potential problem with this. The ResizeObserver fires `setMenuScrolling`, which has the potential to change the observed `_menuEl`, which would cause it to fire again. I can't come up with an actual test that will have this problem, just noticed it as a potential issue we may want to do something about now.

libs/web-components/src/components/work-side-menu/WorkSideMenu.svelte
```diff
@@ -74,21 +77,49 @@
...
+    _resizeObserver?.disconnect();
   });
 
   // *********
   // Functions
   // *********
 
   // Menu
+  function updateScrollPosition() {
+    const innerEl = _scrollEl?.shadowRoot?.querySelector(".goa-scrollable");
```
ArakTaiRoth: I'm not sure if this will work or not. But the `Scrollable.svelte` dispatches `_scroll` event that I think should be used for this instead if we can. Otherwise we're linked to the component in a more brittle fashion. You can reference `Drawer.svelte` on line 181 for how it's used there.

twjeffery:
Quick update on this one @bdfranck @ArakTaiRoth: 

I got the suggestions and review points sorted out locally and tests are still passing. 

I reviewed and merged the token PR from Benji, but the npm release failed due to a bad token. Now waiting on another fix PR to re-trigger the publish before I squash everything together and push

libs/web-components/src/components/work-side-menu/WorkSideMenu.svelte
twjeffery: Good call. I switched it to use the `_scroll` event for the scroll position instead of reaching into the shadow DOM, same approach as `Drawer.svelte`.

libs/web-components/src/components/work-side-menu/WorkSideMenu.svelte
twjeffery: I wrapped the callback in `performOnce` with its own timer so it doesn't interfere with the existing one in `addMenuLink`, Which should prevent any re-triggering.

twjeffery:
@bdfranck @ArakTaiRoth Should be good to review again now

libs/web-components/src/components/work-side-menu/WorkSideMenu.svelte
ArakTaiRoth: @twjeffery Not quite the same issue:
1. `ResizeObserver` fires, line 84
2. It schedules `setMenuScrolling`, line 85
3. That function can change `_isScrolling`, line 111
4. `_isScrolling` controls the `scrolling` class, line 349
5. The `.scrolling` class can change the width, line 550
6. The `ResizeObserver` in step 1 above, is watching `_menuEl`, which is inside that container, and thus can have its width change, which would refire ResizeObserver

bdfranck (approved):

- ✅ Borders and shadows work as expected
- ✅ There isn't extra space between the primary and secondary menus

Image

Looks good to me! 👍

libs/web-components/src/components/work-side-menu/WorkSideMenu.svelte
```diff
@@ -74,21 +79,49 @@
...
+
+    if (typeof ResizeObserver !== "undefined") {
+      _resizeObserver = new ResizeObserver(() => {
+        _resizeTimeoutId = performOnce(_resizeTimeoutId, setMenuScrolling, 1);
+      });
+      if (_menuEl) _resizeObserver.observe(_menuEl);
+    }
   });
 
   onDestroy(() => {
```
ArakTaiRoth, line 91: Sorry, last thing for this one. You set a lot of timeout ID's. `_bindTimeoutId`, `_resizeTimeoutId`, `_mouseEnterTimeoutId`, and `_mouseLeaveTimeoutId`. But none of them are ever cleared in `onDestroy`. Refer to the `onDestroy` function in `AppHeader.svelte` for an example.

twjeffery: Good catch. I added clearTimeout for all four in onDestroy. Three of them were pre-existing but never cleaned up, so this covers those too.

ArakTaiRoth (approved):

PR #3718, opened 2026-03-31
chore: Updating web-components from 1.x to 2.x

ArakTaiRoth:
# Before (the change)

We referenced web-components 1.x in the migration guide. This will not actually be factual in our current release, as our web-components will be upgrading to 2.x

# After (the change)

The references was changed to reference 2.x instead.


Spark450 (approved):
Thanks Dustin

bdfranck (approved):
Image

Looks good to me! 👍

PR #3727, opened 2026-03-31
fix(#3726): highlight last TOC heading when scrolled to bottom

### LINKED ISSUE #3726: Docs: sticky inline TOC highlights wrong section for short sections near bottom of page
The table of contents sidebar highlights the wrong section when navigating to a short section near the bottom of the page. For example, on the workspace example page, clicking "Get started" highlights "Components built for workspace" instead.

Image

### Why it happens

The TOC active state is based on scroll position. It finds the last heading with `top < 120px`. When a section is short and near the bottom, the page can't scroll far enough for that heading to reach the threshold, so the previous section stays active.

### Suggested fix

When the user has scrolled to the bottom of the page, set the last TOC item as active:

```js
const atBottom = window.innerHeight + window.scrollY >= document.body.scrollHeight - 10;
if (atBottom) {
  setActiveId(lastVisibleHeadingId);
  return;
}
```

`docs/src/components/TableOfContents.tsx`, in the `handleScroll` function.

twjeffery:
## Summary

- Fixes the sticky TOC highlighting the wrong section for short sections near the bottom of the page (e.g. "Get started" on the workspace page)
- Adds a bottom-of-page check in `handleScroll`: when `window.innerHeight + window.scrollY >= document.body.scrollHeight - 10`, the last visible heading is set as active instead of relying on the 120px threshold

https://github.com/user-attachments/assets/dc976102-18f6-454a-97ce-a15f1ac5bb8f

Closes #3726

## Test plan

- [ ] Navigate to a docs page with a short final section (e.g. workspace example page)
- [ ] Scroll to the very bottom of the page — the last TOC item should be highlighted
- [ ] Click the last TOC item — it should become active
- [ ] Scroll back up — TOC highlighting should still work normally for all other sections

https://claude.ai/code/session_01Rcc9hcmnjnE8vdXhiz7VMM

### DIFF
--- a/docs/src/components/TableOfContents.tsx
+++ b/docs/src/components/TableOfContents.tsx
@@ -82,12 +82,21 @@ export function TableOfContents({ cssQuery }: TOCProps) {
     // Update active heading on scroll - only check visible headings
     function handleScroll() {
       const headings = document.querySelectorAll<HTMLHeadingElement>(cssQuery);
+      const visibleHeadings = Array.from(headings).filter(isVisible);
+
+      if (visibleHeadings.length === 0) return;
+
+      // When scrolled to the bottom, activate the last visible heading
+      const atBottom =
+        window.innerHeight + window.scrollY >= document.body.scrollHeight - 10;
+      if (atBottom) {
+        const id = visibleHeadings[visibleHeadings.length - 1].getAttribute("id");
+        if (id) setActiveId(id);
+        return;
+      }
+
       let activeHeading: HTMLHeadingElement | null = null;
-
-      for (const heading of headings) {
-        // Only consider visible headings
-        if (!isVisible(heading)) continue;
-
+      for (const heading of visibleHeadings) {
         const rect = heading.getBoundingClientRect();
         if (rect.top < 120) {
           activeHeading = heading;
@@ -95,7 +104,7 @@ export function TableOfContents({ cssQuery }: TOCProps) {
       }
 
       if (activeHeading) {
-        const id = (activeHeading as HTMLHeadingElement).getAttribute("id");
+        const id = activeHeading.getAttribute("id");
         if (id) setActiveId(id);
       }
     }


Spark450 (commented):
I think it is certainly working better but it looks like the selection "bounces" a bit then settles.
https://jam.dev/c/08f2e673-e30e-4d7d-a609-a1a81e62637a

twjeffery:
> I think it is certainly working better but it looks like the selection "bounces" a bit then settles. [jam.dev/c/08f2e673-e30e-4d7d-a609-a1a81e62637a](https://jam.dev/c/08f2e673-e30e-4d7d-a609-a1a81e62637a)

I think that's expected for what it's fixing and doing. This is a fix to make it change (override what's set from the scroll position) to the last item if scroll is at the bottom. It's kind of like, it follows the heading based on the scroll position, and then if the scroll is at the bottom, then set it to bottom as active.

ArakTaiRoth (approved):

docs/src/components/TableOfContents.tsx
```diff
@@ -82,6 +82,22 @@ export function TableOfContents({ cssQuery }: TOCProps) {
...
+        for (const heading of headings) {
+          if (!isVisible(heading)) continue;
+          const id = heading.getAttribute("id");
+          if (id) lastVisibleId = id;
+        }
+        if (lastVisibleId) {
+          setActiveId(lastVisibleId);
+          return;
+        }
+      }
```
chrisolsen, line 96: If you are scrolling to the bottom, is there any reason you can't just set the last of the headings as the current, rather than this logic + calculations?  Is there ever a time that the last heading isn't visible when you scroll to the bottom of the screen?

```
const atBottom = window.innerHeight + window.scrollY >= document.body.scrollHeight - 10;
if (atBottom) {
	const lastHeading = headings[headings.length - 1];
    const lastVisibleId = lastHeading.getAttribute("id");
    setActiveId(lastVisibleId);
    return;
}
```

Spark450 (approved):
If this approach is valid I think the behaviour is acceptable.

docs/src/components/TableOfContents.tsx
twjeffery: Good point. The reason we need to filter is that headings here is all the matching headings in the DOM, not just the active tab. On component pages we have 4 tabs (Properties, Examples, Usage, Accessibility) and the headings from all of them are in the DOM at once. If we just grabbed the last one we'd always end up on a heading from the Accessibility tab no matter which tab the user is on.

You're right that we were doing more work than we needed to though. I refactored it to filter visible headings once at the top and both branches reuse that. It should be simpler now.

chrisolsen (approved):

PR #3736, opened 2026-04-01
chore: nx update

chrisolsen:


ArakTaiRoth (dismissed):

ArakTaiRoth (approved):

libs/angular-components/src/lib/components/base.component.ts
```diff
@@ -119,9 +120,7 @@ export abstract class GoabControlValueAccessor
   @ViewChild("goaComponentRef", { static: false, read: ElementRef })
```
vanessatran-ddi, line 120: At line 53 above, I think we should update the comment, right now we do not use `constructor()` pattern and replace vs `inject()`

vanessatran-ddi (approved):
I just have 1 small comment about js doc comment on base.component.ts, but it is not a blocker.

libs/angular-components/src/lib/components/base.component.ts
chrisolsen: I am not sure what that comment would need to be changed to, or why it would need to be changed.

PR #3758, opened 2026-04-02
fix(#3640): adjust alignment for work side menu badge (both component and text)

### LINKED ISSUE #3640: WorkSideMenuItem: Badge text not vertically centered
## Description

The text inside the badge pill on `GoabxWorkSideMenuItem` is not vertically centered. The text sits slightly high within the pill, making it look off-balance. Visible on badges like "New" (type="success") and "Pending" (type="important").

## Expected

- The badge text should be vertically centered within the pill.
- the icon and the menu item label should be all vertically centered with that badge as well

## Screenshots

Current

Image

Design

Image

mxsoco:
# Before (the change)
image

# After (the change)
Added a `1px` margin to the top of goa-badge to center text more. Additionally, adjusted `.menu-item` items to be center aligned.

image

image

## Steps needed to test
- [ ] Make sure text within the badge is centered vertically (on both primary and secondary items).
- [ ] Ensure the single menu item components are also center aligned.


chrisolsen (approved):

libs/web-components/src/components/work-side-menu/WorkSideMenuItem.svelte
```diff
@@ -393,7 +393,7 @@
       var(--goa-border-radius-m)
     );
     text-decoration: none;
-    align-items: flex-start;
+    align-items: center;
```
bdfranck: The `flex-start` value is used to keep the icon and badge at the top when the label spans multiple lines.

Here's how it looks in Figma:

Image

If we keep `flex-start` and add `space-3xs` to the badge top margin then we get the alignment for one line labels...

Image

...but have top alignment for two line labels.

Image

bdfranck (commented):
Let's adjust the badge with a margin to preserve alignment for multi-line labels.

bdfranck (approved):

- ✅ I see a badge is aligned when the label is one line

Image

- ✅ I see a badge is top aligned when the label is multiple lines

Image

Looks good to me! 👍

PR #3759, opened 2026-04-02
fix(3760): close v1 filter chip when clicking icon

### LINKED ISSUE #3760: FilterChip focusing not closing on icon click
# Info

If you click the close icon directly in a version 1 Filter Chip, it focuses the chip, but doesn't close it, you have to click it a second time to close it.

If you click anywhere else on the chip, it closes it immediately as expected.

This is with **version 1** only.

## Acceptance Criteria:

1. Filter chip should close immediately upon click anywhere within the component

bdfranck:
# Before (the change)

When in Chrome, clicking the v1 filter chip icon doesn't trigger a `_click` event until the second click.

click-issue

# After (the change)

Clicking the v1 filter chip icon triggers a `_click` event.

click-fixed

# The reason

The problem is a race between Chrome's focus event and the `goa-icon` shadow DOM re-render:

| Step | What happens in Chrome |
|------|------------------------|
| 1 | `mousedown` on `goa-icon` → browser focuses the parent `div` |
| 2 | `focus` fires → `_focused = true` → Svelte re-renders |
| 3 | `goa-icon` receives new `theme="filled"` prop → **shadow DOM re-renders under the cursor** |
| 4 | `mouseup` / `click` fire — but Chrome sees the hit-target changed, so **`click` is dropped** |

Firefox is more permissive and fires `click` even when the shadow DOM target shifts mid-gesture, which is why it works there on the first click.

Adding `pointer-events: none` to `.delete-icon` makes the `goa-icon` element transparent to all mouse events, so every click falls straight through to the outer `div.chip` that holds the `on:click={onDelete}` handler. The icon in the v1 chip is purely decorative anyway — the whole chip is meant to be the button.

### DIFF
--- a/libs/web-components/src/components/filter-chip/FilterChip.svelte
+++ b/libs/web-components/src/components/filter-chip/FilterChip.svelte
@@ -186,6 +186,7 @@
     flex-shrink: 0;
     display: flex;
     align-items: center;
+    pointer-events: none; /* Icon itself should not capture pointer events, the chip handles it */
   }
 
   .label-container {
--- a/apps/prs/web/src/app/App.svelte
+++ b/apps/prs/web/src/app/App.svelte
@@ -3,6 +3,7 @@
   import { Router, Route } from "svelte-routing";
   import Issue2333 from "../routes/2333.svelte";
   import Issue3279 from "../routes/3279.svelte";
+  import Issue3760 from "../routes/3760.svelte";
 </script>
 
 <svelte:head>
@@ -12,4 +13,5 @@
 <Router>
   <Route path="/issues/2333" component={Issue2333} />
   <Route path="/issues/3279" component={Issue3279} />
+  <Route path="/issues/3760" component={Issue3760} />
 </Router>
--- a/apps/prs/web/src/routes/3760.svelte
+++ b/apps/prs/web/src/routes/3760.svelte
@@ -1,0 +1,68 @@
+<script lang="ts">
+  let clickCount = 0;
+  let lastEvent = "Waiting for _click from onDelete. Click the close icon once on the v1 filter chip.";
+  let eventLog: string[] = [];
+
+  function handleChipClick() {
+    const timestamp = new Date().toLocaleTimeString();
+    clickCount += 1;
+    lastEvent = `_click received from onDelete at ${timestamp}`;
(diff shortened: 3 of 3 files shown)


chrisolsen (approved):

libs/web-components/src/components/filter-chip/FilterChip.svelte
```diff
@@ -186,6 +186,7 @@
     flex-shrink: 0;
```
ArakTaiRoth, line 186: I'm putting this on this file, but it has nothing to do with this file. The only test we have for FilterChip requires clicking on the Chip itself, it doesn't actively click on the icon. Might be best to add a test for this issue so we don't see this crop up again. Maybe in the browser tests?

bdfranck: @ArakTaiRoth I tried adding this to the web component tests...

```typescript
it("should emit _click after focus causes icon re-render (Chrome v1 regression)", async () => {
     const { container } = render(GoAFilterChip, {
       testid: "chip",
       content: "Status: Active",
       version: "1",
     });
     const chip = container.querySelector(
       "[data-testid='chip']",
     ) as HTMLElement;
     const deleteIcon = container.querySelector(".delete-icon") as HTMLElement;
     const onClick = vi.fn();

     chip.addEventListener("_click", onClick);

     // Simulate the Chrome race: focus fires first → _focused = true → goa-icon re-renders with theme="filled"
     await fireEvent.focus(chip);
     // Click should still reach the chip handler despite the icon re-rendering under the cursor
     await fireEvent.click(deleteIcon);

     expect(onClick).toHaveBeenCalledOnce();
   });
``` 

...but I couldn't get it to fail. Is there a way to easily create a browser test that uses the v1 component. Now the React wrapper is hardcoded to the v2 component. 🤔

ArakTaiRoth: @bdfranck Totally fair. We might not be able to test an exclusive v1 issue, and to be fair it's not an issue with v2 and never will be because v2 is built different. So maybe just ignore my initial request.

ArakTaiRoth (approved):

ArakTaiRoth:
@chrisolsen This is ready to be merged

PR #3766, opened 2026-04-06
fix(#3625): Accordion Refinements

### LINKED ISSUE #3625: Accordion refinement
## Overview

## Acceptance Criteria:
- Large heading Accordion should be 64px tall
- Small heading Accordion should be 48px tall
- After size adjustments above: 
  - verify that the small heading text is vertically centred for both primary and secondary text
  - Large heading secondary text is vertically centered
- Ensure other slotted elements are still vertically centered after update
- Add motion tokens to animate the following: 
  - "hover" - Background transform from white to grey use --goa-motionDuration-short-2
  - "open" - Show accordion content area use --goa-motionCurve-expressive and --goa-motionDuration-short-3
  - "close" - Hide accordion content area, use --goa-motionCurve-expressive-exit and --goa-motionDuration-short-3

willcodeforcoffee:
# Before (the change)

Accordion is all unrefined and gross. 🤢

image
	
# After (the change)

Fixed all discrepancies from issue #3625.

- Large heading Accordion is 64px tall
- Small heading Accordion is 48px tall
- After size adjustments above:
	- the small heading text is vertically centered for
		- primary text 
		- secondary text	
	- large heading secondary text is vertically centered
- Ensure other slotted elements are still vertically centered after update
- Add motion tokens to animate the following:
	- "hover" - Background transform from white to grey use `--goa-motion-duration-short-2`
	- "open" - Show accordion content area use `--goa-motion-curve-expressive` and `--goa-motionDuration-short-3`
	- "close" - Hide accordion content area, use `--goa-motion-curve-expressive-exit` and `--goa-motionDuration-short-3`

image

## Steps needed to test

Make sure you check out [the `eric/3625-accordion-refinements` branch](https://github.com/GovAlta/design-tokens/pull/145) from design-tokens and update `package.json` `devDependencies` to use it:

```json
"@abgov/design-tokens": "file://../design-tokens",
"@abgov/design-tokens-v2": "npm:@abgov/design-tokens@2.2.1",
```

Run `npm i` after making the change.

This PR will close [Issue #3625](https://github.com/GovAlta/ui-components/issues/3625)

### DIFF
--- a/libs/web-components/src/components/accordion/Accordion.svelte
+++ b/libs/web-components/src/components/accordion/Accordion.svelte
@@ -10,8 +10,11 @@
     validateRequired,
     generateRandomId,
     ensureSlotExists,
+    dispatch,
+    parseCssTimeToMilliseconds,
   } from "../../common/utils";
   import type { Spacing } from "../../common/styling";
+  import { Once } from "@abgov/ui-components-common";
 
   // Validators
 
@@ -20,9 +23,15 @@
     ["small", "medium"],
   );
 
+  const [IconPositions, validateIconPosition] = typeValidator(
+    "Accordion icon position",
+    ["left", "right"],
+  );
+
   // Types
 
   type HeadingSize = (typeof HeadingSizes)[number];
+  type IconPosition = (typeof IconPositions)[number];
 
   // Props
 
@@ -49,16 +58,27 @@
   /** Left margin. */
   export let ml: Spacing = null;
   /** Sets the position of the expand/collapse icon. */
-  export let iconposition: "left" | "right" = "left";
+  export let iconposition: IconPosition = "left";
 
   // Private
 
   let _hovering: boolean = false;
   let _titleEl: HTMLElement;
   let _rootEl: HTMLElement;
+  let _detailsEl: HTMLDetailsElement;
+  let _summaryEl: HTMLElement;
   let _slotEl: HTMLElement;
   let _headingSlotChildren: Element[] = [];
   let _accordionId: string = "";
+
+  // Store the animation object (so we can cancel it if needed)
+  let _animation: Animation | null = null;
+  // Store if the element is closing
+  let _isClosing = false;
+  // Store if the element is expanding
+  let _isExpanding = false;
+  // Show error messages only once
+  let _once: Once = new Once();
 
   // Reactive
 
@@ -69,6 +89,7 @@
   onMount(() => {
     validateRequired("GoAAccordion", { heading });
     validateHeadingSize(headingsize);
+    validateIconPosition(iconposition);
     ensureSlotExists(_slotEl);
 
     _headingSlotChildren = getHeadingChildren();
@@ -81,22 +102,172 @@
       if (slot) {
         return slot.assignedElements();
       } else {
-        // @ts-expect-error
         return [..._titleEl.children] as Element[]; // unit tests
       }
     }
     return [];
   }
 
-  function dispatchClickEvent(open?: boolean) {
+  function dispatchChangeEvent(accordionIsOpen: boolean) {
     if (!_rootEl) return;
 
-    _rootEl.dispatchEvent(
-      new CustomEvent("_change", {
-        composed: true,
-        bubbles: true,
-        detail: { open: open },
-      }),
+    dispatch(
+      _rootEl,
+      "_change",
+      {
+        open: accordionIsOpen,
+      },
+      { bubbles: true },
+    );
+  }
+
+  function onAccordionToggle(target: Event) {
+    // Stop default behaviour from the browser
+    target.preventDefault();
+    // Add an overflow on the <details> to avoid content overflowing
+    _detailsEl.style.overflow = "hidden";
+
+    // Check if the element is being closed or is already closed
+    if (_isClosing || !_detailsEl.open) {
+      openAccordion();
+      // Check if the element is being openned or is already open
+    } else if (_isExpanding || _detailsEl.open) {
+      shrink();
+    }
+  }
+
+  function shrink() {
+    // Set the element as "being closed"
+    _isClosing = true;
+
+    // Store the current height of the element
+    const startHeight = `${_detailsEl.offsetHeight}px`;
+    // Calculate the height of the summary
+    const endHeight = `${_summaryEl.offsetHeight}px`;
+
+    // If there is already an animation running
+    if (_animation) {
+      // Cancel the current animation
+      _animation.cancel();
+    }
+
+    // Start a WAAPI animation
+    const duration = getAnimationDuration();
+    const easing = getAnimationEasingExit();
+    _animation = _detailsEl.animate(
+      {
+        // Set the keyframes from the startHeight to endHeight
+        height: [startHeight, endHeight],
+      },
+      {
+        duration,
+        easing,
+      },
+    );
+
+    // When the animation is complete, call onAnimationFinish()
+    _animation.onfinish = () => onAnimationFinish(false);
+    // If the animation is cancelled, isClosing variable is set to false
+    _animation.oncancel = () => (_isClosing = false);
+  }
+
+  function openAccordion() {
+    // Apply a fixed height on the element
+    _detailsEl.style.height = `${_detailsEl.offsetHeight}px`;
+    // Force the [open] attribute on the details element
+    _detailsEl.open = true;
+    // Wait for the next frame to call the expand function
+    window.requestAnimationFrame(() => expand());
+  }
+
+  function expand() {
+    // Set the element as "being expanding"
+    _isExpanding = true;
+    // Get the current fixed height of the element
+    const startHeight = `${_detailsEl.offsetHeight}px`;
+    // Calculate the open height of the element (summary height + content height)
+    const endHeight = `${_summaryEl.offsetHeight + _slotEl.offsetHeight}px`;
+
+    // If there is already an animation running
+    if (_animation) {
+      // Cancel the current animation
+      _animation.cancel();
+    }
+
+    // Start a WAAPI animation
+    const duration = getAnimationDuration();
+    const easing = getAnimationEasingReveal();
+    _animation = _detailsEl.animate(
+      {
+        // Set the keyframes from the startHeight to endHeight
+        height: [startHeight, endHeight],
+      },
+      {
+        duration,
+        easing,
+      },
+    );
+    // When the animation is complete, call onAnimationFinish()
+    _animation.onfinish = () => onAnimationFinish(true);
+    // If the animation is cancelled, isExpanding variable is set to false
+    _animation.oncancel = () => (_isExpanding = false);
+
+    open = "true";
+  }
+
+  function onAnimationFinish(isAccordionOpen: boolean) {
+    // Set the open attribute based on the parameter
+    _detailsEl.open = isAccordionOpen;
+    // Clear the stored animation
+    _animation = null;
+    // Reset isClosing & isExpanding
+    _isClosing = false;
+    _isExpanding = false;
+    // Remove the overflow hidden and the fixed height
+    _detailsEl.style.height = _detailsEl.style.overflow = "";
+
+    dispatchChangeEvent(isAccordionOpen);
+
+    open = isAccordionOpen ? "true" : "false";
+  }
+
+  // Because the animation is done using JavaScript we can't use CSS variables
+  // in the animation, this helps to work around that.
+  function getCssVariableValue(variableName: string): string | null {
+    const value = getComputedStyle(_rootEl)
+      .getPropertyValue(variableName)
+      .trim();
+
+    if (!value || value === "") {
+      // Show a warning only once
+      _once.do(variableName, () =>
+        console.warn(
+          `CSS variable ${variableName} is not defined. Please make sure you're using the correct version of @abgov/design-tokens`,
+        ),
+      );
+      return null;
+    }
+    return value;
+  }
+
+  function getAnimationDuration(): number {
+    const durationValueStr =
+      getCssVariableValue("--goa-motion-duration-short-3") ?? "100ms";
+
+    return parseCssTimeToMilliseconds(durationValueStr, 100);
+  }
+
+  function getAnimationEasingReveal(): string {
+    return (
+      getCssVariableValue("--goa-motion-curve-expressive-reveal") ??
+      "cubic-bezier(0.7, 0, 0.25, 1)"
+    );
+  }
+
+  function getAnimationEasingExit(): string {
+    return (
+      getCssVariableValue("--goa-motion-curve-expressive-exit") ??
+      "cubic-bezier(0.42, 0, 1, 1)"
     );
   }
 </script>
@@ -111,8 +282,13 @@
   class="goa-accordion"
   bind:this={_rootEl}
   data-testid={testid}
+  {id}
 >
-<details open={isOpen} on:toggle={({ target }) => { open = `${target?.open}`; dispatchClickEvent(target?.open); }}>
+  <details
+    bind:this={_detailsEl}
+    open={isOpen}
+    data-testid={`${testid}-details`}
+  >
     <!-- svelte-ignore a11y-no-static-element-interactions -->
     <summary
       class={`container-${headingsize}`}
@@ -121,10 +297,12 @@
       on:focus={() => (_hovering = false)}
       on:blur={() => (_hovering = false)}
       aria-controls={`${_accordionId}-content`}
-      aria-expanded={open === "true"}
+      aria-expanded={isOpen}
       class:iconRight={iconposition === "right"}
+      bind:this={_summaryEl}
+      on:click={onAccordionToggle}
+      data-testid={`${testid}-summary`}
     >
-
       {#if iconposition === "left"}
         <goa-icon
           type="chevron-forward"
@@ -145,6 +323,7 @@
         <div
           class="heading-content"
           class:heading-content-top={_headingSlotChildren.length}
+          on:click|stopPropagation
         >
           <slot name="headingcontent" />
         </div>
@@ -207,6 +386,8 @@
 
     /* safari hack (see below) */
     position: relative;
+    transition: var(--goa-motion-duration-short-2) background-color
+      var(--goa-motion-curve-expressive);
   }
 
   summary.iconRight {
@@ -242,7 +423,7 @@
   }
 
   summary goa-icon {
-    padding: 0.125rem 1rem;
+    padding: 0.25rem 1rem;
   }
 
   .title {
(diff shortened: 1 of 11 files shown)


libs/web-components/src/components/accordion/Accordion.svelte
```diff
@@ -222,6 +222,8 @@
     background-color: var(--goa-accordion-color-bg-heading-hover);
     border: var(--goa-accordion-border-hover, var(--goa-accordion-border));
     color: var(--goa-accordion-color-heading-hover);
+    transition: var(--goa-motion-duration-medium-1) background-color
```
Spark450: Lets use --goa-motionDuration-short-2 to align more closely with the "short-2" we are using for other similar elements.

libs/web-components/src/components/accordion/Accordion.svelte
```diff
@@ -217,6 +359,8 @@
     background-color: var(--goa-accordion-color-bg-heading-hover);
     border: var(--goa-accordion-border-hover, var(--goa-accordion-border));
     color: var(--goa-accordion-color-heading-hover);
+    transition: var(--goa-motion-duration-short-2) background-color
```
bdfranck: Add the `transition` to the base `summary` selector on line 338 so the transitions both works on hover and blur.

libs/web-components/src/components/accordion/Accordion.svelte
```diff
@@ -242,43 +386,64 @@
...
 
-  .title span {
-    padding-bottom: var(--goa-space-2xs, 0);
+  .heading {
+    font: var(--goa-accordion-heading);
+    padding: 0.125rem 1rem 0 0;
   }
 
-  .heading {
+  .heading-small {
```
bdfranck, line 439: Image

I'm still seeing the small heading is `50px` instead of `48px`. Is the dependency because of the borders?

libs/web-components/src/components/accordion/Accordion.svelte
```diff
@@ -242,43 +386,64 @@
...
+    padding-top: 0.25rem;
+    line-height: 1.75rem;
     font: var(--goa-accordion-heading);
-    padding-right: 1rem;
+  }
+
+  .heading-medium {
+    line-height: 2rem;
```
bdfranck, line 446: I'm still seeing the medium heading is `66px` instead of `64px`. Is the dependency because of the borders?

Image

libs/web-components/src/components/accordion/Accordion.svelte
```diff
@@ -99,6 +108,137 @@
...
   }
+
+  // This is needed to avoid the click event on the heading content to trigger
+  // the summary click event and toggle the accordion.
+  // Previously we used on:toggle, but that is not supported by
+  function clickSoak(event: MouseEvent) {
+    event.stopPropagation();
+  }
+
+  function onAccordionToggle(target: EventHandler) {
```
bdfranck: When I slowed down the animation in the [Chrome Dev Tools Animation panel](https://developer.chrome.com/docs/devtools/css/animations/), I noticed that it wasn't including the component's bottom border. Is there an easy way to address this? 🤔 

Image

bdfranck (commented):
I tested the changes in the latest browsers...

### Safari (Latest)
- ✅ The sizes look good
- ✅ The animation works as expected

### Chrome (Latest)
- ✅ The sizes look good
- ✅ The animation works as expected

### Chrome (Latest)
- ✅ The sizes look good
- ✅ The animation works as expected

Looks good! I left a couple minor comments.

libs/web-components/src/components/accordion/Accordion.svelte
willcodeforcoffee: 🤔 Hmm, you're right! I must not have included the borders in my calculations.

libs/web-components/src/components/accordion/Accordion.svelte
willcodeforcoffee: @bdfranck 🤔 🤔 🤔  I believe this is a consequence of animating it using height attributes..? Do you have a suggestion? Maybe if I animate the height of an internal `div` instead? I can try that and see what I can do...

libs/web-components/src/components/accordion/Accordion.svelte
willcodeforcoffee: @bdfranck I was able to fix this, you just need to get the latest of `design-tokens` first.

libs/web-components/src/components/accordion/Accordion.svelte
```diff
@@ -99,6 +108,137 @@
...
+    _detailsEl.style.height = _detailsEl.style.overflow = "";
+  }
+
+  // Because the animation is done using JavaScript we can't use CSS variables
+  // in the animtion, this helps to work around that.
+  function getCssVariableValue(variableName: string): string {
+    return getComputedStyle(_rootEl).getPropertyValue(variableName).trim();
+  }
+
+  function getAnimationDuration(): number {
```
vanessatran-ddi, line 253: I think the value we are using here is wrong

Image

Also why don't we use CSS instead of a function to get a token value? 
For example: 
https://css-tricks.com/css-grid-can-do-auto-height-transitions/

```
.content-wrapper {
    display: grid;
    grid-template-rows: 0fr;
    transition: grid-template-rows
      var(--goa-motion-duration-short-3)
      var(--goa-motion-curve-expressive);
  }
  details[open] .content-wrapper {
    grid-template-rows: 1fr;
  }
  .content-wrapper > .content {
    overflow: hidden;
  }
```

libs/web-components/src/components/accordion/Accordion.svelte
```diff
@@ -99,6 +108,137 @@
...
+    const durationStr = getCssVariableValue("--goa-motion-duration-short-2");
+    return parseFloat(durationStr) || 100;
+  }
+
+  function getAnimationEasingReveal(): string {
```
vanessatran-ddi, line 260: Also I think we should have a fallback value if developers haven't installed the latest design-tokens on their app

libs/web-components/src/components/accordion/Accordion.svelte
vanessatran-ddi: EventHandler type does not exist, it is not a DOM type built-in. Suggestion: `function onAccordionToggle(event: MouseEvent)` instead, rename the variable `target` because it makes me feel we receive an `event.target`, but it is not a case here

libs/web-components/src/components/accordion/Accordion.svelte
```diff
@@ -121,10 +261,11 @@
       on:focus={() => (_hovering = false)}
       on:blur={() => (_hovering = false)}
       aria-controls={`${_accordionId}-content`}
-      aria-expanded={open === "true"}
+      aria-expanded={isOpen}
```
vanessatran-ddi, line 300: I do not see we update `isOpen` or `open` when we toggle the accordion, so in this case aria-expanded will not be updated when we toggle the accordion

libs/web-components/src/components/accordion/Accordion.svelte
```diff
@@ -99,6 +108,137 @@
...
+    _detailsEl.style.overflow = "hidden";
+    // Check if the element is being closed or is already closed
+    if (_isClosing || !_detailsEl.open) {
+      openAccordion();
+      // Check if the element is being openned or is already open
+    } else if (_isExpanding || _detailsEl.open) {
+      shrink();
+    }
+
+    dispatchClickEvent(_detailsEl?.open);
```
vanessatran-ddi: I did a quick test like this: 

Image

I see 2 issues:

Image

1. I click close and open, the property `open` is not synced
2. There is an error if we don't have a token set up, I think we need to have a fallback value to prevent this animation issue, otherwise if developers don't update their `design-tokens` but update our library, it may break. (https://github.com/GovAlta/goa-workspace-playground/blob/main/src/index.css#L2 How they set up now)

libs/web-components/src/components/accordion/Accordion.svelte
```diff
@@ -294,6 +459,11 @@
     margin-bottom: 0 !important;
   }
 
+  /* details[open] .content {
```
vanessatran-ddi: We can clean come comments if it is no more useful.

libs/web-components/src/components/accordion/Accordion.svelte
willcodeforcoffee: After team discussion, we'll leave this animation as is.

libs/web-components/src/components/accordion/Accordion.svelte
willcodeforcoffee: 😬 I feel bad about leaving these in, but they're cleaned up now!

libs/web-components/src/components/accordion/Accordion.svelte
willcodeforcoffee: Good suggestion. I added a default along with a warning message. 👌 
I also made sure the warning shows only once for each missing token.

libs/web-components/src/components/accordion/Accordion.svelte
willcodeforcoffee: Good catch. I'm not sure why I thought it was EventHandler... I just used `Event` for the click, same as `Button` uses. We didn't need anything from `MouseEvent` or other Event types. 🤷‍♂️

libs/web-components/src/components/accordion/Accordion.svelte
willcodeforcoffee: 🤔 That is a neat trick... I tried a few different animation types for details accordions that didn't work, that is why I ended up with the JavaScript version.

willcodeforcoffee:
Addressed feedback:
- moved the transition to the base summary
- fixed the sizes of the headings
- fixed animation speed to `--goa-motion-duration-short-3` instead of `short-2`
- added warning message and fallback so the accordion still works if the wrong version of `design-tokens` is included
- changed `EventHandler` type to `Event`
- `open` and `aria-expanded` attributes work correctly
- removed commented CSS

@bdfranck @vanessatran-ddi ready for a re-review! 🙏

libs/web-components/src/components/accordion/Accordion.svelte
vanessatran-ddi: The Grid trick does not animate `<details>` itself. It is worth a quick try because using css we will consume the token directly via `var()`, and keep us aligned vs how other component in the repo handles transitions (ex Drawer.svelte). If one day we mistakenly assign the token value instead of 100ms to 0.1s for example then the way we make it here will assume 0.1ms ` _animation = _detailsEl.animate(...)`.

libs/web-components/src/components/accordion/Accordion.svelte
vanessatran-ddi: One concern: fall back value are generic css keywords like `ease-in`, `ease-out` and `0`, not the actual token values. so when tokens are missing, the animation runs with a different curve than designed.

libs/web-components/src/components/accordion/Accordion.svelte
vanessatran-ddi: I think we do not need this separate function. You can do like this `<div class="heading-content" on:click|stopPropagation>`

libs/web-components/src/components/accordion/Accordion.svelte
vanessatran-ddi: I tested this one, `isOpen` updated internally, but `open` not updated

https://github.com/user-attachments/assets/d0b7fc65-cb9c-4b0d-b59d-56dba3fc3c9b

So for example on the video, if I close the accordion by not clicking the button but clicking on the accordion heading, the `onChange` dispatched vs the wrong `open` state, so the button label doesn't change or updated.

libs/web-components/src/components/accordion/Accordion.svelte
vanessatran-ddi: Almost - the commented block at line 494 is still there

vanessatran-ddi (commented):
I leave some comment and push 1 example for React related to `open`

libs/web-components/src/components/accordion/Accordion.svelte
willcodeforcoffee: @vanessatran-ddi I totally agree. I 100% preferred a CSS solution but I couldn't find one that worked in all browsers. I played with the [CodePen in the article you linked](https://codepen.io/chriscoyier/pen/qBXoEMV) but I couldn't get it to work, either. I tried a _lot_ of other solutions too, but none worked. 😞 
The problem seems to be that the browser treats a `details`/`summary` control differently.

libs/web-components/src/components/accordion/Accordion.svelte
willcodeforcoffee: Oh that is awesome! I didn't know about this feature! Thank you @vanessatran-ddi!

libs/web-components/src/components/accordion/Accordion.svelte
willcodeforcoffee: It was a pretty strange bug, but basically I'd put the `dispatch` method in the wrong spot, and before the animations had completed.

I also changed it to use the correct `dispatch` method from utils.

libs/web-components/src/components/accordion/Accordion.svelte
willcodeforcoffee: @vanessatran-ddi I thought hard about it...  At first I figured it should fall back to something wrong, but still functional. That way the developers will know there is something they need to fix.
But on the other hand there is no loss in just using the correct values while still showing a warning the library is the wrong version.
As long as we don't feel like we have to keep the fallback values up-to-date with design-tokens if they get changed. I don't want to create that kind of extra maintenance headache.

bdfranck (approved):
I looked at the latest changes with the tokens from https://github.com/GovAlta/design-tokens/pull/145...

- ✅ The small heading height is `48px` and the medium heading height is `64px`
- ✅ I see `aria-expanded` is `true` when the accordion is open
- ✅ The animation works as expected

Image

Looks good to me! 👍

libs/web-components/src/components/accordion/Accordion.svelte
vanessatran-ddi: Agree. If that is the case, we can add a function to parseDurationMs, for example: if we make a change on `design-tokens` and we set `--goa-motion-duration-short-3: 0.3s` instead of `300ms`, then your function `getAnimationDuration` will become 0.3ms. I will recommend to detect if we `endsWith("s")` or `endsWith("ms")` to returns correct value?

libs/web-components/src/components/accordion/Accordion.svelte
```diff
@@ -81,22 +101,171 @@
...
+        duration,
+        easing,
+      },
+    );
+    // When the animation is complete, call onAnimationFinish()
+    _animation.onfinish = () => onAnimationFinish(true);
+    // If the animation is cancelled, isExpanding variable is set to false
+    _animation.oncancel = () => (_isExpanding = false);
+
+    isOpen = true;
```
vanessatran-ddi: I think here should be `open="true"` instead of `isOpen=true` because at line 84, we have this hook. We should have 1 single of truth only, rely on line 84 will handle the action when user clicks a button to expand/collapse the accordion

libs/web-components/src/components/accordion/Accordion.svelte
```diff
@@ -81,22 +101,171 @@
...
+        easing,
+      },
+    );
+
+    // When the animation is complete, call onAnimationFinish()
+    _animation.onfinish = () => onAnimationFinish(false);
+    // If the animation is cancelled, isClosing variable is set to false
+    _animation.oncancel = () => (_isClosing = false);
+
+    isOpen = false;
```
vanessatran-ddi: Update to `open="false"` instead of `isOpen` since we have line 84 `isOpen=tooBoolean(open);`

libs/web-components/src/components/accordion/Accordion.svelte
willcodeforcoffee: The only downside to this is that `open` is a string instead of a boolean.

libs/web-components/src/components/accordion/Accordion.svelte
willcodeforcoffee: Claude came up with a cool regex that checks for 's' or 'ms' at the end or else does a fallback.
I put it into the `utils` file so that I could write specs for it, too.

willcodeforcoffee:
Finished addressing comments:

- `open` instead of `isOpen`
- refactored the CSS time parsing

Thanks again for the thoughtful feedback @vanessatran-ddi .

vanessatran-ddi (approved):
Looks good!

willcodeforcoffee:
@ArakTaiRoth This is ready to merge! 🙏

libs/web-components/src/components/accordion/Accordion.svelte
```diff
@@ -81,22 +102,172 @@
...
+        easing,
+      },
+    );
+
+    // When the animation is complete, call onAnimationFinish()
+    _animation.onfinish = () => onAnimationFinish(false);
+    // If the animation is cancelled, isClosing variable is set to false
+    _animation.oncancel = () => (_isClosing = false);
+
+    open = "false";
```
ArakTaiRoth: Because we're setting `open` to `false` here, before `onAnimationFinish` actually finishes. This causes the content to disappear before the Accordion actually closes.

[jam.dev](https://jam.dev/c/e41f6c15-ceb2-4f28-b331-d28f6ccd772a)

libs/web-components/src/components/accordion/Accordion.svelte
```diff
@@ -81,22 +102,170 @@
...
+    _animation.onfinish = () => onAnimationFinish(true);
+    // If the animation is cancelled, isExpanding variable is set to false
+    _animation.oncancel = () => (_isExpanding = false);
+
+    open = "true";
+  }
+
+  function onAnimationFinish(isAccordionOpen: boolean) {
+    // Set the open attribute based on the parameter
+    _detailsEl.open = isAccordionOpen;
```
ArakTaiRoth, line 220: I'm really sorry about this. But your previous fix for my last issue introduced a new issue. Now because `open` isn't being set to false anymore in `shrink()`, `aria-expanded` isn't being updated properly, so now when you close an accordion, `aria-expanded` still reads as `true`, you can test this with any accordion in your PR example.

I would suggest maybe moving the `open` property handling to `onAnimationFinish` instead of handling it in `shrink()` or `expand()`, and just changing what the value is everytime it's run, `open = isAccordionOpen ? "true" : "false";`.

You have this line here that handles `_detailsEl.open`, but that doesn't modify `open`, which is what's deriving `isOpen`, which is what's used for `aria-expanded`.

libs/web-components/src/components/accordion/Accordion.svelte
willcodeforcoffee: @ArakTaiRoth Good catch, I wasn't looking close enough at the animation afterwards!
This is now fixed. I didn't see any issues or regressions.

libs/web-components/src/components/accordion/Accordion.svelte
willcodeforcoffee: @ArakTaiRoth another good catch! 🤦 Sorry I missed that one, too. 😞

libs/web-components/src/components/accordion/Accordion.svelte
```diff
@@ -81,22 +102,173 @@
...
+    // When the animation is complete, call onAnimationFinish()
+    _animation.onfinish = () => onAnimationFinish(true);
+    // If the animation is cancelled, isExpanding variable is set to false
+    _animation.oncancel = () => (_isExpanding = false);
+
+    open = "true";
+  }
+
+  function onAnimationFinish(isAccordionOpen: boolean) {
+    console.log("onfinish");
```
ArakTaiRoth: Remove this

willcodeforcoffee: falling down

PR #3768, opened 2026-04-07
fix(#3743): tertiary background colour fix on mobile

### LINKED ISSUE #3743: Button: tertiary background colour should not change on mobile
Tertiary buttons show a grey background on mobile. They should not change background colour on mobile.

### Context

Previously tertiary buttons had no outline, so on mobile when they were stacked full width they visually floated. The grey background was added specifically for mobile to remedy that.

With the updated tertiary button designs (which now have an outline) and those buttons not always going full width on mobile, the mobile-specific background colour is no longer needed. It should be removed so tertiary buttons render consistently across breakpoints.

## Acceptance Criteria

1. Tertiary buttons should not change background colour on mobile

mxsoco:
# Before (the change)
The background colour still exists on mobile for the tertiary button. Same with the inverse, but only on mobile.

image

# After (the change)
Removed 4 lines that bring in the background. Including tokens:
- `--goa-button-tertiary-color-bg-mobile`
- `--goa-button-tertiary-color-bg`

image

## Steps needed to test

- [ ] Ensure background is transparent for tertiary buttons on mobile.
- [ ] Check the same for the inverse variant.

### DIFF
--- a/libs/web-components/src/components/button/Button.svelte
+++ b/libs/web-components/src/components/button/Button.svelte
@@ -218,13 +218,13 @@
       background-color: var(--goa-button-tertiary-color-bg-mobile) !important;
     }
     button.tertiary:hover {
-      background-color: var(--goa-color-greyscale-200) !important;
+      background-color: var(--goa-button-tertiary-color-bg-mobile) !important;
     }
     button.tertiary.inverse {
       background-color: var(--goa-button-tertiary-color-bg) !important;
     }
     button.tertiary.inverse:hover {
-      background-color: var(--goa-color-greyscale-200) !important;
+      background-color: var(--goa-button-tertiary-color-bg-mobile) !important;
     }
   }
 
--- a/libs/web-components/src/components/button/Button.svelte
+++ b/libs/web-components/src/components/button/Button.svelte
@@ -214,6 +214,18 @@
       width: 100%;
       display: flex;
     }
+    button.tertiary {
+      background-color: var(--goa-button-tertiary-color-bg-mobile) !important;
+    }
+    button.tertiary:hover {
+      background-color: var(--goa-color-greyscale-200) !important;
+    }
+    button.tertiary.inverse {
+      background-color: var(--goa-button-tertiary-color-bg) !important;
+    }
+    button.tertiary.inverse:hover {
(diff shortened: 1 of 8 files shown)


chrisolsen:
If this is as per design then it looks good. @twjeffery ?

vanessatran-ddi (approved):
Not sure if we should fix this on this PR, or maybe we can do in a separate one: When I focus on the button I see this, it is difficult to read the text and I think the text color should be white?
cc @twjeffery 
Image

libs/web-components/src/components/button/Button.svelte
```diff
@@ -214,18 +214,6 @@
       width: 100%;
       display: flex;
     }
-    button.tertiary {
```
bdfranck, line 217: We should update the token value instead of removing it. Removing it could lead to frustrated teams who update the `-goa-button-tertiary-color-bg-mobile` value and find that nothing happens.

libs/web-components/src/components/button/Button.svelte
```diff
@@ -214,18 +214,6 @@
...
-      background-color: var(--goa-button-tertiary-color-bg-mobile) !important;
-    }
-    button.tertiary:hover {
-      background-color: var(--goa-color-greyscale-200) !important;
```
bdfranck, line 221: Let's add a new token with the value we want instead of removing the CSS here.

bdfranck (commented):
We can update and add tokens instead of removing them to prevent future issues.

libs/web-components/src/components/button/Button.svelte
mxsoco: If I update `--goa-button-tertiary-color-bg-mobile` to just be transparent, could that just replace ` --goa-color-greyscale-200` here?

bdfranck (approved):
I looked at the latest changes with tokens from (https://github.com/GovAlta/design-tokens/pull/153/changes)...

https://github.com/user-attachments/assets/8a8c38e4-c0f1-4174-83d6-da97d24baf9f

I see the tertiary button background is transparent on mobile. Looks good to me! 👍

chrisolsen (approved):

PR #3786, opened 2026-04-08
fix(#3735): fix tooltip position in Work Side Menu and add to toggle button

### LINKED ISSUE #3735: Search and Menu-expand menu items are missing a tooltip when work-side-menu is collapsed
### Bug
The Search item and the menu-expand item in the work side menu has no tooltip on hover when the work side menu is collapsed. All other menu items show a tooltip on hover when collapsed, but Search and the menu-expand item does not. 

### Acceptance Criteria
They should show "Search" and "Expand" as the tooltips on hover.

### Currently
https://v2.design.alberta.ca/
Image
Image
Image

bdfranck:
This PR adds the following fixes to the Work Side Menu:

- Adds tooltips to the Toggle button and the Search button in Docs for better usability.
- Ensures consistent tooltip alignment, both horizontally and vertically, even when the menu is offset.
- Adds a component token to allow dynamic setting of the menu height.

tooltip-fixed

### DIFF
--- a/docs/src/components/nav/MenuSecondaryContent.tsx
+++ b/docs/src/components/nav/MenuSecondaryContent.tsx
@@ -8,10 +8,7 @@
  */
 
 import { useCallback, useState, useEffect } from "react";
-import {
-  GoabSpacer,
-  GoabWorkSideMenuItem,
-} from "@abgov/react-components";
+import { GoabIcon, GoabSpacer, GoabWorkSideMenuItem } from "@abgov/react-components";
 
 import "./menu-secondary.css";
 
@@ -33,15 +30,51 @@ export function MenuSecondaryContent({ isOpen }: MenuSecondaryContentProps) {
     window.dispatchEvent(new CustomEvent("goa-search-open"));
   }, []);
 
+  const showSearchTooltip = useCallback(
+    (
+      event: React.MouseEvent<HTMLButtonElement> | React.FocusEvent<HTMLButtonElement>,
+    ) => {
+      event.currentTarget.dispatchEvent(
+        new CustomEvent("_hoverItem", {
+          bubbles: true,
+          composed: true,
+          detail: {
+            label: "Search",
+            el: event.currentTarget,
+          },
+        }),
+      );
+    },
+    [],
+  );
+
+  const hideSearchTooltip = useCallback(
+    (
+      event: React.MouseEvent<HTMLButtonElement> | React.FocusEvent<HTMLButtonElement>,
+    ) => {
+      event.currentTarget.dispatchEvent(
+        new CustomEvent("_blurItem", {
+          bubbles: true,
+          composed: true,
+        }),
+      );
+    },
+    [],
+  );
+
   return (
     <>
       <button
         className="search-menu-button"
         onClick={openSearch}
+        onMouseEnter={showSearchTooltip}
+        onMouseLeave={hideSearchTooltip}
+        onFocus={showSearchTooltip}
+        onBlur={hideSearchTooltip}
         type="button"
         aria-label="Search"
       >
-        <goa-icon type="search" size="small" />
+        <GoabIcon type="search" size="small" />
         {isOpen && (
           <>
             <span className="search-menu-label">Search</span>
--- a/libs/web-components/src/components/work-side-menu/WorkSideMenu.svelte
+++ b/libs/web-components/src/components/work-side-menu/WorkSideMenu.svelte
@@ -55,8 +55,10 @@
 
   let _menuEl: HTMLElement;
   let _menuLinks: HTMLElement[] = [];
+  let _containerEl: HTMLElement;
   let _rootEl: HTMLElement;
   let _scrollEl: HTMLElement;
+  let _toggleButtonEl: HTMLElement;
   let _tooltipEl: HTMLElement;
   let _tooltipLabel: string = "";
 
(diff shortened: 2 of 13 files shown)


libs/web-components/src/components/work-side-menu/WorkSideMenu.svelte
```diff
@@ -225,12 +244,6 @@
   // Menu-scoped keyboard navigation (only fires when focus is inside menu)
   function handleMenuKeyDown(e: KeyboardEvent) {
     switch (e?.key) {
-      case "ArrowDown":
```
bdfranck, line 265: I forgot to remove this from a previous PR. It was causing errors so I removed it here.

twjeffery (approved):
Looks great to me

Image

libs/web-components/src/components/work-side-menu/WorkSideMenu.svelte
```diff
@@ -159,18 +161,35 @@
...
 
   function hideTooltip() {
+    clearTimeout(_mouseEnterTimeoutId);
     clearTimeout(_mouseLeaveTimeoutId);
     _mouseLeaveTimeoutId = setTimeout(() => {
       _showTooltip = false;
     }, 300);
   }
 
+  function showToggleTooltip() {
```
vanessatran-ddi: We should rename this `handleToggleHover()` instead of `showToggleTooltip()` as it may cause the confusion between `showTooltip` and `showToggleTooltip`. 

Also we can shorten like this: 
```
function handleToggleHover() {
if (open) return;

setTooltipPos("Expand menu", _toggleButtonEl);
showTooltip();
}
```
so under `button` below we can do like this: 
```
<button ...
on:mouseenter={handleToggleHover}
on:mouseleave={hideTooltip}
on:focus={handleToggleHover}
on:blur={hideTooltip}
```

libs/web-components/src/components/work-side-menu/WorkSideMenu.svelte
```diff
@@ -489,7 +509,7 @@
   .root {
     position: relative;
     z-index: 101;
-    height: 100vh;
+    height: var(--goa-work-side-menu-height, 100vh);
```
vanessatran-ddi, line 547: Do you have the token PR for this new variable? I don't see it on 2.2.2:

Image

docs/src/components/nav/MenuSecondaryContent.tsx
```diff
@@ -33,15 +30,51 @@ export function MenuSecondaryContent({ isOpen }: MenuSecondaryContentProps) {
...
+    },
+    [],
+  );
+
   return (
     <>
       <button
         className="search-menu-button"
         onClick={openSearch}
+        onMouseEnter={showSearchTooltip}
```
vanessatran-ddi, line 70: Question: was the reason for using a custom `<button>` here (instead of wrapping WorkSideMenuItem with a sentinel URL like /__search__) specifically to display the ⌘K keyboard shortcut as trailing content?

I asked because the doc site already has an established pattern for non-navigation items using `WorkSideMenuItem` + URL + wrapper `onClick` 
For example:  [ComponentsSubMenu](https://github.com/GovAlta/ui-components/blob/dev/docs/src/components/nav/ComponentsSubMenu.tsx#L56) 

If the ⌘K  is the only blocker, we could drop it and use the badge like now instead, then we can add a story to follow up for `trailingSlot` on the WorkSideMenuItem. 

If we can use `WorkSideMenuItem` instead of `button` here then we can have tooltip feature without having to add a customized function here. 

cc @twjeffery  How do you think?

libs/web-components/src/components/work-side-menu/WorkSideMenu.svelte
bdfranck: I've created a related PR here: https://github.com/GovAlta/design-tokens/pull/146

It's not a blocker because of the fallback value.

libs/web-components/src/components/work-side-menu/WorkSideMenu.svelte
bdfranck: Thanks! Right, we don't need this for the open state. I've amended my commit with the suggested changes.

docs/src/components/nav/MenuSecondaryContent.tsx
twjeffery: @vanessatran-ddi I agree with this, I think it's a good idea. The custom button is only there because we need the ⌘K badge as trailing content, and WorkSideMenuItem doesn't support that yet.

But I'd rather not lose the shortcut hint. I just created https://github.com/GovAlta/ui-components/issues/3814 for a trailingContent slot on WorkSideMenuItem now, and then once that's in, we swap the search button over and all this custom tooltip code goes away?

So for now I think we leave the custom button as-is in this PR, and maybe just comment that it will be reverted to use the component when https://github.com/GovAlta/ui-components/issues/3814 is complete?

libs/web-components/src/components/work-side-menu/WorkSideMenu.svelte
vanessatran-ddi: @bdfranck  Can you update the `@abgov/design-tokens-v2` this repo, under `package.json` line 39,  latest version now is [2.4.0](https://www.npmjs.com/package/@abgov/design-tokens/v/2.4.0)

docs/src/components/nav/MenuSecondaryContent.tsx
vanessatran-ddi: Yes I agree.

vanessatran-ddi (approved):
Looks good!

bdfranck:
@vanessatran-ddi I've updated the `@abgov/design-tokens-v2` dependency to `2.4.0`, rebased against the latest dev, and resolved the merge conflicts.

vanessatran-ddi (approved):

vanessatran-ddi (approved):

chrisolsen (approved):

bdfranck:
I've rebased against the latest `dev` and resolved the merge conflicts.

bdfranck:
I've resolved the additional merge conflict.

PR #3788, opened 2026-04-09
chore: add bug fix workflow rule

twjeffery:
## Summary
Adds .claude/rules/bug-fix-workflow.md describing the workflow for picking up small bug fix issues from the team backlog, and adds a "Token-First Changes" section to .claude/rules/styling.md.

### Bug fix workflow rule (new file)
- Reading the issue body as the source of truth for scope
- Targeting V2 code paths
- Fixing at the source rather than working around bugs in components that use the broken thing
- Asking before guessing on ambiguity
- Branch naming, commit format, and PR template requirements
- Adding playground bug pages
- Opening PRs as drafts for review

### Token-first styling rule (addition to styling.md)
- When changing component styling, update token values rather than removing CSS that references tokens. Removing the CSS breaks the customization contract for consumers who override that token.
- Component CSS must reference component tokens, not global tokens directly. If no component token exists, create one that references the global.


Spark450 (approved):
This all makes sense, and may help with some of the issues we saw from those 1st generated issues in backlog grooming today.

bdfranck (approved):
Looks good to me too! 👍 I thought I had a comment about linking design tokens and CSS custom properties but I noticed you already addressed it in the file:

Image

.claude/rules/bug-fix-workflow.md
```diff
@@ -0,0 +1,67 @@
...
+
+If the issue has screenshots, download and view them. Extract the image URLs from the fetched page and use WebFetch to download each image, then Read the file to view it. Screenshots often contain critical visual context for the fix.
+
+## 2. Target V2 code paths
+
+Wrappers (`GoabButton`, `GoabModal`, etc.) hardcode `version="2"` on the underlying web component. Fixes should land on the V2 code paths in the Svelte source: `.v2` class selectors, `version === "2"` branches. Do not touch V1-only code paths unless the issue explicitly says to.
+
+## 3. Fix at the source
+
+The Svelte file under `libs/web-components/src/components/{component}/` is the source of truth. React and Angular wrappers are thin passthroughs and rarely need changes for a CSS-only fix.
```
ArakTaiRoth: This line directly competes with what's stated in `framework-wrappers.md`, which says "Every Svelte change requires corresponding React and Angular updates". Have to keep in mind that AI are highly literal, and if that is saying every change requires React and Angular updates, and this is saying they rarely need changes for CSS-only, it might not know what to do in certain updates.

.claude/rules/styling.md
ArakTaiRoth: This is written as an absolute, but it's not followed in every situation in our own CSS. We do use hardcoded CSS properties everywhere, not everything uses tokens.

.claude/rules/styling.md
```diff
@@ -8,6 +8,31 @@ Token naming: `--goa-[component]-[variant]-[state]-[property]`
 
 Use `rem` over `px` even when there is no token value (`20px` = `1.25rem`).
 
+## Token-First Changes
+
+When changing component styling, work through the token layer. Don't remove CSS rules that reference component tokens.
+
+**Change behavior by updating token values.** If a mobile tertiary button should become transparent, change the mobile token value to `transparent` in the token JSON. The CSS stays, the token drives the change. Removing the CSS breaks the customization contract for consumers who override that token.
+
+**Component CSS must reference component tokens, not global tokens.** If you need a value in component CSS and no component token exists, create one that references the global token.
```
ArakTaiRoth: This is also written as an absolute, and isn't followed by our own CSS. We use global-tokens in a lot of scenarios, just look at all the background-color used in `Button.svelte` as an example.

.claude/rules/styling.md
```diff
@@ -8,6 +8,31 @@ Token naming: `--goa-[component]-[variant]-[state]-[property]`
...
+```
+
+The component token is defined in `button-design-tokens.json` and can reference the global:
+```json
+"button-tertiary-hover-color-bg": {
+  "value": "{color.greyscale.200}"
+}
+```
+
+This keeps the token architecture intact: global tokens feed component tokens, component tokens feed CSS. Consumers can override at the component token level without knowing the global structure.
```
ArakTaiRoth: This competes with what's said on line 69 of this same file. This says people can override tokens, and that line says there is no CSS customization.

.claude/rules/bug-fix-workflow.md
```diff
@@ -0,0 +1,67 @@
...
+
+## 6. Make the fix and run tests
+
+Run the relevant component tests:
+
+```
+npx vitest run libs/web-components/src/components/{component}/
+```
+
+If the change affects V2-specific behavior, add or extend a browser test per `.claude/rules/testing.md`.
```
ArakTaiRoth: This kind of contradicts what's stated in `testing.md`. This says "if `x`, add a browser test". Line 50 in `testing.md` says "V2 components `NEED` their own browser tests"

twjeffery:
I appreciate the thorough review @ArakTaiRoth, all good points. I've made changes to address all of your comments.

.claude/rules/bug-fix-workflow.md
```diff
@@ -0,0 +1,67 @@
+# Bug fix workflow
+
+This is the workflow for picking up a small style or refinement bug from the team backlog. It assumes the issue has already been triaged as small and ready (issue body has clear scope, expected behavior, and any target values).
+
+## 1. Read the issue body
+
+Fetch the issue from GitHub: `https://github.com/GovAlta/ui-components/issues/{NUMBER}`. Use WebFetch on that URL if `gh` CLI is not available. The issue body is the source of truth for scope. Do not infer additional work from related issues or "while you're in the file" thinking.
```
ArakTaiRoth: The reference to `WebFetch` might be Claude specific? Could we generalize this so it works for Claude, but could also provide guidance for other AI's? Same with below.

twjeffery: Good call. Dropped the WebFetch references entirely and just described the action ("fetch the URL directly", "download each image, then view it"). Keeps the rule portable across AI environments without naming any tool specifically.

PR #3793, opened 2026-04-09
chore: Modified route generation for PR apps

### LINKED ISSUE #3439: Adjust Playgrounds
# Info

Currently our playgrounds (for both Reach and Angular), require editing route files and html files to create links and routes to the various PRs that we create.

We want an adjustment to our playgrounds to something that doesn't require that. That would build its links and structure based off of the folder structure instead.

This will help deal with ongoing conflict issues we have in PRs, consistently.

## Acceptance Criteria:

1. We have an idea of what we want to implement that will achieve our goal of a self route creation playground that works for both Angular and React

ArakTaiRoth:
# Before (the change)

We would always get conflicts for every PR because of the need to have a shared route list for the PR apps, for both React and Angular.

# After (the change)

We no longer have a shared route list for either Angular or React

## React

Leverages Vite and uses `import.meta.glob` to import definition files for every route. These files are one per route, and thus not needing to be edited for every PR, they are instead created for the sole PR that needs it. This is used to create the routes for use inside of React.

## Angular

Is different because Angular leverages esbuild and not Vite, so we don't have access to `import.meta.glob`. Instead we have a route generator that uses *.json files inside of every route (just like React, created per PR instead of edited for every PR) to generate a manifest which is then used to create the routes for Angular. This manifest is not shared (in `.gitignore`). This generator is run when you use `npm run build`.

## Steps needed to test

You'll note that this isn't done or complete. I didn't convert every single existing route, because I don't know if people like this or not. Instead I converted 3 bug routes and 1 feature route for both Angular and React to give a test or an example of what we could do. If people like this, and it eventually gets accepted, I'll convert all of our existing routes to what's being done here.


apps/prs/angular/src/routes/bugs/3337/bug3337.route.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "bug",
+  "id": "3337",
+  "path": "bugs/3337",
+  "title": "Input autocomplete styling",
+  "componentImportPath": "../../routes/bugs/3337/bug3337.component",
+  "componentClassName": "Bug3337Component"
```
willcodeforcoffee: Since we're already imposing a naming convention for paths, files and classes it would be great if we could just cut out and derive `path`, `componentImportPath` and `componentClassName` from the `type` and `id`.
It would make things less error prone from copypasta.

apps/prs/angular/src/app/app.component.html
```diff
@@ -53,356 +22,20 @@
...
-          <goab-work-side-menu-item
-            label="3544 Optional Side Menu Icons"
-            url="/features/3544"
-          ></goab-work-side-menu-item>
+          @for (route of featureRouteDefinitions; track route.path) {
+            <goab-work-side-menu-item
+              [label]="route.id + ' - ' + route.title"
+              [url]="'/' + route.path"
+            ></goab-work-side-menu-item>
+          }
```
willcodeforcoffee, line 41: so small

apps/prs/angular/src/routes/bugs/3337/bug3337.route.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "bug",
+  "id": "3337",
+  "path": "bugs/3337",
+  "title": "Input autocomplete styling",
+  "componentImportPath": "../../routes/bugs/3337/bug3337.component",
```
vanessatran-ddi: can we make this `componentImportPath` simpler like this `"componentImportPath": "./bug3337.component"`, reason is this file is next to the component file, so we can make a mistake to get the correct path. Could the generator accept paths relative to the route JSON itself?

apps/prs/angular/project.json
```diff
@@ -6,8 +6,16 @@
...
+    "generate-route-manifest": {
+      "executor": "nx:run-commands",
+      "outputs": ["{projectRoot}/src/app/generated/pr-route-manifest.generated.ts"],
+      "options": {
+        "command": "node scripts/generate-angular-prs-route-manifest.js"
+      }
+    },
     "build": {
       "executor": "@angular-devkit/build-angular:application",
+      "dependsOn": ["generate-route-manifest", "^build"],
```
vanessatran-ddi, line 18: we should add `dependsOn` at `serve` and `lint` steps too, otherwise the first time when we run `npm run serve:prs:angular` will have an error as it cannot find the generated file

apps/prs/react/src/app/route-manifest.ts
```diff
@@ -0,0 +1,33 @@
...
+  path: string;
+  component: ComponentType;
+  title: string;
+};
+
+type PrRouteModule = {
+  default: PrRouteDefinition;
+};
+
+const routeModules = import.meta.glob<PrRouteModule>("./routes/**/*.{ts,tsx}", {
```
vanessatran-ddi: if there are helper ts files it will be counted here, so I suggest to narrow it down to `./routes/**/*.route.{ts,tsx}`

apps/prs/angular/src/routes/bugs/3337/bug3337.route.json
vanessatran-ddi: I will resolve this as I see @willcodeforcoffee  commented above.

apps/prs/angular/project.json
ArakTaiRoth: @vanessatran-ddi I don't think so. `serve` already runs the build when used, adding it there will make it re-run infinitely. For `lint`, it doesn't completely require those files to exist, so it won't break if they don't exist.

apps/prs/react/src/app/route-manifest.ts
ArakTaiRoth: @vanessatran-ddi I made the change

apps/prs/angular/src/routes/bugs/3337/bug3337.route.json
ArakTaiRoth: Change made, it did make the route generation slightly more complicated, but it does ease off in creating these files

twjeffery (commented):
@ArakTaiRoth Looks great overall to me!

I know sometimes we want to show things without the header and sidebar, like full-page examples with custom layouts (bugs/3279 and features/2885 do this currently). With the new setup, is there a way to do that? For example, something like a `standalone` flag in the route file so it skips the header and sidebar.

ArakTaiRoth:
> @ArakTaiRoth Looks great overall to me!
> 
> I know sometimes we want to show things without the header and sidebar, like full-page examples with custom layouts (bugs/3279 and features/2885 do this currently). With the new setup, is there a way to do that? For example, something like a `standalone` flag in the route file so it skips the header and sidebar.

@twjeffery Spoke to you earlier about this, putting it here for posterity. Only `features/2885` does this, and only for Angular. The way it's currently setup for Angular, will remain in this PR and still function as it does in the current branch, there's no need to modify anything to keep what's currently working.

vanessatran-ddi (dismissed):
Looks good!

willcodeforcoffee (dismissed):
@ArakTaiRoth Ironically this needs rebasing for merge confricts, but otherwise LGTM. 😆

willcodeforcoffee (approved):
Its too bad we can't use the same method for React and Angular. It would be really sweet if they both used JSON. I understand the reasons, though.
This looks really great! Its going to be so nice to not have to spend every morning rebasing PRs!

apps/prs/angular/project.json
vanessatran-ddi, line 18: When I check out this branch and run the below steps:
* `npm run dev:watch`
* `npm run serve:prs:angular`
I have the below errors:

Image

`serve` target needs dependsOn: ["generate-route-manifest"]

vanessatran-ddi: Otherwise when we checkout different branches, the routes will be stale and we need to run build script all the time, which doesn't make use of `dev:watch` and `serve:prs:` script now

vanessatran-ddi (approved):
Looks good!

PR #3801, opened 2026-04-10
fix(#3648): remove edge treatment from high emphasis badges

### LINKED ISSUE #3648: Badge refinements
## Overview

## Acceptance Criteria:
- High emphasis badges in the browser have a border that contrasts with the badge background the Figma design does not
- Visual padding with icon and text variants "looks" larger than Figma design (as measured it seems fine)

twjeffery:
## Summary
- Remove the inner box-shadow border from strong (high) emphasis badges in V2
- The edge treatment was an `inset box-shadow` via `--goa-badge-border` that Figma does not include
- Subtle emphasis badges are unaffected, they retain their individual box-shadow tokens

Fixes #3648

## Steps needed to test
2. Navigate to bugs/3648
3. Verify strong emphasis badges (all types, medium and large) have no visible inner border/shadow
4. Verify subtle emphasis badges still show their edge treatment

image

### DIFF
--- a/libs/web-components/src/components/badge/Badge.svelte
+++ b/libs/web-components/src/components/badge/Badge.svelte
@@ -77,7 +77,11 @@
   type BadgeType = (typeof Types)[number];
   type BadgeSize = (typeof badgeSizes)[number];
   type BadgeVersion = (typeof versions)[number];
-  type BadgeJustifyContent = "center" | "flex-start" | "flex-end" | "space-between";
+  type BadgeJustifyContent =
+    | "center"
+    | "flex-start"
+    | "flex-end"
+    | "space-between";
 
   /** Defines the context and colour of the badge. */
   export let type: BadgeType;
@@ -382,6 +386,10 @@
 
   .v2 .goa-badge-content {
     padding-bottom: 0;
+  }
+
+  .v2.badge-strong {
+    box-shadow: none;
   }
 
   /* Version 2: Default Colours */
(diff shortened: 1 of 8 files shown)


apps/prs/react/src/app/app.tsx
```diff
@@ -11,6 +11,7 @@ import {
 } from "@abgov/react-components";
 
 import "@abgov/style";
+import "@abgov/design-tokens-v2/dist/tokens.css";
```
ArakTaiRoth: Like the other refinements, and I highly suspect all of them. This is fine if this is what we want to do, but we also should do it for Angular, because right now the Angular test doesn't load any v2 tokens at all.

libs/web-components/src/components/badge/Badge.svelte
```diff
@@ -384,6 +388,10 @@
     padding-bottom: 0;
   }
 
+  .v2.badge-strong {
+    box-shadow: none;
```
ArakTaiRoth, line 392: There doesn't appear to be any test for the Badge that tests that strong badges don't have a shadow, and subtle badges do have a shadow. Maybe we should add that?

twjeffery: Added. Two tests for V2 emphasis classes (strong and subtle).

ArakTaiRoth: @twjeffery The tests seem to just test that the classes exist. These would still pass if the CSS was changed or if the CSS for them was just flat out removed, which is what I think should actually be tested.

Spark450 (approved):
Verified visual treatments are updated.

libs/web-components/src/components/badge/Badge.svelte
twjeffery: I added a browser test that checks actual computed box-shadow values. Strong should be "none", subtle should contain "inset".

ArakTaiRoth (approved):

PR #3808, opened 2026-04-13
fix(#3681): dropdown selected item highlight state

### LINKED ISSUE #3681: Dropdown refinements
## Overview
Refinements for the dropdown component to match design intent.

## Acceptance Criteria
Menu option states:
- Default (not the currently chosen value, not hovered): text color `text-secondary`
- Hovered (not the currently chosen value): text color `text-default`, background `greyscale-100`
- Currently chosen value (not hovered): text color `text-light`, background `interactive`
- Currently chosen value and hovered: text color `text-light`, background `interactive-hover`

## Scope note
This PR addresses only the menu option color states above.

## Deferred
- Focus ring behavior on mouse vs keyboard is covered by #3605 (systematic focus-visible across all interactive components)
- A scroll shadow treatment for long option lists will be explored in Figma first before any code change

twjeffery:
## Summary
- Clear the highlighted index when an item is selected and when the menu opens, so the selected item shows its proper selected state (regular blue) instead of the selected-hover state (darker blue)
- Previously, selecting an item set `_highlightedIndex` to that item's index, which persisted when the menu reopened, causing both `.dropdown-item--highlighted` and `[aria-selected="true"]` to apply simultaneously

Note: The default (non-selected, non-hovered) menu item text color should also change from `text-default` to `text-secondary` per the issue. This is a token value fix tracked in a separate design-tokens PR.

Fixes #3681

## Steps needed to test
2. Navigate to bugs/3681
3. Open the first dropdown (Red is pre-selected). Verify Red shows regular blue background.
4. Select a different item (e.g. Blue). Reopen the dropdown.
5. Verify the newly selected item shows regular blue (not the darker hover blue).
6. Hover the selected item. Verify it gets the darker blue on hover only.
7. Test keyboard navigation (arrow keys + Enter) still works.

image
image

### DIFF
--- a/apps/prs/angular/src/routes/bugs/3681/bug3681.component.html
+++ b/apps/prs/angular/src/routes/bugs/3681/bug3681.component.html
@@ -1,0 +1,43 @@
+<div>
+  <h1>3681 - Dropdown refinements</h1>
+  <p>
+    <a
+      href="https://github.com/GovAlta/ui-components/issues/3681"
+      target="_blank"
+      rel="noopener"
+    >
+      View on GitHub
+    </a>
+  </p>
+  <p>
+    Four menu option color states need fixing: Default (not chosen, not hovered),
+    Hovered (not chosen), Currently chosen (not hovered), Currently chosen + hovered.
+  </p>
+
+  <h2>Dropdown with pre-selected value</h2>
+  <p>
+    Open the dropdown and inspect the 4 states: 1) Default items (not Red),
+    2) Hover a non-chosen item, 3) The chosen item (Red), 4) Hover the chosen item.
+  </p>
+  <goab-form-item label="Favorite color">
+    <goab-dropdown [value]="value" (onChange)="handleChange($event)">
+      <goab-dropdown-item value="red" label="Red"></goab-dropdown-item>
+      <goab-dropdown-item value="blue" label="Blue"></goab-dropdown-item>
+      <goab-dropdown-item value="green" label="Green"></goab-dropdown-item>
+      <goab-dropdown-item value="yellow" label="Yellow"></goab-dropdown-item>
+      <goab-dropdown-item value="purple" label="Purple"></goab-dropdown-item>
+    </goab-dropdown>
+  </goab-form-item>
+
+  <h2>Dropdown with no initial value</h2>
+  <p>All items should show as default state. Select one to see the chosen state.</p>
+  <goab-form-item label="Select a province">
+    <goab-dropdown>
+      <goab-dropdown-item value="ab" label="Alberta"></goab-dropdown-item>
+      <goab-dropdown-item value="bc" label="British Columbia"></goab-dropdown-item>
+      <goab-dropdown-item value="sk" label="Saskatchewan"></goab-dropdown-item>
+      <goab-dropdown-item value="mb" label="Manitoba"></goab-dropdown-item>
+      <goab-dropdown-item value="on" label="Ontario"></goab-dropdown-item>
+    </goab-dropdown>
+  </goab-form-item>
+</div>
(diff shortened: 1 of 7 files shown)


twjeffery:
## Design token dependency

The default menu item text color fix (`text-default` to `text-secondary`) is tracked in a separate design-tokens PR:

- GovAlta/design-tokens#151 — dropdown default item text color to text-secondary

apps/prs/angular/src/routes/bugs/3681/bug3681.component.html
```diff
@@ -0,0 +1,35 @@
+<div>
```
ArakTaiRoth, line 1: Like all other refinement stories, Angular is not loading the v2 tokens.

apps/prs/angular/src/routes/bugs/3681/bug3681.component.html
```diff
@@ -0,0 +1,35 @@
...
+      rel="noopener"
+    >
+      View on GitHub
+    </a>
+  </p>
+
+  <h2>Dropdown with pre-selected value</h2>
+  <p>Open and inspect the 4 menu option states.</p>
+  <goab-form-item label="Favorite color">
+    <goab-dropdown [value]="value">
```
ArakTaiRoth: There is no `onChange` here like there is in the React test.

apps/prs/angular/src/routes/bugs/3681/bug3681.component.html
```diff
@@ -0,0 +1,35 @@
+<div>
+  <h1>3681 - Dropdown refinements</h1>
```
ArakTaiRoth, line 2: React includes the issue description and details. Not a big deal, but also not the same. Also this is using standard html components, not our components for text.

apps/prs/angular/src/app/app.routes.ts
```diff
@@ -150,6 +151,7 @@ export const appRoutes: Route[] = [
   { path: "bugs/3685", component: Bug3685Component },
   { path: "bugs/3635", component: Bug3635Component },
   { path: "bugs/3637", component: Bug3637Component },
+  { path: "bugs/3681", component: Bug3681Component },
```
ArakTaiRoth: Route setup, but no link in SideBar

apps/prs/angular/src/routes/bugs/3681/bug3681.component.html
twjeffery: added

apps/prs/angular/src/routes/bugs/3681/bug3681.component.html
twjeffery: added

ArakTaiRoth (approved):

bdfranck (approved):
I looked at the changes...

- ✅ When I open a dropdown with a selected item, I see the selected item is not in the hover state until I hover it

Image

Looks good to me! 👍

chrisolsen:
The majority of the code in the svelte file is to scroll the item into view. Is that the intention of this PR?

As for the highlighting, it doesn't seem to work after you select something.

[Screencast_20260416_105735.webm](https://github.com/user-attachments/assets/66987d44-0053-4d3a-82b6-8f4191331821)

twjeffery:
> As for the highlighting, it doesn't seem to work after you select something.

@chrisolsen I'm not seeing the same background hover issue as you, it's correctly working for me, the background colour of the selected item is changing on hover.

https://jam.dev/c/0a917689-f417-4f37-8bcc-7c237d4c1e16

twjeffery:
> The majority of the code in the svelte file is to scroll the item into view. Is that the intention of this PR?

@chrisolsen fair point on the scope.

I took another look. Most of the extra code wasn't really needed, so I stripped it back to a single line.

Now the only change is: when you click an item, we clear the hover styling from it. So when you reopen the dropdown, the chosen item shows as chosen, not chosen-and-hovered. That's what the four colour states in the issue are about.

Rebased onto dev.

chrisolsen (approved):

PR #3754, opened 2026-04-01
feat(#3737): use wrapper JSDocs in docs

bdfranck:
This PR updates Docs to use the new JSDocs comments in the Angular and React wrappers. Changes include:

- Updates `extract-api` script to include wrappers
- Code snippet tabs controls which properties are visible
- Fixes issues in underlying wrapper and Svelte component JSDocs

docs-tab

Issue #150, opened 2021-10-22
Grid system


twjeffery:
# What is being submitted?

# Why are you submitting?

# Examples, links, and research

# Anything else

PR #3809, opened 2026-04-13
fix(#3667): notification banner height refinements

### LINKED ISSUE #3667: Notification banner refinements
## Overview
Refinements for the notification banner component to match design intent.

## Acceptance Criteria
- Silently undocument the "Event" type and mark it for long-term deprecation. Remove it from the docs site (variant examples and prop docs) so new consumers do not find it. Leave the component code in place so existing users do not break. The Event type is functionally the same as Info, so users can use Info instead.
- Fix banner heights:
  - Compact: 64px (currently ~66px)
  - Default: 80px (currently ~82px)
- Ensure content (icon, banner text, dismiss button) is vertically centered within the banner at both heights

## Scope note
All items above will be addressed in one PR.

twjeffery:
## Summary
- Adjust V2 notification banner padding to hit exact target heights: 80px default, 64px compact
- Nudge text content up 1px for improved vertical centering (margin-top 4px to 3px)
- Event type was already not documented in configs, no docs change needed

Fixes #3667

## Steps needed to test
2. Navigate to bugs/3667
3. Inspect the default information banner height. Should be exactly 80px.
4. Inspect the compact information banner height. Should be exactly 64px.
5. Check vertical centering of icon, text, and close button at both sizes.

image

### DIFF
--- a/apps/prs/angular/src/routes/bugs/3667/bug3667.component.html
+++ b/apps/prs/angular/src/routes/bugs/3667/bug3667.component.html
@@ -1,0 +1,93 @@
+<div>
+  <goab-text tag="h1" mt="m">Bug #3667: Notification banner refinements</goab-text>
+
+  <goab-block>
+    <goab-link trailingIcon="open">
+      <a
+        href="https://github.com/GovAlta/ui-components/issues/3667"
+        target="_blank"
+        rel="noopener"
+      >
+        View on GitHub
+      </a>
+    </goab-link>
+
+    <goab-details heading="Issue Description">
+      <goab-text tag="p">
+        Notification banner height refinements. Compact should be 64px (currently ~66px),
+        default should be 80px (currently ~82px). Also need to verify vertical centering of
+        content (icon, text, dismiss button) at both sizes. The Event type should be silently
+        undocumented (docs removal only, no code change needed).
+      </goab-text>
+    </goab-details>
+  </goab-block>
+
+  <goab-divider mt="l" mb="l"></goab-divider>
+
+  <goab-text tag="h2">Test Cases</goab-text>
+
+  <!-- Test 1: Height measurements -->
+  <goab-text tag="h3">Test 1: Height at default and compact sizes</goab-text>
+  <goab-text tag="p">
+    Measure the rendered height of each banner. Default should be 80px, compact should be
+    64px. Use browser dev tools to inspect the goa-notification element and measure the outer
+    height.
+  </goab-text>
+
+  <goab-text tag="h4" mt="m">Default size (target: 80px)</goab-text>
+  <goab-notification type="information">
+    Default information banner for height measurement.
+  </goab-notification>
+
+  <goab-text tag="h4" mt="m">Compact size (target: 64px)</goab-text>
+  <goab-notification type="information" [compact]="true">
+    Compact information banner for height measurement.
+  </goab-notification>
+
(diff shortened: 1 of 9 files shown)


twjeffery:
## Design token dependency

The height fix (80px default, 64px compact) requires updated padding token values in:

- GovAlta/design-tokens#150 — notification banner padding-tb reduced by 1px each (also includes the important high/low close hover swap fix)

ArakTaiRoth:
Regarding the "silently" undocument the `event` type. This PR does nothing towards that (it's still included in the .svelte file, and the associated wrappers). I think this is something for #3754 @bdfranck . But we might need to figure out how to undocument something while still keeping it as an option for backwards compatibility.

apps/prs/angular/src/routes/bugs/3667/bug3667.component.ts
```diff
@@ -0,0 +1,10 @@
+import { Component } from "@angular/core";
```
ArakTaiRoth: Doesn't use v2 tokens

apps/prs/angular/src/routes/bugs/3667/bug3667.component.html
```diff
@@ -0,0 +1,27 @@
+<div>
```
ArakTaiRoth, line 1: This is a very basic test compared to the React version. It should be the same, so we can test the same stuff between languages.

apps/prs/angular/src/routes/bugs/3667/bug3667.component.ts
twjeffery: fixed already

apps/prs/angular/src/routes/bugs/3667/bug3667.component.html
twjeffery: Updated to match React with all three test sections.

ArakTaiRoth (approved):

libs/web-components/src/components/notification/Notification.svelte
```diff
@@ -189,7 +184,7 @@
 
   /* V2: Improved vertical alignment with icon */
   .v2 .content {
-    margin-top: 4px;
+    margin-top: 3px;
```
bdfranck, line 187: Is this the only change to the component? The rest look like formatting changes.

bdfranck (commented):
I looked at the changes...

- ✅ I can see the reduced top margin for content.

Image

- 🤔 The component heights are slightly more than `80px` and `64px`

Image

Image

I'm not seeing the height changes. Is there a corresponding `design-tokens` PR? Or maybe something got lost during a rebase?

libs/web-components/src/components/notification/Notification.svelte
twjeffery: The height change comes from design-tokens #150 which just merged. The component uses --goa-notification-banner-padding-tb and the compact version. The repo is still on design-tokens 2.2.2 so the playground shows the old V1 values (82/66). I verified locally with 2.4.2 installed and it shows 80/64. The only component change here is the content margin adjustment. 

The rest is Prettier reformatting.

bdfranck (approved):
I looked at the changes again with the latest tokens...

- ✅ I see the regular version is `80px` high

Image

- ✅  I see the compact version is `64px` high

Image

Looks good to me! 👍

PR #3813, opened 2026-04-13
fix(#3630): drawer exit animation, close button alignment, docs updates

### LINKED ISSUE #3630: Drawer refinements
## Overview
Refinements for the drawer component to match design intent.

## Acceptance Criteria
- Animate entry and exit using the existing motion tokens
- Vertically align the heading and close button. Likely fix: the close icon button in the heading area is size medium when it should be size small, which should bring it into vertical alignment with the heading text.
- On the docs site, the drawer example should use the compact button group (currently correct) and the button group should be aligned with `align="start"` instead of `align="end"`
- Move the component from the "content and layout" functional category to "structure and navigation" on the docs site (MDX file update)

## Scope note
All items above will be addressed in one PR.

twjeffery:
## Summary
- Enable exit animation on drawer close (was disappearing instantly due to `visibility: hidden`)
- Replace hardcoded transition durations with motion tokens (`motion-duration-medium-1`, `motion-duration-short-4`, `motion-curve-expressive`)
- Nudge close icon button up for better alignment with heading text
- Update docs example to use `alignment="start"` on button group (was `end`)
- Move drawer category from content-layout to structure-navigation in docs

Fixes #3630

## Steps needed to test
2. Navigate to bugs/3630
3. Open and close the right drawer. Verify it slides out smoothly (not just disappearing).
4. Open the right drawer with actions. Verify buttons are left-aligned and compact.
5. Open the left drawer. Verify slide-in from left and slide-out animation.
6. Open the bottom drawer. Verify slide-up and slide-down animation.
7. Open the long heading drawer. Verify close button alignment with wrapped heading.
8. Check that the heading and close button are vertically aligned in all drawers.

https://github.com/user-attachments/assets/e7ddee31-7110-4592-be09-f1e9cf123e17

image

### DIFF
--- a/apps/prs/angular/src/routes/bugs/3630/bug3630.component.html
+++ b/apps/prs/angular/src/routes/bugs/3630/bug3630.component.html
@@ -1,0 +1,107 @@
+<div>
+  <h1>Bug #3630: Drawer refinements</h1>
+  <p>
+    <a
+      href="https://github.com/GovAlta/ui-components/issues/3630"
+      target="_blank"
+      rel="noopener"
+    >
+      View on GitHub
+    </a>
+  </p>
+
+  <hr />
+
+  <h2>Test Cases</h2>
+
+  <h3>Test 1: Right drawer (animation + alignment)</h3>
+  <p>
+    Open and close. Check slide-in and slide-out animation. Check heading and close button
+    vertical alignment.
+  </p>
+  <goab-button (onClick)="rightOpen = true">Open right drawer</goab-button>
+
+  <goab-drawer
+    heading="Right drawer"
+    position="right"
+    maxSize="400px"
+    [open]="rightOpen"
+    (onClose)="rightOpen = false"
+  >
+    <p>
+      Check the slide-out animation when closing. The close button should align well with
+      the heading text.
+    </p>
+  </goab-drawer>
+
+  <hr />
+
+  <h3>Test 2: Right drawer with actions</h3>
+  <p>Actions should use start alignment with compact buttons.</p>
+  <goab-button (onClick)="rightActionsOpen = true">Open right drawer with actions</goab-button>
+
+  <ng-template #rightActionsTemplate>
+    <goab-button-group alignment="start">
+      <goab-button type="primary" size="compact">Save</goab-button>
+      <goab-button type="tertiary" size="compact" (onClick)="rightActionsOpen = false">Cancel</goab-button>
+    </goab-button-group>
+  </ng-template>
+
+  <goab-drawer
+    heading="Edit settings"
+    position="right"
+    maxSize="400px"
+    [open]="rightActionsOpen"
+    [actions]="rightActionsTemplate"
+    (onClose)="rightActionsOpen = false"
+  >
+    <p>The action buttons below should be left-aligned and compact.</p>
+  </goab-drawer>
+
+  <hr />
+
+  <h3>Test 3: Left drawer</h3>
+  <p>Check slide-in from left and slide-out animation.</p>
+  <goab-button (onClick)="leftOpen = true">Open left drawer</goab-button>
+
+  <goab-drawer
+    heading="Left drawer"
+    position="left"
+    maxSize="400px"
+    [open]="leftOpen"
+    (onClose)="leftOpen = false"
+  >
+    <p>Check the slide-out animation when closing from the left side.</p>
+  </goab-drawer>
+
+  <hr />
+
+  <h3>Test 4: Bottom drawer</h3>
+  <p>Check slide-up from bottom and slide-down animation.</p>
+  <goab-button (onClick)="bottomOpen = true">Open bottom drawer</goab-button>
+
+  <goab-drawer
+    heading="Bottom drawer"
+    position="bottom"
+    [open]="bottomOpen"
+    (onClose)="bottomOpen = false"
+  >
+    <p>Check the slide-down animation when closing from the bottom.</p>
+  </goab-drawer>
+
+  <hr />
+
+  <h3>Test 5: Long heading</h3>
+  <p>Check alignment when the heading wraps to two lines.</p>
+  <goab-button (onClick)="longHeadingOpen = true">Open drawer with long heading</goab-button>
+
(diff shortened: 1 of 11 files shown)


twjeffery:
## Design token dependency

The drawer content vertical padding fix (16px to 24px) is tracked in a separate design-tokens PR:

- GovAlta/design-tokens#152 — drawer content vertical padding to space-l

apps/prs/angular/src/routes/bugs/3630/bug3630.component.html
```diff
@@ -0,0 +1,45 @@
+<div>
```
ArakTaiRoth, line 1: This doesn't test the same stuff as the React PR does. The React PR covers actions/start-alignment, and a long-heading alignment. Also this is using base html text components, rather than our components.

apps/prs/angular/src/routes/bugs/3630/bug3630.component.ts
```diff
@@ -0,0 +1,14 @@
+import { Component } from "@angular/core";
+import { GoabDrawer, GoabButton, GoabButtonGroup } from "@abgov/angular-components";
```
ArakTaiRoth, line 2: `GoabButtonGroup` is currently unused. It would be used though if Angular had the same tests as React

libs/web-components/src/components/drawer/Drawer.svelte
```diff
@@ -434,6 +432,10 @@
     border-bottom: none; /* Remove border by default */
   }
 
+  .v2 .header goa-icon-button {
+    margin-top: -0.1875rem;
```
ArakTaiRoth, line 436: This is a weird fix, as this is a highly specific number, and not a token. Did we try the suggested fix of changing the size of the icon?

twjeffery: yea, I tried to change the icon, but then it's too small. This just offsets it so the heading and the icon can vertically align without moving the header down (and taking up more space)

apps/prs/angular/src/routes/bugs/3630/bug3630.component.html
twjeffery: Updated with all 5 test cases matching React, including actions and long heading.

apps/prs/angular/src/routes/bugs/3630/bug3630.component.ts
twjeffery: Now used in the actions test.

libs/web-components/src/components/drawer/drawer.spec.ts
```diff
@@ -10,45 +10,47 @@ beforeEach(() => {
...
+        expect(drawer?.style.pointerEvents).toBe("auto");
+      },
+      { timeout: 1000 },
+    );
 
     // Click background to close
     const background = el.queryByTestId("background");
     await fireEvent.click(background);
 
     setTimeout(() => {
```
ArakTaiRoth, line 55: Because these tests are in a setTimeout, the test doesn't actually wait for them to run before finishing the test. So the test can actually finish before these expect statements are actually run. `setTimeout` doesn't actually halt execution.

libs/react-components/specs/drawer.browser.spec.tsx
ArakTaiRoth: Sorry, I just noticed this and we should probably fix it now, rather then continue with breaking tests possibly. This inner waitFor isn't actually being `awaited`, so how this works is everything before this line is being waited for, actual code execution has stopped until it happens. However everything inside this waitFor isn't actually being waited on, so it may or may not happen before the test is over.

A better way to do this is to flatten it out. Remove the outer waitFor, and add `await` in front of the inner waitFor. Test this before though, because I could be wrong.

libs/react-components/specs/pushdrawer.browser.spec.tsx
ArakTaiRoth: Same as what I said above for `drawer.browser.spec.tsx`

libs/web-components/src/components/drawer/Drawer.svelte
bdfranck, line 436: Did you try setting the component design tokens? Here's how it looks when you set the following:

```CSS
--goa-icon-button-medium-padding: var(--goa-space-2xs) var(--goa-space-xs) var(--goa-space-2xs);
```

Image

Image

bdfranck (commented):
I looked at the changes...

- ✅ The drawer close animation works on left, right, and bottom Drawers
- ✅ The animations use design tokens
- ✅ Docs: The "With actions" example Drawer has a left-aligned button group
- ✅ Docs: The component category is "Structure and Navigation"

Looks good! I just have one small comment about the button.

libs/web-components/src/components/drawer/Drawer.svelte
twjeffery: @bdfranck if we do that, then the size of the icon button (and the touch target) changes from 40x40px to 40x32px. Maybe that's just fine? I'm not sure if the trade off is worth it there

### adjusted icon button positioning
image

### smaller icon button padding
image

libs/react-components/specs/drawer.browser.spec.tsx
twjeffery: Fixed, flattened it out like you suggested.

libs/web-components/src/components/drawer/drawer.spec.ts
twjeffery: This was already broken on dev. The close event doesn't propagate in the unit test environment, so properly awaiting it just reveals a pre-existing issue. Left the setTimeout for now with a TODO. Needs a separate investigation.

libs/web-components/src/components/drawer/Drawer.svelte
bdfranck: Makes sense! Yeah, the end user won't notice the negative margin but will be impacted by the smaller touch target.

bdfranck (approved):
Looks good to me! 👍

ArakTaiRoth (approved):

PR #3835, opened 2026-04-16
chore: set up github pages per pull request

### LINKED ISSUE #3829: Setup Github Preview
# Info

Netlify preview will no longer work for our needs. With every PR creating a deploy preview link and us using a private Netlify account, it doesn't give us the amount of uses we need.

We have tried to sign up for a free "paid" account because we're open source, that was denied.

Our next best option is using Github Preview on our builds to see if this works.

## Acceptance Criteria:

1. Setup Github Preview in our `dev` branch pr builds to see if it works

vanessatran-ddi:
# Before (the change)

# After (the change)

### DIFF
--- a/.github/workflows/pr-preview.yml
+++ b/.github/workflows/pr-preview.yml
@@ -1,0 +1,110 @@
+name: PR Preview (GitHub Pages)
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, closed]
+    branches:
+      - dev
+
+concurrency:
+  group: preview-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  deploy-preview:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
+
+      - uses: actions/setup-node@v6
+        with:
+          node-version: "24"
+          cache: "npm"
+          cache-dependency-path: |
+            package-lock.json
+            docs/package-lock.json
+
+      - name: Install npm 11
+        run: npm install -g npm@11
+
+      - name: Install dependencies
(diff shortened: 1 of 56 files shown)


twjeffery (approved):
@vanessatran-ddi I tested the three previews, everything works. Nav links show correctly, subpaths work, docs search loads correctly, and both playgrounds work well

willcodeforcoffee (approved):
Looks good! Can't wait to try it out! 😄

ArakTaiRoth:
@vanessatran-ddi Just a quick question? Would it be possible to modify this to only deploy the paths that need to be deployed based on changes made? If a PR is a docs only PR, then we don't need to deploy Angular and React previews for example, and vice versa. Not a blocker, just a question if we can do something quickly and easily.

vanessatran-ddi:
> @vanessatran-ddi Just a quick question? Would it be possible to modify this to only deploy the paths that need to be deployed based on changes made? If a PR is a docs only PR, then we don't need to deploy Angular and React previews for example, and vice versa. Not a blocker, just a question if we can do something quickly and easily.

@ArakTaiRoth yes technically it is possible with this [add-on plugin](https://github.com/dorny/paths-filter) but I don't think it is quick and easy. 
1. Most of our PRs touch the svelte components, so it may affect React and Angular wrapper, so in practice, both React and Angular playgrounds would still need to rebuild on the majority of Prs. 
2. To use the above plugin, I will need to modify the yaml file to split the job, add condition for each deploy step to build dynamically. 
3. The current workflow is simple and predictable - every 3 Prs get 3 links, reviewers always know where to look. Conditional gate may lead to an error if the first commit doesn't touch the Svelte, but the second does and so on ... 
4. For some Copilot bump dependencies, having 3 preview links are a good way to make sure nothing broke across the stack. 

We can revisit if `gh-pages` branch becomes bigger and bigger, but for now I would rather keep this PR simple and ship it.  What do you think?

PR #3839, opened 2026-04-17
fix(#3643): Popover aligns right when out of space

### LINKED ISSUE #3643: Popover: Overflows right edge when trigger is near viewport boundary
## Description

The Popover rewrite (PR #3478) switched to CSS Anchor Positioning, which always left-aligns the popover to the trigger via `inset-inline-start: anchor(left)`. When the trigger is near the right edge of the viewport, the popover content extends off-screen.

The previous JS-based positioning handled this automatically by clamping to the viewport.

## To reproduce

Place a Popover trigger near the right viewport edge with a `maxWidth` wider than the trigger (e.g. 280px). Open it. The content overflows right.

## Possible fixes

- Add an `alignment` prop (`start` / `center` / `end`) so consumers can control horizontal alignment
- Use `position-try-fallbacks` in the anchor positioning CSS to automatically flip when it would overflow

## Workaround

Setting `--offset-left` as a CSS custom property on a parent element shifts the popover via the `translate` property. It inherits into the shadow DOM. Fragile since it's a fixed pixel value.

## Screenshots

Image

Image

bdfranck:
This PR right aligns the popover with the right edge target if...
- The popover is positioned `above` or `below`
- There isn't enough horizontal space to show the left-aligned popover in the viewport

popover-alignment

It supports both our popover position implementations:
- Anchor positioning for browsers that support it
- JS fallback for browsers that don't (Safari 18)

safari-test

### Not included

It doesn't include an additional feature of a public alignment property. We'll need to decide how it fits with the current "position" property.

### DIFF
--- a/libs/web-components/src/components/popover/Popover.svelte
+++ b/libs/web-components/src/components/popover/Popover.svelte
@@ -76,6 +76,7 @@
   let _targetEl: HTMLElement;
   let _isOpen = false;
   let _autoPosition: "above" | "below" = "below";
+  let _alignment: "left" | "right" = "left";
   const _popoverId = `goa-popover-${generateRandomId()}`;
 
   $: _disabled = toBoolean(disabled);
@@ -122,14 +123,14 @@
 
     showDeprecationWarnings();
     addGlobalCloseListener();
-    window.addEventListener("resize", updateAutoPosition);
+    window.addEventListener("resize", recalculatePosition);
   });
 
   onDestroy(() => {
     if (_needsManualPositioning || _positionRafId) {
       stopManualPositioning();
     }
-    window.removeEventListener("resize", updateAutoPosition);
+    window.removeEventListener("resize", recalculatePosition);
     // true was passed when the listener was added, so it's necesary to be passed here as well
     window.removeEventListener("popstate", handleUrlChange, true);
   });
@@ -140,12 +141,12 @@
     if (!_isOpen || !_targetEl || !_popoverEl) return;
 
     const targetRect = _targetEl.getBoundingClientRect();
+    const popoverRect = _popoverEl.getBoundingClientRect();
     const xOffset = hoffset ? parseFloat(hoffset) : 0;
     const yOffset = voffset ? parseFloat(voffset) : 3;
 
     // Recalculate auto position based on current viewport space
     if (position === "auto") {
-      const popoverRect = _popoverEl.getBoundingClientRect();
       const spaceAbove = targetRect.top;
       const spaceBelow = window.innerHeight - targetRect.bottom;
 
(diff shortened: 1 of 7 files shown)


willcodeforcoffee (approved):
Oh man I was so jiddy when I saw this PR! I'm so glad you're doing this, @bdfranck!
I have only one question, would it be worth recalculating after scroll or resize?

https://github.com/user-attachments/assets/10a65314-f8f4-4994-b619-adc03a53269a

I added MenuButtons and they work, too:

Image

Tested on BrowserStack and it looks all good there, too. The only difference I noticed is that on iPad and iPhone Safari the MenuButtonItems are blue instead of black.

Image

I'm pretty sure that isn't related to your PR though? That is probably part of the MenuButton Refinement tickets.

I think I can approve this PR unless you wanted to add handling for resize/scrolling. This looks great, Benji. Nice stuff!

bdfranck:
@willcodeforcoffee Yeah, the original workaround only included handling on `resize` but not `scrolling` so I didn't worry about it. I don't think it will cause any issues?

I'll see if I can replicate the menu button issue on `dev` as it seems unrelated.

libs/web-components/src/components/popover/Popover.svelte
```diff
@@ -161,13 +164,17 @@
...
     } else {
       _popoverEl.style.top = `${targetRect.bottom + yOffset}px`;
-      _popoverEl.style.left = `${targetRect.left + xOffset}px`;
       _popoverEl.style.transform = "";
     }
+
+    if (_alignment === "left") {
+      _popoverEl.style.left = `${targetRect.left + xOffset}px`;
+    } else {
+      _popoverEl.style.left = `${targetRect.right + xOffset - popoverRect.width}px`;
```
vanessatran-ddi: should be `${targetRect.right - xOffset - popoverRect.width}px`

libs/web-components/src/components/popover/Popover.svelte
```diff
@@ -123,13 +124,15 @@
     showDeprecationWarnings();
     addGlobalCloseListener();
     window.addEventListener("resize", updateAutoPosition);
+    window.addEventListener("resize", updateAlignment);
```
vanessatran-ddi: Is there a reason we split `updateAlignment` into a separate function? Why dont we merge into `updateAutoPosition` too?

bdfranck: I put it in a separate function for two reasons:

1. `updateAutoPosition` only works when the position is set to `auto`. We still want to do alignment for `above` and `below`.
https://github.com/GovAlta/ui-components/blob/42b654cfb50a5a7fd105e91a507606b069744a5b/libs/web-components/src/components/popover/Popover.svelte#L313-L316
2. Alignment felt distinct from positioning so it made sense to keep separate

Is there another approach that you recommend I take?

bdfranck:
I've updated the PR with the following:
- Rebased against `dev` with routing changes
- Added new routing manifest files
- Fixed xoffset error
- Responded to the one question

libs/web-components/src/components/popover/Popover.svelte
vanessatran-ddi: Here is what I recommend: $IMG. with the current way, we calls `getBoundingClientRect()` 4 times when 2 would do, and we register 2 resize listeners when 1 would do.

vanessatran-ddi: Sorry, posting with the image: 
image

bdfranck: Makes sense! I've updated the PR to have a single function on resize: `recalculatePosition`. This function makes two `getBoundingcLientRect` calls and passes them to `updateAutoPosition` and `updateAlignment`.

bdfranck:
I've updated the PR again to reduce the number of resize listeners and `getBoundingClientRect` calls.

vanessatran-ddi (approved):

willcodeforcoffee (approved):
Looks good! Thanks Benji!

ArakTaiRoth (approved):

PR #3842, opened 2026-04-17
fix(#3602): FileUploadInput error state to match FileUploadCard

### LINKED ISSUE #3602: Document `FileUploadCard` and fix invalid file upload card state
## Summary

The `FileUploadCard` component is documented as a "subcomponent" and is not surfaced for users that may want to use it independently from the uploader. There should be examples that show both uploader and uploader cards.

There is a display issue in the upload experience. When file size validation is triggered, uploaded files are not showing in the intended card format. Files should still appear as upload cards even when they are invalid or have validation errors.

## More Info

This design issue with the File Upload Card is typically only an issue with the first uploaded file if it errors, it's not an issue with subsequent files with errors.

First file:
Image

Multiple files:
Image

## Expected behaviour
- All uploaded files should appear in `FileUploadCard` format, regardless of whether they are valid or have an error state.

## Acceptance criteria
* Add `FileUploadCard` examples to the new documentation website, ideally at least one example with the uploader and one without the uploader elements (e.g. a list of uploaded files).
* Ensure uploaded files always render as cards
* Ensure validation errors, including file size errors, are shown within the card design
* Confirm the error state matches the [intended design](https://www.figma.com/design/Jpy1Ea5qglwnp1SgGnagY9/%E2%9D%96-BETA-Component-library?node-id=14458-398198&t=PG3LdCbcDtc0Wbts-1) and is consistent with valid file cards

willcodeforcoffee:
# Before (the change)

image

File upload errors were listed in a unique state, which could cause them to not be visible when used in conjunction with the `FileUploadCards`.

# After (the change)

File upload errors are listed in the same format as FileUploadCard.

image

### DIFF
--- a/apps/prs/angular/src/routes/bugs/3602/bug3602.component.ts
+++ b/apps/prs/angular/src/routes/bugs/3602/bug3602.component.ts
@@ -1,0 +1,77 @@
+import { Component } from "@angular/core";
+import { CommonModule } from "@angular/common";
+import {
+  GoabFileUploadInputOnSelectFileDetail,
+  GoabFileUploadOnDeleteDetail,
+} from "@abgov/ui-components-common";
+import {
+  GoabFileUploadCard,
+  GoabFileUploadInput,
+  GoabFormItem,
+} from "@abgov/angular-components";
+
+type UploadedFile = {
+  id: string;
+  filename: string;
+  size: number;
+  type?: string;
+  progress: number;
+};
+
+@Component({
+  standalone: true,
+  selector: "abgov-bug3602",
+  templateUrl: "./bug3602.component.html",
+  imports: [CommonModule, GoabFileUploadInput, GoabFileUploadCard, GoabFormItem],
+})
+export class Bug3602Component {
+  smallFiles: UploadedFile[] = [];
+  largeFiles: UploadedFile[] = [];
+
+  onSmallFileSelect(detail: GoabFileUploadInputOnSelectFileDetail) {
+    const file = detail.file;
+    this.smallFiles = [
+      ...this.smallFiles,
+      {
+        id: `${file.name}-${Date.now()}`,
+        filename: file.name,
+        size: file.size,
+        type: file.type,
+        progress: -1,
+      },
+    ];
+  }
+
+  onLargeFileSelect(detail: GoabFileUploadInputOnSelectFileDetail) {
+    const file = detail.file;
+    this.largeFiles = [
+      ...this.largeFiles,
+      {
+        id: `${file.name}-${Date.now()}`,
+        filename: file.name,
+        size: file.size,
+        type: file.type,
+        progress: -1,
+      },
+    ];
+  }
+
+  private removeFirstMatchingFile(
+    files: UploadedFile[],
+    detail: GoabFileUploadOnDeleteDetail,
+  ): UploadedFile[] {
+    const index = files.findIndex((f) => f.filename === detail.filename);
+    if (index === -1) {
+      return files;
+    }
+    return [...files.slice(0, index), ...files.slice(index + 1)];
+  }
+
+  onSmallFileDelete(detail: GoabFileUploadOnDeleteDetail) {
+    this.smallFiles = this.removeFirstMatchingFile(this.smallFiles, detail);
+  }
+
+  onLargeFileDelete(detail: GoabFileUploadOnDeleteDetail) {
+    this.largeFiles = this.removeFirstMatchingFile(this.largeFiles, detail);
+  }
+}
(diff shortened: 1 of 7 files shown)


willcodeforcoffee:
@copilot apply changes based on the comments in [this thread](https://github.com/GovAlta/ui-components/pull/3842#pullrequestreview-4164260623)

libs/web-components/src/components/file-upload-input/FileUploadInput.svelte
```diff
@@ -183,6 +209,18 @@
   function onMouseOut() {
     _state = "default";
   }
+
+  function dismissOnDelete(node: HTMLElement, dismiss: () => void) {
```
vanessatran-ddi: We can remove this function `dismissOnDelete`, you can replace `use:dismissOnDelete` with `on:_delete` directly on the element

willcodeforcoffee: @vanessatran-ddi I keep learning new stuff from Svelte. Thank you!

vanessatran-ddi (approved):

vanessatran-ddi (approved):
Looks good!

bdfranck (approved):
I looked at the changes...

- ✅ When I attempt to upload a file that exceeds the limits, I see the error on the file card.
- ✅ I don't see the error message below the drag-and-drop upload input.
- ✅ The error message aligns with our "[File too large](https://design.alberta.ca/foundations/content-guidelines/error-messages/)" error message content guidelines.


Image

Looks good to me! 👍

willcodeforcoffee:
@ArakTaiRoth ready for merge when you are! :pray:

ArakTaiRoth:
@Spark450 In the story attached to this PR, you have "The FileUploadCard component is documented as a "subcomponent" and is not surfaced for users that may want to use it independently from the uploader. There should be examples that show both uploader and uploader cards."

I just want to check about this, as this requirement doesn't make sense to me. Reading this implies to me that we have File Upload Card as a completely separate component from File Upload Input and documented as such. That doesn't make sense to me because Card is only meant to be used with Input, and is not intended to be used independently. As this isn't included in the Acceptance Criteria, I just want to make sure that this is a mistake and not intended to be taken this way?

apps/prs/angular/src/routes/bugs/3602/bug3602.component.ts
```diff
@@ -0,0 +1,77 @@
...
+        progress: -1,
+      },
+    ];
+  }
+
+  private removeFirstMatchingFile(
+    files: UploadedFile[],
+    detail: GoabFileUploadOnDeleteDetail,
+  ): UploadedFile[] {
+    const index = files.findIndex((f) => f.filename === detail.filename);
```
ArakTaiRoth: Here you're deleting files based off of filename while they're tracked in the html file by id. If 2 files have the same filename (which would work because tracked by id), deleting the second item will remove the first matching item instead of the one selected.

This doesn't happen in the React playground, because it deletes via id and tracks via id.

willcodeforcoffee:
> @Spark450 In the story attached to this PR, you have "The FileUploadCard component is documented as a "subcomponent" and is not surfaced for users that may want to use it independently from the uploader. There should be examples that show both uploader and uploader cards."
> 
> I just want to check about this, as this requirement doesn't make sense to me. Reading this implies to me that we have File Upload Card as a completely separate component from File Upload Input and documented as such. That doesn't make sense to me because Card is only meant to be used with Input, and is not intended to be used independently. As this isn't included in the Acceptance Criteria, I just want to make sure that this is a mistake and not intended to be taken this way?

Mark and I had discussed this. I was planning to do the documentation updates in a separate PR. That said, Vanessa just did a _huge_ PR with a ton of documentation updates, including stuff for FileUploadInput and FileUploadCard and now I'm not sure its even necessary.

apps/prs/angular/src/routes/bugs/3602/bug3602.component.ts
willcodeforcoffee: @ArakTaiRoth Good point. It is just for the PR code, but I updated it.

Spark450:
> @Spark450 In the story attached to this PR, you have "The FileUploadCard component is documented as a "subcomponent" and is not surfaced for users that may want to use it independently from the uploader. There should be examples that show both uploader and uploader cards."
> 
> I just want to check about this, as this requirement doesn't make sense to me. Reading this implies to me that we have File Upload Card as a completely separate component from File Upload Input and documented as such. That doesn't make sense to me because Card is only meant to be used with Input, and is not intended to be used independently. As this isn't included in the Acceptance Criteria, I just want to make sure that this is a mistake and not intended to be taken this way?

@ArakTaiRoth There has been a use case surfaced to me a few times where a team would want to use the upload card without the uploader. For example a worker could be reviewing a submission that includes uploaded files. Designers have expressed interest in showing those uploaded file (cards) on the page of that submission. I think its likely that they would need a [this new feature](https://github.com/GovAlta/ui-components/issues/3899) to round out what they would want to do though.

PR #3845, opened 2026-04-20
chore(#3844): Add runtime V1/V2 token toggle to PR playgrounds

### LINKED ISSUE #3844: Easier way to switch between V1 and V2 tokens in the playground
Testing a PR in both V1 and V2 tokens was a pain. You had to comment out the V2 import in `app.tsx` or drop a `useEffect` injection on the page you're testing. Easy to forget which mode you were in, and no way to compare side by side without restarting.

Added a toggle to the side menu of the React and Angular playgrounds. Click it to switch tokens on the fly. `?tokens=v1` in the URL also works if you want a specific mode by link, and your choice persists per tab so side-by-side comparison is just opening two tabs.

Also cleaned up the per-page `useEffect` / `ngOnInit` token injections (6 React pages, 3 Angular) since the toggle makes them redundant.

twjeffery:
# Before (the change)

Testing a PR in both V1 and V2 tokens meant commenting out the V2 import in `app.tsx` or dropping a `useEffect` injection on the page you're testing. Easy to forget which mode you were in, and no way to compare side by side without restarting.

# After (the change)

A "Switch to V1 tokens" / "Switch to V2 tokens" item in the side menu of the React and Angular playgrounds flips tokens on click. Mode persists per-tab via `sessionStorage`, and `?tokens=v1` in the URL overrides if you want a specific mode by link. Default is V2, unchanged from the current global default on `dev`.

Also cleaned up the per-page `useEffect` / `ngOnInit` token injections (6 React pages, 3 Angular) since the toggle makes them redundant. The `:root` stylesheet-rule-removal hack on `feat3229` is gone too, replaced by a "V2 link always last in `<head>`" invariant in the new helper module.

No library code touched. Playground infrastructure only.

Fixes #3844.

## Steps needed to test

1. `npm run serve:prs:react`. Side menu shows "Switch to V1 tokens" as a secondary item (V2 is the default).
2. Navigate to `/features/3344`. Click the toggle. Styles flip to V1 immediately, no refresh needed. Click again to return to V2.
3. Navigate to `/features/3229`. Verify MenuButton renders correctly in both V1 and V2. This page used to carry a `:root`-rule-removal hack that has been removed.
4. Refresh after toggling. `sessionStorage` keeps your choice. Open a second tab to confirm it defaults to V2 independently.
5. Append `?tokens=v1` to any URL. Page loads in V1 regardless of sessionStorage.
6. Repeat steps 1-5 in Angular (`npm run serve:prs:angular`).

image
image

### DIFF
--- a/apps/prs/angular/src/main.ts
+++ b/apps/prs/angular/src/main.ts
@@ -1,6 +1,19 @@
 import { platformBrowserDynamic } from "@angular/platform-browser-dynamic";
 import { AppModule } from "./app/app.module";
+// This import has a side effect: token-version.ts calls applyTokenVersion at
+// module load, which puts the V2 stylesheet link in <head> before Angular
+// bootstraps. Without this, the page would flash V1 on first paint.
+import {
+  applyTokenVersion,
+  resolveTokenVersion,
+} from "./app/token-version/token-version";
 
 platformBrowserDynamic()
   .bootstrapModule(AppModule)
+  .then(() => {
+    // Re-apply after Angular's bundled styles.css is in <head>, so the V2
+    // link lands last in the cascade. Without this, the bundled V1 @import
+    // overrides V2 and the toggle appears to do nothing.
+    applyTokenVersion(resolveTokenVersion());
+  })
   .catch((err) => console.error(err));
--- a/apps/prs/angular/project.json
+++ b/apps/prs/angular/project.json
@@ -32,10 +32,7 @@
             "output": "/v2-tokens"
           }
         ],
-        "styles": [
-          "apps/prs/angular/src/styles.css",
-          "node_modules/@abgov/design-tokens-v2/dist/tokens.css"
-        ],
+        "styles": ["apps/prs/angular/src/styles.css"],
         "scripts": []
       },
       "configurations": {
--- a/apps/prs/angular/src/app/app.component.html
+++ b/apps/prs/angular/src/app/app.component.html
@@ -15,8 +15,17 @@
         [open]="true"
         (onNavigate)="handleNavigate($event)"
         [primaryContent]="primaryTemplate"
+        [secondaryContent]="tokenToggleTemplate"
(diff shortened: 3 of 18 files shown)


willcodeforcoffee (changes requested):
@twjeffery it doesn't seem to be working in Angular for me! 😭

twjeffery:
> @twjeffery it doesn't seem to be working in Angular for me! 😭

@willcodeforcoffee Good catch. V2 was loading, but Angular's V1 styles were coming in after and overriding it. Should be fixed now so V2 loads properly. Can you check again?

willcodeforcoffee (approved):
This works great now!

ArakTaiRoth:
@twjeffery Can you rebase this before I test? There's discrepancies around routing that are different here than they are in dev, and I'm not sure how this will work in that code.

apps/prs/angular/src/main.ts
```diff
@@ -1,6 +1,18 @@
 import { platformBrowserDynamic } from "@angular/platform-browser-dynamic";
 import { AppModule } from "./app/app.module";
+// Apply the selected token version (default V2) to <head> before Angular
```
vanessatran-ddi: Can you update this comment into something like this: "This import has a side effect: it initialize the V2 stylesheet link in <head> as the module loads."  When I first read it I was confused until I check the `token-version.ts` and I see thje function `applyTokenVersion(resolveTokenVersion());` was called when this is import, so update the comment to make it clearer.

vanessatran-ddi (approved):
Looks good!

ArakTaiRoth:
@twjeffery bug3625.tsx still loads a constant v2 token, so it would overwrite it if v1 tokens were selected.

twjeffery:
@ArakTaiRoth I removed the per-page V2 from bug3625.tsx. Should be under the toggle's control now.

PR #3861, opened 2026-04-22
fix(#3605): use :focus-visible across interactive components

### LINKED ISSUE #3605: Use focus-visible consistently across all interactive components
## Description

Interactive components should use `:focus-visible` instead of `:focus` for focus styling, so that focus indicators only appear during keyboard navigation and not on mouse click.

Currently some components show focus rings on mouse click, which causes two issues:
1. **Unnecessary focus rings on click** — visually noisy for mouse users
2. **Incorrect focus ring rendering on click** — the focus border has no gap/offset between the ring and the element when triggered by mouse, though it displays correctly when triggered by keyboard

Switching to `:focus-visible` resolves both issues since the mouse-triggered focus state simply won't render.

## Scope

This applies broadly across all interactive elements:
- Inputs (text, checkbox, radio, dropdown, date picker, etc.)
- Buttons (button, icon button, link button)
- Links
- Tabs
- Accordion triggers
- Table (interactive cells, sort headers)
- Data grid (cells, controls)
- Any other interactive element that shows a focus state

## Expected behavior

- **Keyboard navigation:** Focus ring visible (with correct gap/offset, already working)
- **Mouse click:** No focus ring shown

twjeffery:
Fixes #3605

# Before

Mouse click on interactive components left the focus ring visible for mouse users. In some cases the ring rendered without the designed gap/offset because `:focus` fired on the mouse path alongside `:focus-visible`.

# After

Mouse click no longer shows the focus ring. Keyboard Tab and arrow-key navigation show the ring with the correct gap/offset. V1 and V2 paths both behave correctly.

## Components updated

- Button.svelte
- Chip.svelte (V1, deprecated)
- Checkbox.svelte (V1)
- DataGrid.svelte
- Details.svelte
- FilterChip.svelte
- Link.svelte
- MicrositeHeader.svelte
- Notification.svelte
- TextArea.svelte
- Tooltip.svelte
- assets/css/reset.css

## Steps needed to test

1. Check out this branch and run `npm run serve:prs:react` (or `:angular` or `:web`).
2. Navigate to `/bugs/3605` (React/Angular) or `/issues/3605` (Web).
3. For each section, press Tab through the interactive elements. A focus ring should appear with the correct gap/offset.
4. Click the same elements with a mouse. No focus ring should remain after the click. Holding mouse down (`:active`) should still show the pressed/focused state on buttons.
5. For DataGrid: use arrow keys inside both the table example and the cards example. Focus ring appears on keyboard nav, not on mouse click. Cells containing interactive components (checkbox, button, menu-button) do not stack our ring on top of the component's native ring.

https://github.com/user-attachments/assets/d6fe2cd1-e431-45a1-86b5-a6a6628fedaa

https://github.com/user-attachments/assets/07a3560d-fdc7-4d42-b090-770606f25214

### DIFF
--- a/apps/prs/react/src/routes/bugs/bug3605.tsx
+++ b/apps/prs/react/src/routes/bugs/bug3605.tsx
@@ -1,0 +1,378 @@
+import "@abgov/style";
+import {
+  GoabBadge,
+  GoabBlock,
+  GoabButton,
+  GoabCheckbox,
+  GoabChip,
+  GoabContainer,
+  GoabDataGrid,
+  GoabDetails,
+  GoabDivider,
+  GoabFilterChip,
+  GoabLink,
+  GoabMenuAction,
+  GoabMenuButton,
+  GoabMicrositeHeader,
+  GoabNotification,
+  GoabTable,
+  GoabText,
+  GoabTextarea,
+  GoabTooltip,
+} from "@abgov/react-components";
+
+export function Bug3605Route() {
+  return (
+    <div style={{ padding: "2rem", display: "flex", flexDirection: "column", gap: "1rem", maxWidth: "72rem", margin: "0 auto" }}>
+      <GoabText tag="h5" mb="xs" color="secondary">
+        Issue 3605
+      </GoabText>
+      <GoabText tag="h1" mt="0" mb="0">
+        Use focus-visible across interactive components
+      </GoabText>
+
+      <GoabBlock>
+        <GoabLink trailingIcon="open">
+          <a href="https://github.com/GovAlta/ui-components/issues/3605" target="_blank" rel="noopener">
+            View on GitHub
+          </a>
+        </GoabLink>
+      </GoabBlock>
+
+      <GoabText mb="0">
+        For every section: Tab should show the focus ring with the correct gap and offset.
+        Click should not leave a lingering ring. Mouse-down on buttons should still show the
+        pressed state.
+      </GoabText>
+
+      <GoabDivider mt="l" mb="l" />
+
+      {/* MicrositeHeader */}
+      <section>
+        <GoabText tag="h2" mt="0" mb="xs">MicrositeHeader</GoabText>
+        <GoabText mb="s">Tab through the links inside the header. Click any link.</GoabText>
+        <GoabMicrositeHeader type="alpha" version="UAT" />
+      </section>
+
+      <GoabDivider mt="l" mb="l" />
+
+      {/* Chip + FilterChip */}
+      <section>
+        <GoabText tag="h2" mt="0" mb="xs">Chip and FilterChip</GoabText>
+        <GoabText tag="h4" mt="s" mb="xs">Chip (deprecated)</GoabText>
+        <GoabBlock gap="m" alignment="center">
+          <GoabChip content="Status: active" />
+          <GoabChip content="Priority: high" deletable />
+          <GoabChip content="Owner: Tom" />
+        </GoabBlock>
+
+        <GoabText tag="h4" mt="m" mb="xs">FilterChip</GoabText>
+        <GoabBlock gap="m" alignment="center">
+          <GoabFilterChip content="Status: active" />
+          <GoabFilterChip content="Priority: high" />
+          <GoabFilterChip content="Owner: Tom" />
+        </GoabBlock>
+      </section>
+
+      <GoabDivider mt="l" mb="l" />
+
+      {/* Plain link reset */}
+      <section>
+        <GoabText tag="h2" mt="0" mb="xs">Plain link reset</GoabText>
+        <GoabText mb="s">Unstyled anchors pick up the global reset from assets/css/reset.css.</GoabText>
+        <div style={{ display: "flex", flexWrap: "wrap", gap: "1rem" }}>
+          <a href="#first">First link</a>
+          <a href="#second">Second link</a>
+          <a href="#third">Third link</a>
+        </div>
+      </section>
+
+      <GoabDivider mt="l" mb="l" />
+
+      {/* TextArea error */}
+      <section>
+        <GoabText tag="h2" mt="0" mb="xs">TextArea error state</GoabText>
+        <GoabText mb="s">
+          Tab into the error text area. Click into it. The focus-within container pattern
+          applies on focus, which is correct for text inputs.
+        </GoabText>
+        <GoabTextarea name="err" error />
+      </section>
+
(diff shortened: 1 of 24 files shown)


apps/prs/react/src/routes/bugs/bug3605.tsx
```diff
@@ -0,0 +1,378 @@
+import "@abgov/style";
```
ArakTaiRoth, line 1: Only comment here, the React PR doesn't match the Angular PR, the Angular one includes a lot of v1 references, though I'm not sure they're explicitly necessary? Not sure what we want, just bringing it up in case you want to modify one or both of them.

ArakTaiRoth (approved):

libs/web-components/src/components/data-grid/DataGrid.svelte
```diff
@@ -46,6 +46,10 @@
   let _bindTimeoutId: any = null; // Track pending grid rebuilds to prevent race conditions
   let _isRebuilding = false; // Flag to prevent navigation during grid rebuild
   let _gridHasFocus = false; // Track if the grid currently has focus (persists through DOM mutations)
+  // Track last input type so cell outline only appears on keyboard-driven focus.
+  // Avoids browser :focus-visible heuristics around the first interaction on a fresh page,
+  // which can match on the first mouse click.
+  let _lastInputType: 'pointer' | 'keyboard' = 'keyboard';
```
vanessatran-ddi: Can we rename it to be `_lastInteractionMode` or `_lastInputModality` (https://www.w3.org/WAI/WCAG21/Understanding/input-modalities.html). Input type makes me think it is our `input` html or component.

libs/web-components/src/components/data-grid/DataGrid.svelte
```diff
@@ -73,6 +77,8 @@
         }
       }
       document.removeEventListener('click', handleOutsideClick);
+      document.removeEventListener('pointerdown', handlePointerdown, true);
```
vanessatran-ddi, line 80: function `handlePointerdown` and `handleGlobalKeydown` has only 1 code statement, we don't need to create 2 functions, we can add inline logic here, it will be easier to read.

libs/web-components/src/components/data-grid/DataGrid.svelte
```diff
@@ -437,13 +472,40 @@
   function setFocusedStyle(el: HTMLElement) {
     if (!el) return;
     el.style.outline = '3px solid var(--goa-color-interactive-focus)';
+
+    if (keyboardNav === 'table') {
```
vanessatran-ddi: Hi @twjeffery , I tested 2 cases below:
1. When there is a checkbox inside the table header:
I am not sure if this is a missing case, or it is as expected, you can see the first column, we don't have the left border radius because we must focus on `goa-checkbox` inside, while on the right hand side, the last column, we can see the border-radius highlighted. 

https://github.com/user-attachments/assets/58b9153e-9732-461b-a8a0-efe2f8cc67b3

2. When we have `goa-table-sort-header`: 

https://github.com/user-attachments/assets/a6060d89-4e80-4581-b0a8-4e161b72a5b9

libs/web-components/src/components/data-grid/DataGrid.svelte
```diff
@@ -437,13 +472,40 @@
...
+      // otherwise cells on the edges get clipped by the table's overflow.
+      el.style.outlineOffset = '-0.1875rem';
+
+      // Match the table's corner radius on corner cells so the inset outline
+      // follows the rounded table corner instead of looking square.
+      const row = _focusingRowIndex;
+      const col = _focusingColIndex;
+      const lastRow = _grid.length - 1;
+      const lastCol = _grid[row] ? _grid[row].length - 1 : 0;
+      const radius = 'var(--goa-table-border-radius-container, 16px)';
```
vanessatran-ddi: This function works for the current use cases but I want to flag a new concern about how the focus ring is applied here. 
1. `--goa-table-border-radius-container` is a table token but it is being applied here which is DataGrid. 
2. mixed concerns: datagrid is taking on responsibility for table-specific visual styling. Ideally datagrid only manages the focus state and delegates the visual to whoever owns the cell. 
3. Rely only on `keyboardNav` is not good enough, especially `keyboardNav` by default is `table`. 

Suggested approach:
we can move to a `data-focused` attribute pattern. so here, we only flip the attribute, and we can style it under Table, or css alongside vs the existing `goa-table td`.. (i know it is a tech debt that we have css inside the `components.css` but it should work too. 
For example:
the above can be shorten like:
`el.setAttribute('data-focused', '');`

and in table
`goa-table td[data-focused] {...}`

Combining vs the above cases for `goa-table-header`, if this requires a big refactoring I will suggest to remove this logic from this current pull request and make another pull request for the purpose of styling the table header inside DataGrid.

libs/web-components/src/components/data-grid/DataGrid.svelte
twjeffery: Renamed to `_lastInputModality`. Updated the comments too.

libs/web-components/src/components/data-grid/DataGrid.svelte
twjeffery: They need to be named functions because we pass the same reference to `removeEventListener` in cleanup. Inline arrows would create new references each time and leak listeners on `document`.

libs/web-components/src/components/data-grid/DataGrid.svelte
twjeffery: I think this is right as-is. When a focusable child takes focus, the cell stays bare and the child shows its own ring. Adding the corner to the cell would imply the cell itself is focused.

Also tightened the sort-header focus while in here. V2 was ringing just the icon. Now it wraps the label and icon.

libs/web-components/src/components/data-grid/DataGrid.svelte
twjeffery: Took your suggestion. DataGrid only flips `data-focused` and `data-focus-corner` attributes now. The visual rules live with the rest of the table styles in `components.css`, so DataGrid no longer reaches into table-specific tokens. Layout-mode cells got their own selector so they don't depend on the table either.

Also fixed the `GoabTableSortHeader` wrapper (React and Angular) to hardcode `version="2"` like the others. It was missing before, so V2 styling wasn't actually applying.

twjeffery:
Continued in #3911 (this PR was auto-closed when the branch was renamed) and couldn't reopen.

PR #3909, opened 2026-05-05
fix(#3762): AppHeader Slots

### LINKED ISSUE #3762: AppHeader examples using `slot` property
# Info

Reported by Alex T

Our AppHeader examples are using a `slot` property on some of our components, https://design.alberta.ca/components/app-header/#tab-1, specifically, Header with Navigation example.

This works fine in Angular because Angular correctly identifies `slot` as a global property available on all components. React however does not make this same identification and triggers an error on build, because the `slot` property isn't actually placed on the component in the wrapper.

This was discussed [here](https://github.com/GovAlta/ui-components/pull/3650#discussion_r3012452605) briefly in a PR.

## Workaround

There is a workaround for this right now. People can surround the component using the `slot` property with a `div` using that same property and all of this will work as intended.

## Acceptance Criteria:

1. Rewrite AppHeader to allow named properties that function as `slots` as our other components currently use
2. The examples may or may not need to be changed to reflect the new way

willcodeforcoffee:
# Before (the change)

Our AppHeader examples are using a `slot` property on some of our components, [design.alberta.ca/components/app-header#tab-1](https://design.alberta.ca/components/app-header/#tab-1), specifically, the _Header with Navigation_ example.
See:
1. https://design.alberta.ca/components/app-header/
2. https://design.alberta.ca/examples/header-with-navigation/

# After (the change)

Several changes have been made to make these controls more consistent with other controls.

- React: Added props for `banner`, `phase`, `navigation`, and `utilities` slots.
- Angular: Added template attributes for `banner`, `phase`, `navigation`, and `utilities` slots.
- Renamed Angular files from `header` to `app-header` and `header-menu` to `app-header-menu`
- All React and Angular examples updated to use appropriate prop/templates
- Other documentation updated as well

Backward compatibility with slots is maintained.

## Steps needed to test

- Note: There are some problems with the PRs page:
  - The PR page is not resizable so you can't test mobile versions
  - There seems to be an issue a separate issue where `window.innerWidth` increases for the different Headers on the PR page
  - Try using the docs page where you can http://localhost:4201/docs/app-header for testing different screen widths
- backward compatibility maintained
- props for React components work as expected
- templates for Angular components work as expected
- documentation is correct and makes sense


libs/react-components/src/lib/app-header/app-header.tsx
```diff
@@ -81,6 +93,10 @@ export function GoabAppHeader({
       version="2"
     >
       {children}
+      {banner && <div slot="banner">{banner}</div>}
+      {phase && <div slot="phase">{phase}</div>}
+      {navigation && <div slot="navigation">{navigation}</div>}
+      {utilities && <div slot="utilities">{utilities}</div>}
```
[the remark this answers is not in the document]
willcodeforcoffee: No, if I understand you right this is not how we want to do things.

willcodeforcoffee:
Copilot Feedback Addressed:

- Added unit tests for the slots in Angular and React components
- removed unreachable code in docs
- formatting and other stuff

libs/angular-components/src/lib/components/app-header/app-header.ts
```diff
@@ -46,6 +69,14 @@ export class GoabAppHeader implements OnInit {
   @Input({ transform: numberAttribute }) fullMenuBreakpoint?: number;
   /** Sets a data-testid attribute for automated testing. */
   @Input() testId?: string;
+  /** V2 only: Banner content displayed above the header. */
+  @Input() banner?: TemplateRef<any>;
+  /** V2 only: Phase badge content displayed beside the service name. */
+  @Input() phase?: TemplateRef<any>;
+  /** Links and app header menus appear in the navigation bar below the header. Use plain links for single items and app header menu for grouped items with a dropdown. */
```
vanessatran-ddi, line 76: I see `banner` and `phase` has a comment `V2 Only` should we do the same for `navigation` and `utilities` since it is for V2 only too?

libs/angular-components/src/lib/components/app-header/app-header.spec.ts
```diff
@@ -137,6 +137,112 @@ describe("GoabAppHeader", () => {
...
+      schemas: [CUSTOM_ELEMENTS_SCHEMA],
+    }).compileComponents();
+
+    const fixture = TestBed.createComponent(TestAppHeaderSlotsComponent);
+    fixture.detectChanges();
+    tick();
+    fixture.detectChanges();
+
+    const el = fixture.debugElement.query(By.css("goa-app-header")).nativeElement;
+    expect(el.querySelector("div[slot='banner']")).not.toBeNull();
```
vanessatran-ddi: Can we verify the content inside slot instead of just verify it is not null? The pattern we can follow is `container.spec.ts`

libs/angular-components/src/lib/components/app-header/app-header.spec.ts
```diff
@@ -137,6 +137,112 @@ describe("GoabAppHeader", () => {
...
+@Component({
+  standalone: true,
+  imports: [GoabAppHeader],
+  template: `<goab-app-header heading="Test heading" />`,
+})
+class TestAppHeaderNoSlotsComponent {}
+
+describe("GoabAppHeader slot props", () => {
+  it("should render banner prop into a div with slot='banner'", fakeAsync(() => {
+    TestBed.configureTestingModule({
```
vanessatran-ddi, line 170: We can set up a `beforeEach` and then add `TestBed` instead of having it repeated on every test on this file, follow `container.spec.ts` for example

libs/react-components/src/lib/app-header/app-header.tsx
```diff
@@ -37,18 +37,30 @@ export interface GoabAppHeaderProps extends DataAttributes {
...
+  children?: ReactNode;
   /** Callback fired when the menu button is clicked. When provided, clicking the menu button dispatches a custom event instead of toggling the menu. */
   onMenuClick?: () => void;
   /** Sets a data-testid attribute for automated testing. */
   testId?: string;
+  /** V2 only: Banner content displayed above the header. */
+  banner?: ReactNode;
+  /** V2 only: Phase badge content displayed beside the service name. */
+  phase?: ReactNode;
+  /** Links and app header menus appear in the navigation bar below the header. Use plain links for single items and app header menu for grouped items with a dropdown. */
```
vanessatran-ddi, line 49: `V2 Only` comment is for `banner` and `phase` only, should we add it the comment for `navigation` and `utitlities` too?

apps/prs/angular/src/routes/bugs/3762/bug3762.component.html
```diff
@@ -0,0 +1,115 @@
...
+  </goab-column-layout>
+</goab-container>
+<ng-template #heroActions>
+  <goab-button type="primary">Get started</goab-button>
+  <goab-button type="secondary">Secondary CTA</goab-button>
+</ng-template>
+<ng-template #navigationTemplate>
+  <goab-block direction="row" gap="s">
+    <a href="#">Dashboard</a>
+    <goab-app-header-menu heading="Applications">
```
vanessatran-ddi: Image

The alignment for App Header Menu doesn't look correct. Do we have a ticket to address it? Or in this slot, we shouldn't use app header menu? 
cc @twjeffery  for your input  Example Link: https://govalta.github.io/ui-components/pr-preview-angular/pr-3909/bugs/3762

vanessatran-ddi: I think it is a bug @willcodeforcoffee I am using production package https://stackblitz.com/github/vanessatran-ddi/ui-components-angular-sandbox/tree/ds2.0?file=src%2Fapp%2Fcomponents%2Fapp-header%2Fapp-header-examples.component.html 
And the Applications aligned nicely 

Image
The problem is we need to use the `slotName` under `goab-app-header-menu`. If that is the case, can you help update this angular/react example and update this docs to use `slotName` for `GoabAppHeaderMenu`, or we need to fix it to make the code on this doc works. 

Image

apps/prs/react/src/routes/bugs/bug3762.tsx
```diff
@@ -0,0 +1,160 @@
...
+      </div>
+      <GoabContainer type="interactive" padding="relaxed">
+        <GoabText tag="h3" size="heading-s">
+          Site chrome
+        </GoabText>
+        <GoabOneColumnLayout>
+          <GoabAppHeader
+            heading="Sample App"
+            navigation={navigationSlot}
+            utilities={utilitiesSlot}
```
vanessatran-ddi: Current, there is no spacing between navigation and the app header
Image 


while on production there is a spacing (padding)

Image

apps/prs/react/src/routes/docs/app-header/app-header.tsx
```diff
@@ -26,62 +26,69 @@ export function DocsAppHeaderRoute() {
...
-          <a href="#">Archived</a>
-        </GoabAppHeaderMenu>
-        <a slot="navigation" href="#">
-          Reports
-        </a>
-        <a slot="navigation" href="#">
-          Settings
-        </a>
-        <div slot="utilities">
+      <GoabAppHeader
```
vanessatran-ddi, line 71: It looks different between `dev` and this current PR. Current PR makes the navigation collapsed like a side menu 

Dev: For example https://govalta.github.io/ui-components/pr-preview-react/pr-3955/docs/app-header 

Image

Current: https://govalta.github.io/ui-components/pr-preview-react/pr-3909/docs/app-header 

Image

libs/angular-components/src/lib/components/app-header/app-header.ts
willcodeforcoffee: Since these comments are used for the v2 documentation I just removed the "V2 only" comments from all of these props.

libs/react-components/src/lib/app-header/app-header.tsx
willcodeforcoffee: As above, the "V2" portion of this comment is removed.

apps/prs/angular/src/routes/bugs/3762/bug3762.component.html
willcodeforcoffee: @vanessatran-ddi you're right, this was a bug. I had to add some functionality to switch the DOM around. 😬
There was actually another bug where Tom had put in logic to wrap a longer menu into a "More" popover also. That wouldn't have worked without the DOM switcheroo.

On large screens everything fits:

image

On smaller screens (faked with Dev Tools) the overflow goes into a "More" item with dropdown.
image

apps/prs/react/src/routes/bugs/bug3762.tsx
willcodeforcoffee: This is fixed.

apps/prs/react/src/routes/docs/app-header/app-header.tsx
willcodeforcoffee: @vanessatran-ddi this was a good catch. I only had one item in the `utilities` slot so it didn't appear in my test page. I wrapped things in a Block with gap to make them look better.
I also updated the test page and docs to show it is working.

willcodeforcoffee:
Updated the PR with the following:

- Removed "V2 only" from all prop comments, they're not necessary anymore
- Updated tests to use `beforeEach` 
- Testing for content within slots instead of `not.null`
- Programatically moved slotted content up in the DOM so that the Navigation Menu can calculate menu overflow, also fixed how the menu appears, but also works as expected in other React and Angular components
- Fixed `utilities` slot examples in docs since containing `div` is gone

apps/prs/angular/src/routes/bugs/3762/bug3762.component.html
```diff
@@ -0,0 +1,120 @@
...
+    </goab-block>
+  </goab-column-layout>
+</goab-container>
+<ng-template #heroActions>
+  <goab-button type="primary">Get started</goab-button>
+  <goab-button type="secondary">Secondary CTA</goab-button>
+</ng-template>
+<ng-template #navigationTemplate>
+  <a href="#">Dashboard</a>
+  <goab-app-header-menu heading="Applications">
```
vanessatran-ddi, line 18: Current: https://govalta.github.io/ui-components/pr-preview-angular/pr-3909/bugs/3762

Image

The Current Application Menu is highlighted and mis-aligned. I compare vs design system website, 

Image

apps/prs/react/src/routes/docs/app-header/app-header.tsx
```diff
@@ -26,62 +26,69 @@ export function DocsAppHeaderRoute() {
...
-        <div slot="utilities">
-          <GoabButton type="tertiary" size="compact" leadingIcon="person">
-            Sign in
-          </GoabButton>
-        </div>
-      </GoabAppHeader>
+      <GoabAppHeader
+        heading="My Application"
+        url="/"
+        utilities={
```
vanessatran-ddi, line 33: It looks different. 
Now: 

Image

apps/prs/react/src/routes/docs/app-header/app-header.tsx
```diff
@@ -26,62 +26,69 @@ export function DocsAppHeaderRoute() {
...
-        <a slot="navigation" href="#">
-          Settings
-        </a>
-        <div slot="utilities">
+      <GoabAppHeader
+        heading="Service Portal"
+        url="/"
+        navigation={
+          <>
+            <a href="#">Dashboard</a>
```
vanessatran-ddi, line 76: There is a difference between docs and react when put under mobile width, for Navigation example:

On mobile: 

Image

(Some menu items will be put under "More") 

But on our React apps:

Image

apps/prs/react/src/routes/docs/app-header/app-header.tsx
```diff
@@ -26,62 +26,69 @@ export function DocsAppHeaderRoute() {
...
-        <div slot="utilities">
-          <GoabButton type="tertiary" size="compact">Help</GoabButton>
-        </div>
-        <div slot="utilities">
-          <GoabButton type="tertiary" size="compact" leadingIcon="person">
-            Sign in
-          </GoabButton>
-        </div>
-      </GoabAppHeader>
+      <GoabAppHeader
```
vanessatran-ddi, line 30: There is a difference between docs and React example (even the same code snippet), on mobile:

Image
It is collapsed 

But on our current React

Image

Can you check if the wrapper makes this logic on mobile breaks, or the docs is already wrong?

apps/prs/react/src/routes/docs/app-header/app-header.tsx
willcodeforcoffee: These are consistent now. 👍

apps/prs/react/src/routes/docs/app-header/app-header.tsx
willcodeforcoffee: The correct behaviour is the top. where it shows "Menu".
This works as expected now.

willcodeforcoffee:
I've made the following changes:
- made a fix to adding items in the navigation slot
- made a fix to items in the utilities slot
- made a fix to how the menu shows on the Angular component
- squashed onto `dev` so hopefully specs pass in CI

apps/prs/react/src/routes/docs/app-header/app-header.tsx
vanessatran-ddi: I checked again (both local and the github link) https://govalta.github.io/ui-components/pr-preview-angular/pr-3909/docs/app-header So when I make it mobile, now I see like this: 
image

In the docs I see like this: 
https://govalta.github.io/ui-components/pr-preview/pr-3909/components/app-header/#react#properties
Using the same code snippet 
image

I am good with what we have now on React and Angular, but I would like to ask if we should align the doc too, or whether there is a difference between web component (our Astro doc is using `goa-app-header`) and angular/react wrappers.

apps/prs/angular/src/routes/bugs/3762/bug3762.component.ts
```diff
@@ -0,0 +1,44 @@
...
+} from "@abgov/angular-components";
+import { CommonModule } from "@angular/common";
+import { Component } from "@angular/core";
+
+@Component({
+  standalone: true,
+  selector: "abgov-bug3762",
+  templateUrl: "./bug3762.component.html",
+  imports: [
+    CommonModule,
```
vanessatran-ddi: Do we need to import `CommonModule`?

apps/prs/angular/src/routes/docs/app-header/app-header.component.ts
```diff
@@ -16,6 +17,7 @@ import {
     GoabAppHeader,
     GoabAppHeaderMenu,
     GoabBadge,
+    GoabBlock,
```
vanessatran-ddi: We can remove import `GoabBlock` here

libs/angular-components/src/lib/components/app-header-menu/app-header-menu.ts
```diff
@@ -17,10 +17,12 @@ import {
...
       [attr.testid]="testId"
+      version="2"
     >
       <ng-content />
     </goa-app-header-menu>
   }`,
+
   schemas: [CUSTOM_ELEMENTS_SCHEMA],
 })
 /** Menu items within the app header. */
```
vanessatran-ddi, line 28: Do we still need `slotName` attribute line 39 below?

libs/web-components/src/common/utils.ts
```diff
@@ -613,3 +613,32 @@ export function parseCssTimeToMilliseconds(
...
+  slotName: string,
+): void {
+  const hostElement = hostEl.getRootNode() as ShadowRoot;
+  const lightDomChildren = hostElement.host?.children || [];
+
+  Array.from(lightDomChildren).forEach((el) => {
+    if (el.getAttribute("slot") === slotName && el.tagName === "DIV") {
+      const divFromWrapper = el;
+
+      Array.from(divFromWrapper.children)
```
vanessatran-ddi, line 631: Can be short like this: 
```
Array.from(divFromWrapper.children).forEach((child) => {
    divFromWrapper.removeChild(child);
    child.setAttribute("slot", slotName);
    hostElement.host.appendChild(child);
  });
```

libs/web-components/src/common/utils.ts
```diff
@@ -613,3 +613,32 @@ export function parseCssTimeToMilliseconds(
 
   return durationInMs;
 }
+
+export function restructureSlotContentOntoHost(
+  hostEl: HTMLElement,
+  slotName: string,
+): void {
+  const hostElement = hostEl.getRootNode() as ShadowRoot;
```
vanessatran-ddi: Picky: we should name this `hostElement` as `shadowElement` (hostElement should be `shadowElement.host`)

libs/web-components/src/components/app-header/AppHeader.svelte
```diff
@@ -98,6 +98,7 @@
     // V2-specific initialization
     if (version === "2") {
       detectV2Slots();
+      restructureUtilitiesSlotItems();
```
vanessatran-ddi: We can put the inline logic here, dont need the function `restructureUtitiliesSlotItems` because when I read this, I was having a question why we do not put the logic `structureUtilitiesSlotItems` inside function `detectUtilitiesItems()` which is also another Utilities function, it is a bit harder to read

libs/web-components/src/components/app-header-navigation/AppHeaderNavigation.svelte
```diff
@@ -27,6 +32,16 @@
 
   // Navigation Functions - Batch 1: Core Detection (3 functions)
 
+  // Move all items from the div[slot="navigation"] in the React and Angular
+  // wrapper up to the host level to enable menu overflow processing and styling.
+  function restructureNavigationSlotItems() {
+    if (!_navigationPlaceholderEl || version !== "2") return;
+
+    const slotName = "navigation";
```
vanessatran-ddi: We do not need this seperate function, perhaps just add 1 line `restructureSlotContentOntoHost(_navigationPlaceholderEl, \"navigation\" under `onMount`

libs/web-components/src/common/utils.ts
```diff
@@ -613,3 +613,32 @@ export function parseCssTimeToMilliseconds(
 
   return durationInMs;
 }
+
+export function restructureSlotContentOntoHost(
```
vanessatran-ddi: I suggest to rename this function to `unwrapSlotWrapperDiv` or `flattenWrapperDivIntoSlot` so it is easier to understand, restructure is a bit ambiguous

libs/angular-components/src/lib/components/app-header-menu/app-header-menu.ts
willcodeforcoffee: @vanessatran-ddi If we take it away we lose backward compatibility. What do you think if I were to deprecate it instead?

libs/web-components/src/components/app-header/AppHeader.svelte
willcodeforcoffee: I thought maybe wrapping it in a method would make it easier. Since that isn't the case I've unwrapped it.

libs/web-components/src/common/utils.ts
willcodeforcoffee: How does `unwrapNestedSlotContent` sound?
I also added a description comment.

willcodeforcoffee:
Addressed your feedback @vanessatran-ddi , including:

- fixed the docs differences
- fixed the padding around buttons on Angular to match React
- Renamed some variables and functions

I did not remove the `slotName` from `AppHeaderNavigation` because some users might still be using that. I don't think we want to break backward compatibility yet?

willcodeforcoffee:
Updates:
- deprecated the `slotName` attribute on AppHeaderMenu React and Angular components
- added a demo of combining the two in the PRs pages

vanessatran-ddi (approved):
Looks good!

docs/generated/component-apis/app-header.json
```diff
@@ -30,7 +30,7 @@
           "type": "string",
           "required": false,
           "default": null,
-          "description": "V2 only: Secondary text displayed under the service name."
+          "description": "Secondary text displayed under the service name."
```
vanessatran-ddi, line 33: I run `npm run build:prod` local, and I see `app-header-menu.json` on my local has a change, 

Image

Image

I don't understand why this PR doesn't commit this `app-header-menu.json` but I can see the `slotName` property be removed from preview link: https://govalta.github.io/ui-components/pr-preview/pr-3909/components/app-header/#react 

Image

Looks like even we don't commit, the script run magically somewhere? cc @twjeffery  or @bdfranck

twjeffery (commented):
@willcodeforcoffee the new props are a nice improvement over the raw `slot` attribute.

One thing I ran into while testing: if the header's content changes after it's first on screen, it breaks. The clearest case is signing in. Signed out shows a "Sign in" button, the person signs in, and you swap in their profile menu. When that swap happens, React goes fully blank (a `removeChild` error in the console), and Angular doesn't crash but the utilities just disappear, the Sign in goes and the profile never shows.

From what I can tell it's because the slot content gets moved up and the wrapper div removed when the header first loads, so the frameworks lose track of those nodes when they try to change them later. A header that's set up once and never changes is totally fine, which is everything the examples cover today. But anything that swaps header content on sign-in would hit this.

I noticed Copilot suggested putting the `slot` on each item instead of wrapping them, and you didn't want to go that way at the time. I tried it for this sign-in case though, and it's what stops the crash, the swap works fine and the profile shows up. Maybe worth another look now that we know it can crash? Happy to dig into it together.

libs/web-components/src/common/utils.ts
```diff
@@ -613,3 +613,34 @@ export function parseCssTimeToMilliseconds(
...
+        .forEach((child) => {
+          // Move all children of the div up to the host level slot
+          divFromWrapper.removeChild(child);
+
+          child.setAttribute("slot", slotName);
+
+          shadowRootEl.host.appendChild(child);
+        });
+
+      shadowRootEl.host.removeChild(divFromWrapper);
```
ArakTaiRoth, line 643: I gotta be honest, this seems really really gross to me, and is actually a problem, that can be seen on the very PR bug page that was created for this purpose (on Angular). Essentially what I think is happening here is this:
1. Inside the wrapper, `navigation` and `utilities` is created inside framework owned divs (`<div slot="utilities">`)
2. But after the mount, this function moves it to `<goab-block slot="utilities">`

But now, on any updates after this, the framework (React or Angular) still think it owns that wrapper, but they've actually been moved. So they can't actually make any changes if necessary inside those divs. Like I said, you can see this on the associated bug3762 PR in Angular. Just click the Sign In button in the very first example, that whole slot disappears. I can explain why this doesn't happen in React if you want, just ask.

I think the better option to doing all this mutating of the DOM (which again, really really gross), is to see if we can assign `slot="navigation"` or `slot="utilities"` to all children elements before render. That way they can stay in the same area, and will still be able to be manipulated later. Instead of this, and instead of wrapping them in a `<div slot="utilities">`.

docs/src/data/configurations/app-header-menu.ts
```diff
@@ -17,22 +17,27 @@ export const appHeaderMenuConfigurations: ComponentConfigurations = {
...
+  heading="My Application"
+  navigation={
+    <GoabAppHeaderMenu heading="Applications">
+      <a href="/dashboard">Dashboard</a>
+      <a href="/reports">Reports</a>
+      <a href="/settings">Settings</a>
+    </GoabAppHeaderMenu>
+  }
+/>`,
+        angular: `<goab-app-header
```
ArakTaiRoth, line 30: This angular code example doesn't even actually use `goab-app-header-menu`

libs/angular-components/src/lib/components/app-header-menu/app-header-menu.ts
```diff
@@ -33,7 +35,7 @@ export class GoabAppHeaderMenu implements OnInit {
   @Input() heading?: string;
   /** Sets a data-testid attribute for automated testing. */
   @Input() testId?: string;
-  /** Sets the slot name for the component. */
+  /** @deprecated Sets the slot name for the component. Deprecated: use AppHeader navigation temnplateRef instead. */
```
ArakTaiRoth, line 38: `templateRef` should be changed to `TemplateRef`

apps/prs/angular/src/routes/bugs/3762/bug3762.component.html
```diff
@@ -0,0 +1,306 @@
...
+</ng-template>
+
+<ng-template #signedInUtilitiesSlot>
+  <goab-block alignment="center" direction="row" gap="m">
+    <goab-button type="tertiary" size="compact"> Help </goab-button>
+    <goab-menu-button
+      text="John Smith"
+      type="tertiary"
+      size="compact"
+      (action)="signOut()"
```
ArakTaiRoth, line 49: `(onAction)`

apps/prs/angular/src/routes/bugs/3762/bug3762.component.html
```diff
@@ -0,0 +1,306 @@
...
+      <goab-menu-action text="User settings" action="user-settings"></goab-menu-action>
+      <goab-menu-action text="Sign out" action="sign-out"></goab-menu-action>
+    </goab-menu-button>
+  </goab-block>
+</ng-template>
+
+<ng-template #signedOutUtilitiesSlot>
+  <goab-block alignment="center" direction="row" gap="m">
+    <goab-button type="tertiary" size="compact"> Help </goab-button>
+    <goab-button type="tertiary" size="compact" leadingIcon="person" (click)="signIn()">
```
ArakTaiRoth, line 60: `(onClick)`

Issue #159, opened 2021-10-22
Destructive button


twjeffery:
# What is being submitted?

# Why are you submitting?

# Examples, links, and research
Here is an example from a document management screen we have for CCM. The ‘remove’ document button triggers a confirmation modal so it’s not a one click destructive action. Therefore, the red seemed a bit too aggressive and I decided to keep the remove button blue for now.

![image](https://user-images.githubusercontent.com/82968683/138529955-da7e39b4-b3b2-44a9-b072-2157223996ae.png)

### Advanced education:
![image](https://user-images.githubusercontent.com/82968683/138529986-fb681958-3050-427e-b54c-c2c2df759cb9.png)

### Inline: 
![image](https://user-images.githubusercontent.com/82968683/138530010-3e5c4b67-7f01-4712-8c30-cd61f831c586.png)

### Delete and archive buttons:
![image](https://user-images.githubusercontent.com/82968683/138530029-afed4b32-1529-43d2-a4b2-2e2f3926a4d3.png)

### Confirmation modal:
![image](https://user-images.githubusercontent.com/82968683/138530058-548df1c3-92c0-406f-9aff-381abf206b93.png)

### Delete user example:
![image](https://user-images.githubusercontent.com/82968683/138530084-7ad3257d-c8a6-464d-95ea-935811da8dfb.png)

# Anything else
Related to [button component](https://github.com/GovAlta/ui-components/projects/4#card-76433764)

PR #3911, opened 2026-05-05
fix(#3605): use :focus-visible across interactive components

### LINKED ISSUE #3605: Use focus-visible consistently across all interactive components

twjeffery:
Fixes #3605

Replaces #3861 (auto-closed when the branch was renamed.)

# Before

Mouse click on interactive components left the focus ring visible for mouse users. In some cases the ring rendered without the designed gap/offset because `:focus` fired on the mouse path alongside `:focus-visible`.

# After

Mouse click no longer shows the focus ring. Keyboard Tab and arrow-key navigation show the ring with the correct gap/offset. V1 and V2 paths both behave correctly.

## Components updated

- Button.svelte
- Chip.svelte (V1, deprecated)
- Checkbox.svelte (V1)
- DataGrid.svelte
- Details.svelte
- FilterChip.svelte
- Link.svelte
- MicrositeHeader.svelte
- Notification.svelte
- TextArea.svelte
- Tooltip.svelte
- assets/css/reset.css

## Steps needed to test

1. Check out this branch and run `npm run serve:prs:react` (or `:angular` or `:web`).
2. Navigate to `/bugs/3605` (React/Angular) or `/issues/3605` (Web).
3. For each section, press Tab through the interactive elements. A focus ring should appear with the correct gap/offset.
4. Click the same elements with a mouse. No focus ring should remain after the click. Holding mouse down (`:active`) should still show the pressed/focused state on buttons.
5. For DataGrid: use arrow keys inside both the table example and the cards example. Focus ring appears on keyboard nav, not on mouse click. Cells containing interactive components (checkbox, button, menu-button) do not stack our ring on top of the component's native ring.

https://github.com/user-attachments/assets/d6fe2cd1-e431-45a1-86b5-a6a6628fedaa

https://github.com/user-attachments/assets/07a3560d-fdc7-4d42-b090-770606f25214

### DIFF
--- a/apps/prs/angular/src/routes/bugs/3605/bug3605.component.html
+++ b/apps/prs/angular/src/routes/bugs/3605/bug3605.component.html
@@ -1,0 +1,447 @@
+<div style="padding: 2rem; display: flex; flex-direction: column; gap: 1rem; max-width: 72rem; margin: 0 auto;">
+  <goab-text tag="h5" mb="xs" color="secondary">Issue 3605</goab-text>
+  <goab-text tag="h1" mt="0" mb="0">
+    Use <code>:focus-visible</code> across interactive components
+  </goab-text>
+
+  <goab-block>
+    <goab-link version="2" trailingIcon="open">
+      <a href="https://github.com/GovAlta/ui-components/issues/3605" target="_blank" rel="noopener">
+        View on GitHub
+      </a>
+    </goab-link>
+  </goab-block>
+
+  <goab-text mb="0">
+    For every section: <strong>Tab</strong> should show the focus ring with the correct gap/offset.
+    <strong>Click</strong> should not leave a lingering ring. <strong>Mouse-down (<code>:active</code>)</strong>
+    on buttons should still show the pressed/focused look.
+  </goab-text>
+
+  <goab-divider mt="l" mb="l"></goab-divider>
+
+  <!-- 1. MicrositeHeader -->
+  <section>
+    <goab-text tag="h2" mt="0" mb="xs">MicrositeHeader</goab-text>
+    <goab-text mb="s">Tab through the links inside the header. Click any link.</goab-text>
+    <goab-microsite-header type="beta" version="2" headerUrl="https://www.alberta.ca" feedbackUrl="#"></goab-microsite-header>
+  </section>
+
+  <goab-divider mt="l" mb="l"></goab-divider>
+
+  <!-- 2. Chip + FilterChip -->
+  <section>
+    <goab-text tag="h2" mt="0" mb="xs">Chip and FilterChip</goab-text>
+    <goab-text tag="h4" mt="s" mb="xs">Chip (deprecated)</goab-text>
+    <goab-block gap="m" alignment="center">
+      <goab-chip content="Status: active"></goab-chip>
+      <goab-chip content="Priority: high" [deletable]="true"></goab-chip>
+      <goab-chip content="Owner: Tom"></goab-chip>
+    </goab-block>
+
+    <goab-text tag="h4" mt="m" mb="xs">FilterChip</goab-text>
+    <goab-block gap="m" alignment="center">
+      <goab-filter-chip content="Status: active" version="2"></goab-filter-chip>
+      <goab-filter-chip content="Priority: high" version="2"></goab-filter-chip>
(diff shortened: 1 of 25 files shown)


vanessatran-ddi (approved):
Looks good!

apps/prs/web/src/routes/3605.svelte
```diff
@@ -0,0 +1,478 @@
...
+  </section>
+
+  <hr />
+
+  <!-- 7. Link -->
+  <section>
+    <goa-text as="h2" mt="0" mb="xs">7. Link — <code>::slotted(a:focus-visible)</code></goa-text>
+    <goa-text mb="s">
+      Ring is now on the slotted anchor directly (reliable across shadow DOM).
+      <strong>Trade-off:</strong> ring wraps the anchor text only, not the icons.
```
ArakTaiRoth: You point this out here, and it is true, focus on a link with trailing or leading icons, the focus border only goes around the text, not the icons. Do we want that?

If we don't, a possible solution could be to use the `focusin` event for the Link, and set a variable if it's triggered, then set a class on the outer container if that variable is true. And finally set the border based on that class.

twjeffery: @ArakTaiRoth yea agreed, the focus ring looked off when the icons are outside it

I Went with your `focusin` approach:

1. The Link listens for `focusin` on the slotted anchor.
2. Checks `.matches(':focus-visible')` to gate on keyboard input.
3. Toggles a state class on the outer container.
4. CSS draws the ring there so it wraps icon + text together.

twjeffery:
@ArakTaiRoth FYI, I opened #159 to bring notification banner and details focus border thickness in line with the rest. It's independent of this PR, just related work surfaced while testing. Tagged you to review as you have this context

apps/prs/angular/src/routes/bugs/3605/bug3605.component.html
```diff
@@ -0,0 +1,445 @@
...
+  <goab-divider mt="l" mb="l"></goab-divider>
+
+  <!-- 7. Link -->
+  <section>
+    <goab-text tag="h2" mt="0" mb="xs">
+      7. Link — <code>::slotted(a:focus-visible)</code>
+    </goab-text>
+    <goab-text mb="s">
+      Ring is now on the slotted anchor directly (reliable across shadow DOM).
+      <strong>Trade-off:</strong> ring wraps the anchor text only, not the icons.
```
ArakTaiRoth: I'm trying to not comment on PR's, but this would have someone testing and looking for the wrong thing now that you made the change required to wrap the icons as well.

apps/prs/react/src/routes/bugs/bug3605.tsx
```diff
@@ -0,0 +1,483 @@
...
+        </div>
+      </section>
+
+      <GoabDivider mt="l" mb="l" />
+
+      {/* Link */}
+      <section>
+        <GoabText tag="h2" mt="0" mb="xs">Link</GoabText>
+        <GoabText mb="s">
+          Ring is on the slotted anchor directly (reliable across shadow DOM). Ring wraps
```
ArakTaiRoth: Same issue here

apps/prs/angular/src/routes/bugs/3605/bug3605.component.html
twjeffery: should be fixed now

ArakTaiRoth (approved):

PR #3933, opened 2026-05-12
feat(#3347): add goa-scroll-panel and goa-workspace-layout

### LINKED ISSUE #3347: Create reusable sticky header/footer component

vanessatran-ddi:
# Before (the change)

# After (the change)

## Steps needed to test

https://govalta.github.io/ui-components/pr-preview/pr-3933/examples/workspace-layout/
https://govalta.github.io/ui-components/pr-preview/pr-3933/components/scroll-panel/#react#properties

React Links: 
https://govalta.github.io/ui-components/pr-preview-react/pr-3933/features/3347
https://govalta.github.io/ui-components/pr-preview-react/pr-3933/features/3347-push

Angular Links: 
https://govalta.github.io/ui-components/pr-preview-angular/pr-3933/features/3347-push
https://govalta.github.io/ui-components/pr-preview-angular/pr-3933/features/3347


Spark450:
Here is some feedback on the docs pages, thanks for sharing @twjeffery:
**Workspace example**
- The ["card" data view](https://govalta.github.io/ui-components/pr-preview/pr-3933/examples/workspace/#card) image seems show the tab in an "in between" state 
Screenshot 2026-06-04 at 10 09 39 AM

**Workspace layout**
- On the [workspace layout page](https://govalta.github.io/ui-components/pr-preview/pr-3933/examples/workspace-layout/#:~:text=of%20selected%20items.-,Push%20drawer,-An%20optional%20drawer), when talking about the push drawer I think the point of emphasis we should make about it vs the modal drawer is that the user can still interact with the page content when the drawer is open. something like: (...so users can still interact with the main page content.)
- For the [considerations](https://govalta.github.io/ui-components/pr-preview/pr-3933/examples/workspace-layout/#:~:text=or%20GoabWorkspaceLayoutScrollStateService%20(Angular).-,Considerations,-Use%20the%20workspace) Assuming we update the our React and Angular examples would it make sense to add a point that lets the user know if they use the template as a starting point it comes included with the base functionality wired up? This may be super obvious and not worth mentioning though...

libs/web-components/src/components/modal/Modal.svelte
```diff
@@ -490,17 +446,10 @@
     }
 
     .modal-content {
-      margin: var(--goa-modal-content-margin, 0 -2rem);
-      padding: var(--goa-modal-content-padding, 0);
+      margin: 0;
+      padding: 0 var(--goa-modal-scrollable-padding-desktop, var(--goa-space-l));
```
vanessatran-ddi, line 449: @twjeffery  and @bdfranck  I also have a PR on https://github.com/GovAlta/design-tokens/pull/162/commits 
Since I replace the goa-scroll to goa-scroll-panel, I need to adjust the padding og the modal action to make it better, example below: https://govalta.github.io/ui-components/pr-preview-react/pr-3933/features/3347 

Image

libs/web-components/src/components/scroll-panel/ScrollPanel.svelte
```diff
@@ -0,0 +1,257 @@
...
+
+  /* When content is scrolled under the header: a hairline border defines the edge
+     crisply and an outset drop shadow falls below it for depth. A real border (not a
+     second box-shadow layer) sits on top of the shadow instead of merging into it,
+     and it's transparent at rest (see base rule) so it only shows when scrolled,
+     with no layout shift. The shadow is the system shallow-below token; the host
+     overflow:hidden clips any side bleed at the edge. */
+  .scroll-panel-header--shadow {
+    border-bottom-color: var(
+      --goa-scroll-panel-header-scroll-border,
```
vanessatran-ddi, line 226: I have this PR https://github.com/GovAlta/design-tokens/pull/162/changes/78c5311faf852ad78260038b4712b3a47e9f1f42 to support those tokens

ArakTaiRoth:
@vanessatran-ddi @twjeffery There's no image for the Scroll Panel in the All Components page.
image

libs/web-components/src/components/scroll-panel/ScrollPanel.svelte
```diff
@@ -0,0 +1,257 @@
...
+      state: _scrollState,
+      isScrollable: _isScrollable,
+    });
+  }
+
+  // *****
+  // Hooks
+  // *****
+  onMount(() => {
+    if (!isValidDimension(height)) {
```
ArakTaiRoth: There's a potential problem here, in the docs it says `height` can be "Any valid CSS height value (e.g. "400px", "100%", "100vh")". This statement is incorrect. The function `isValidDimension` fails to account for `calc()`, `dvh`, `lvh`, or `svh`, `min()`, or `clamp()`. If I use a technical valid value out of that list, it does actually still work, but states a console error incorrectly. I don't think the option is to allow those values in `isValidDimension`, because that would require far more testing wherever that function is used, but maybe we be more clear about what values are allowed in the documentation.

In addition, if I use an actual invalid value for `height`, it still proceeds to try to create the ScrollPanel, the check doesn't actually stop anything. This results in an area in the screen that you can't actually scroll down or up in, it prevents page scrolling, and takes up the full height of the content.

ArakTaiRoth: I pushed a test at the bottom of the React `feat3347.tsx` that shows the second problem.

docs/src/content/examples/workspace-layout/index.mdx
```diff
@@ -0,0 +1,218 @@
...
+import PreviewContainer from "../../../components/PreviewContainer.astro";
+import { CodeSnippet } from "../../../components/CodeSnippet";
+import { withBase } from "@/lib/base-url";
+
+<PreviewContainer
+  image="/images/examples/workspace-layout-basic.png"
+  alt="Complete workspace layout preview"
+  previewUrl="/examples/workspace-layout/basic/preview"
+  figmaUrl="https://www.figma.com/design/Jpy1Ea5qglwnp1SgGnagY9/%E2%9D%96-Component-library?node-id=58855-102956"
+  reactSourceUrl="https://github.com/GovAlta/ui-components/blob/vanessa/3347-sticky-header-footer/docs/src/content/examples/workspace-layout/basic/react.tsx"
```
ArakTaiRoth: This URL and the URL below it will cease to exist once this PR is merged in (because that deletes the PR).

bdfranck (commented):

| Browser | Drawer | Modal | Notification |
| :--- | :---: | :---: | :---: |
| Chrome | ✅ | ✅ | ✅ |
| Safari | ✅ | ✅ | ✅ |
| Firefox | ✅ | ✅ | ✅ |
| Safari mobile | ✅ | ❓  | ✅ |


@vanessatran-ddi It looks like the padding is missing for a Modal on mobile. Or maybe I have the wrong tokens?

Image

libs/web-components/src/components/scroll-panel/ScrollPanel.svelte
vanessatran-ddi: I fixed both. I swapped for `CSS.supports(\"height\", value)` so `calc()/dvh/min()/clamp()` validate correctly with no false error.Invalid values now fall back to a `:host` `height:100%` instead of collapsing to content height.

docs/src/content/examples/workspace-layout/index.mdx
vanessatran-ddi: This is resolved.

ArakTaiRoth:
> @vanessatran-ddi @twjeffery There's no image for the Scroll Panel in the All Components page. image

@vanessatran-ddi The above is still an issue, now even moreso because we added documentation for Workspace Layout, which also doesn't have an image. I know there are other components that also don't have an image, so maybe this is just the way we're going, but it does look sloppy. @twjeffery thoughts?

twjeffery:
> @vanessatran-ddi @twjeffery There's no image for the Scroll Panel in the All Components page. image

@ArakTaiRoth @vanessatran-ddi Good catch. Thumbnails come from Olu, and Push Drawer, Scroll Panel, and Workspace Layout all don't have thumbnail illustrations yet. I've created issue #4088 to track them so Olu can create them, and we'll swap them in when they're ready.

ArakTaiRoth (approved):

bdfranck (approved):

| Browser | Drawer | Modal | Notification |
| :--- | :---: | :---: | :---: |
| Chrome | ✅ | ✅ | ✅ |
| Safari | ✅ | ✅ | ✅ |
| Firefox | ✅ | ✅ | ✅ |
| Safari mobile | ✅ | ✅  | ✅ |

Looks good to me! 👍

PR #3936, opened 2026-05-12
feat(#1233): implement a MultiselectDropdown component by combining popover and checkboxlist

### LINKED ISSUE #1223: (needs design) Dropdown multi-select
1. What is being submitted?
This is an extension of the dropdown component and the combo box components. Multi-select allows a user to select multiple items from a dropdown.

[Example](https://material.angular.io/components/select/overview1.multiple-selection)

1. Why are you submitting?
- What evidence do you have that it’s needed by multiple services across government?
- What evidence do you have that it meets the needs of the users of those services?
- Have you checked that it doesn’t already exist in the DIO Design System?

1. Examples, links, and research
**Considerations**
- When should a designer use a dropdown vs combobox vs multiselect?
- When should a designer use a multi select vs traditional checkboxes?
- Is this a new component or a variant of an existing dropdown?
- Are there any accessibility considerations?
- This component will have combo box functionality
- What is the relationship between dropdown, multi-select, combo box pages (there is currently a drop down page and combo box page in the DS)
- Consider UI behaviour for “All” “None” selection

_What should the designer who picks up this task be aware of? Have other designers built similar components or are there other projects already using this component? Is this an iteration on an existing component?_

**Constraints and scope**
- Will include type-ahead functionality
- Will include filter functionality

1. Anything else
Related to [Dropdown component](https://github.com/GovAlta/ui-components/projects/41.card-80918349)
- Add any other context or screenshots about the feature request here.

### LINKED ISSUE #1233: Design System Website - Home Page
## Acceptance Criteria:

[Figma Design](https://www.figma.com/file/qWGWDvULAAZuVaDpXCRzqP/DS-Resource-Website?type=design&node-id=3348-504908&mode=design&t=AO0B9TMkIngk5Klf-4)

willcodeforcoffee:
This PR creates a dropdown multiselect list component  by combining the CheckboxList and Popover Components.

The only AC left in Story 1233 is a link to a Figma that doesn't exist anymore. 🤷

The Multiselect has documentation as part of the PR https://govalta.github.io/ui-components/pr-preview/pr-3936/components/dropdown-multiselect/#angular#properties

We've added these features:

- Filtering (try filtering by "lake" or "grand" in the PR page filter for Alberta Cities)
- Select All
  - when filtering it should only select all of the visible filtered items 
- Truncating overflow in the placeholder using the `labelFormat` attribute:
	- with ellipsis (`list`)
	- with text (10 items) (`count`)
- Compact size

This closes #1233.

## Steps needed to test

This is a brand new component. Testing pages are available in the React and Angular playground, as well as documentation pages. It would be great testing for you to add it to your pages as well.

### DIFF
--- a/apps/prs/react/src/routes/docs/dropdown-multiselect/dropdown-multiselect.tsx
+++ b/apps/prs/react/src/routes/docs/dropdown-multiselect/dropdown-multiselect.tsx
@@ -1,0 +1,129 @@
+import {
+  GoabDropdownItem,
+  GoabDropdownMultiselect,
+  GoabFormItem,
+} from "@abgov/react-components";
+import { useState } from "react";
+
+export function DocsDropdownMultiselectRoute() {
+  const [fruits, setFruits] = useState<string[]>([]);
+  const [services, setServices] = useState<string[]>([]);
+
+  return (
+    <div>
+      <h2>Dropdown</h2>
+
+      <h3>Basic example</h3>
+
+      <GoabFormItem label="Select fruits" mb="l">
+        <GoabDropdownMultiselect
+          name="fruits"
+          width="320px"
+          value={fruits}
+          onChange={(detail) => setFruits(detail.value)}
+        >
+          <GoabDropdownItem value="apple" label="Apple" />
+          <GoabDropdownItem value="banana" label="Banana" />
+          <GoabDropdownItem value="orange" label="Orange" />
+          <GoabDropdownItem value="pear" label="Pear" />
+        </GoabDropdownMultiselect>
+      </GoabFormItem>
+
+      <h3>Dropdown multiselect with placeholder text when no selection</h3>
+
+      <GoabFormItem label="Select services" mb="l">
+        <GoabDropdownMultiselect
+          name="services"
+          width="320px"
+          placeholder="Select one or more services"
+          value={services}
+          onChange={(detail) => setServices(detail.value)}
+        >
+          <GoabDropdownItem value="health" label="Health benefits" />
+          <GoabDropdownItem value="income" label="Income support" />
+          <GoabDropdownItem value="housing" label="Housing support" />
+        </GoabDropdownMultiselect>
+      </GoabFormItem>
+
+      <h3>With count label</h3>
+
+      <GoabFormItem label="Select fruits" mb="l">
+        <GoabDropdownMultiselect name="fruits" width="320px" labelFormat="count">
+          <GoabDropdownItem value="apple" label="Apple" />
+          <GoabDropdownItem value="banana" label="Banana" />
+          <GoabDropdownItem value="orange" label="Orange" />
+          <GoabDropdownItem value="pear" label="Pear" />
+        </GoabDropdownMultiselect>
+      </GoabFormItem>
+
+      <h3>With list label</h3>
+
+      <GoabFormItem label="Select fruits" mb="l">
+        <GoabDropdownMultiselect name="fruits" width="320px" labelFormat="list">
+          <GoabDropdownItem value="apple" label="Apple" />
+          <GoabDropdownItem value="banana" label="Banana" />
+          <GoabDropdownItem value="orange" label="Orange" />
+          <GoabDropdownItem value="pear" label="Pear" />
+        </GoabDropdownMultiselect>
+      </GoabFormItem>
+
+      <h3>Dropdown multiselect with search/filter capability</h3>
+
+      <GoabFormItem label="Which cities have you visited?" mb="l">
+        <GoabDropdownMultiselect
+          name="cities"
+          filterable={true}
+          maxHeight="400px"
+          width="320px"
+        >
+          <GoabDropdownItem value="calgary" label="Calgary" />
+          <GoabDropdownItem value="edmonton" label="Edmonton" />
+          <GoabDropdownItem value="red-deer" label="Red Deer" />
+          <GoabDropdownItem value="lethbridge" label="Lethbridge" />
+          <GoabDropdownItem value="medicine-hat" label="Medicine Hat" />
+          <GoabDropdownItem value="grande-prairie" label="Grande Prairie" />
+        </GoabDropdownMultiselect>
+      </GoabFormItem>
+
+      <h3>With select all</h3>
+
+      <GoabFormItem label="Select departments" mb="l">
+        <GoabDropdownMultiselect name="departments" showSelectAll={true} width="320px">
+          <GoabDropdownItem value="health" label="Health" />
+          <GoabDropdownItem value="education" label="Education" />
+          <GoabDropdownItem value="transportation" label="Transportation" />
+          <GoabDropdownItem value="justice" label="Justice" />
+        </GoabDropdownMultiselect>
+      </GoabFormItem>
+
+      <h3>Sizes and states</h3>
+
+      <GoabFormItem label="Default size" mb="l">
+        <GoabDropdownMultiselect name="sizeDefault" width="320px">
+          <GoabDropdownItem value="draft" label="Draft" />
+          <GoabDropdownItem value="review" label="In review" />
+          <GoabDropdownItem value="approved" label="Approved" />
+        </GoabDropdownMultiselect>
+      </GoabFormItem>
+      <GoabFormItem label="Compact size" mb="l">
+        <GoabDropdownMultiselect name="sizeCompact" size="compact" width="320px">
+          <GoabDropdownItem value="draft" label="Draft" />
+          <GoabDropdownItem value="review" label="In review" />
+          <GoabDropdownItem value="approved" label="Approved" />
+        </GoabDropdownMultiselect>
+      </GoabFormItem>
+      <GoabFormItem label="Disabled" mb="l">
+        <GoabDropdownMultiselect name="disabled" disabled width="320px">
+          <GoabDropdownItem value="opt1" label="Option 1" />
+          <GoabDropdownItem value="opt2" label="Option 2" />
+        </GoabDropdownMultiselect>
+      </GoabFormItem>
+      <GoabFormItem label="With error" error="Please select at least one option" mb="l">
+        <GoabDropdownMultiselect name="error" error width="320px">
+          <GoabDropdownItem value="opt1" label="Option 1" />
+          <GoabDropdownItem value="opt2" label="Option 2" />
+        </GoabDropdownMultiselect>
+      </GoabFormItem>
+    </div>
+  );
+}
--- a/docs/generated/component-apis/dropdown-multiselect.json
+++ b/docs/generated/component-apis/dropdown-multiselect.json
@@ -1,0 +1,470 @@
+{
+  "componentSlug": "dropdown-multiselect",
+  "extractedFrom": "libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte",
+  "frameworks": {
+    "react": {
+      "props": [
+        {
+          "name": "ariaLabel",
+          "type": "string",
+          "required": false,
+          "default": null,
+          "description": "Provides an accessible label when no visible label is associated."
+        },
+        {
+          "name": "ariaLabelledBy",
+          "type": "string",
+          "required": false,
+          "default": null,
+          "description": "References an external element that labels this component."
+        },
+        {
+          "name": "disabled",
+          "type": "boolean",
+          "required": false,
+          "default": "false",
+          "description": "Disables the component."
+        },
+        {
+          "name": "error",
+          "type": "boolean",
+          "required": false,
+          "default": "false",
+          "description": "Shows an error state."
+        },
+        {
+          "name": "filterable",
+          "type": "boolean",
+          "required": false,
+          "default": null,
+          "description": "Enables filtering of options by typing in the trigger."
+        },
+        {
+          "name": "labelFormat",
+          "type": "GoabDropdownMultiselectLabelFormatOptions",
+          "required": false,
+          "default": "list",
+          "description": "The display label format of the closed dropdown. When 'count' the display label shows only \"n items\" in the label, when 'list' it shows a comma separated list of selected item labels."
+        },
+        {
+          "name": "leadingIcon",
+          "type": "GoabIconType",
+          "required": false,
+          "default": null,
+          "description": "Icon shown to the left of the dropdown input."
+        },
+        {
+          "name": "maxHeight",
+          "type": "string",
+          "required": false,
+          "default": "276px",
+          "description": "Sets the maximum height of the dropdown content area."
+        },
+        {
+          "name": "mb",
+          "type": "Spacing",
+          "required": false,
+          "default": null,
+          "description": ""
+        },
+        {
+          "name": "ml",
+          "type": "Spacing",
+          "required": false,
+          "default": null,
+          "description": ""
+        },
+        {
+          "name": "mr",
+          "type": "Spacing",
+          "required": false,
+          "default": null,
+          "description": ""
+        },
+        {
+          "name": "mt",
+          "type": "Spacing",
+          "required": false,
+          "default": null,
+          "description": ""
+        },
+        {
+          "name": "name",
+          "type": "string",
+          "required": true,
+          "default": null,
+          "description": "Identifier for the group. Used in change events."
+        },
+        {
+          "name": "placeholder",
+          "type": "string",
+          "required": false,
+          "default": null,
+          "description": "Text shown when nothing is selected."
+        },
+        {
+          "name": "showSelectAll",
+          "type": "boolean",
+          "required": false,
+          "default": "false",
+          "description": "Shows a \"Select All\" checkbox at the top of the options list."
+        },
+        {
+          "name": "size",
+          "type": "GoabDropdownMultiselectSize",
+          "required": false,
+          "default": "default",
+          "description": "Sets the size variant."
+        },
+        {
+          "name": "testId",
+          "type": "string",
+          "required": false,
+          "default": null,
+          "description": "Sets a data-testid attribute for automated testing."
+        },
+        {
+          "name": "value",
+          "type": "string[]",
+          "required": false,
+          "default": null,
+          "description": "Array of currently selected checkbox value."
+        },
+        {
+          "name": "width",
+          "type": "string",
+          "required": false,
+          "default": null,
+          "description": "Sets a fixed width for the component and popover panel."
+        }
+      ],
+      "events": [
+        {
+          "name": "onChange",
+          "type": "(detail: GoabDropdownMultiselectOnChangeDetail) => void",
+          "description": "Callback fired when the selected value change.",
+          "frameworks": [
+            "react"
+          ]
+        }
+      ],
+      "slots": []
+    },
+    "angular": {
+      "props": [
+        {
+          "name": "ariaLabel",
+          "type": "string",
+          "required": false,
+          "default": null,
+          "description": "Provides an accessible label when no visible label is associated."
+        },
+        {
+          "name": "ariaLabelledBy",
+          "type": "string",
+          "required": false,
+          "default": null,
+          "description": "References an external element that labels this component."
+        },
+        {
+          "name": "disabled",
+          "type": "boolean",
+          "required": false,
+          "default": null,
+          "description": "Sets the disabled state for the control."
+        },
+        {
+          "name": "error",
+          "type": "boolean",
+          "required": false,
+          "default": null,
+          "description": "Sets the error state for the control."
+        },
+        {
+          "name": "filterable",
+          "type": "boolean",
+          "required": false,
+          "default": null,
+          "description": "Enables filtering of options by typing in the trigger."
+        },
+        {
+          "name": "id",
+          "type": "string",
+          "required": false,
+          "default": null,
+          "description": "Sets the id attribute of the underlying web component."
+        },
+        {
+          "name": "labelFormat",
+          "type": "GoabDropdownMultiselectLabelFormatOptions",
+          "required": false,
+          "default": "list",
+          "description": "The display label format of the closed dropdown. When 'count' the display label shows only \"n items\" in the label, when 'list' it shows a comma separated list of selected item labels."
+        },
+        {
+          "name": "leadingIcon",
+          "type": "GoabIconType",
+          "required": false,
+          "default": null,
+          "description": "Icon shown to the left of the dropdown input."
+        },
+        {
+          "name": "maxHeight",
+          "type": "string",
+          "required": false,
+          "default": "276px",
+          "description": "Sets the maximum height of the dropdown content area."
+        },
+        {
+          "name": "mb",
+          "type": "Spacing",
+          "required": false,
+          "default": null,
+          "description": "Sets the bottom margin spacing token."
+        },
+        {
+          "name": "ml",
+          "type": "Spacing",
+          "required": false,
+          "default": null,
+          "description": "Sets the left margin spacing token."
+        },
+        {
+          "name": "mr",
+          "type": "Spacing",
+          "required": false,
+          "default": null,
+          "description": "Sets the right margin spacing token."
+        },
+        {
+          "name": "mt",
+          "type": "Spacing",
+          "required": false,
+          "default": null,
+          "description": "Sets the top margin spacing token."
+        },
+        {
+          "name": "name",
+          "type": "string",
+          "required": true,
+          "default": null,
+          "description": "Identifier for the group. Used in change events."
+        },
+        {
+          "name": "placeholder",
+          "type": "string",
+          "required": false,
+          "default": null,
+          "description": "Text shown when nothing is selected."
+        },
+        {
+          "name": "showSelectAll",
+          "type": "boolean",
+          "required": false,
+          "default": "false",
+          "description": "Shows a \"Select All\" checkbox at the top of the options list."
+        },
+        {
+          "name": "size",
+          "type": "GoabDropdownMultiselectSize",
+          "required": false,
+          "default": "default",
+          "description": "Sets the size variant."
+        },
+        {
+          "name": "testId",
+          "type": "string",
+          "required": false,
+          "default": null,
+          "description": "Sets the data-testid attribute for automated testing."
+        },
+        {
+          "name": "value",
+          "type": "string[]",
+          "required": false,
+          "default": null,
+          "description": "Array of currently selected checkbox values."
+        },
+        {
+          "name": "width",
+          "type": "string",
+          "required": false,
+          "default": null,
+          "description": "Sets a fixed width for the component and popover panel."
+        }
+      ],
+      "events": [
+        {
+          "name": "onChange",
+          "type": "(event: GoabDropdownMultiselectOnChangeDetail) => void",
+          "description": "Emits when the selected value change.",
+          "frameworks": [
+            "angular"
+          ]
+        }
+      ],
+      "slots": []
+    },
+    "webComponents": {
+      "props": [
+        {
+          "name": "aria-label",
+          "type": "string",
+          "required": false,
+          "default": "",
+          "description": "Provides an accessible label when no visible label is associated."
+        },
+        {
+          "name": "aria-labelledby",
+          "type": "string",
+          "required": false,
+          "default": "",
+          "description": "References an external element that labels this component."
+        },
+        {
+          "name": "disabled",
+          "type": "boolean",
+          "required": false,
+          "default": "false",
+          "description": "Disables the component. @default false"
+        },
+        {
+          "name": "error",
+          "type": "boolean",
+          "required": false,
+          "default": "false",
+          "description": "Shows an error state. @default false"
+        },
+        {
+          "name": "filterable",
+          "type": "boolean",
+          "required": false,
+          "default": "false",
+          "description": "Enables filtering of options by typing in the trigger."
+        },
+        {
+          "name": "label-format",
+          "type": "\"count\" | \"list\"",
+          "values": [
+            "count",
+            "list"
+          ],
+          "required": false,
+          "default": "list",
+          "description": "The display label format of the closed dropdown. When 'count' the display label shows only \"n items\" in the label, when 'list' it shows a comma separated list of selected item labels. @default \"list\""
+        },
+        {
+          "name": "leading-icon",
+          "type": "GoabIconType",
+          "typeLabel": "GoabIconType",
+          "required": false,
+          "default": null,
+          "description": "Icon shown to the left of the dropdown input. When \"filterable\" is true this can only be \"search\"."
+        },
+        {
+          "name": "max-height",
+          "type": "string",
+          "required": false,
+          "default": "276px",
+          "description": "Sets the maximum height of the dropdown content area. @default \"276px\""
+        },
+        {
+          "name": "mb",
+          "type": "Spacing",
+          "typeLabel": "Spacing",
+          "required": false,
+          "default": null,
+          "description": "Bottom margin."
+        },
+        {
+          "name": "ml",
+          "type": "Spacing",
+          "typeLabel": "Spacing",
+          "required": false,
+          "default": null,
+          "description": "Left margin."
+        },
+        {
+          "name": "mr",
+          "type": "Spacing",
+          "typeLabel": "Spacing",
+          "required": false,
+          "default": null,
+          "description": "Right margin."
+        },
+        {
+          "name": "mt",
+          "type": "Spacing",
+          "typeLabel": "Spacing",
+          "required": false,
+          "default": null,
+          "description": "Top margin."
+        },
+        {
+          "name": "name",
+          "type": "string",
+          "required": true,
+          "default": null,
+          "description": "Identifier for the group. Used in change events."
+        },
+        {
+          "name": "placeholder",
+          "type": "string",
+          "required": false,
+          "default": "",
+          "description": "Text shown when nothing is selected."
+        },
+        {
+          "name": "show-select-all",
+          "type": "boolean",
+          "required": false,
+          "default": "false",
+          "description": "Shows a \"Select All\" checkbox at the top of the options list. @default false"
+        },
+        {
+          "name": "size",
+          "type": "\"default\" | \"compact\"",
+          "values": [
+            "default",
+            "compact"
+          ],
+          "required": false,
+          "default": "default",
+          "description": "Sets the size variant."
+        },
+        {
+          "name": "testid",
+          "type": "string",
+          "required": false,
+          "default": "",
+          "description": "Sets a data-testid attribute for automated testing."
+        },
+        {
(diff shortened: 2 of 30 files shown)


bdfranck:
I've updated the styling and added references to the new component design tokens and fallbacks. The desk token PR is https://github.com/GovAlta/design-tokens/pull/161.

bdfranck (commented):

- ✅ I can open the dropdown by keyboard
- ✅ I can can navigate the items by arrow keys or `tab`
- ✅ I can see the label overflow properly
- ✅ I can see the label truncate properly
- ✅ I can see a disabled state
- ✅ I can see an error state
- ✅ I can filter options

Looking great! I added a commit with two small tweaks. There's only two minor things so far:

- When there are no results, can we add "No matches found" text with `secondary` color?
- Is there an easy way to prevent the `cursor: pointer` on the disabled state? It suggests that the item is still interactive.

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,606 @@
+<svelte:options
```
vanessatran-ddi, line 1: Testing feedback: 

- [ ] Home/End: (Optional) https://www.w3.org/WAI/ARIA/apg/patterns/listbox/#keyboardinteraction Can we make Home to move focus to the first option and End to move focus to the last option, W3 recommends for list >= 5 items and I think it makes sense. 
- [ ] When we focus on the multi select, filterable, we should be able to see the cursor appears right away and we can type the keywords. Right now, we need to tab to focus, press Enter before we can type the search string 
- [ ] Search logic: On our current Dropdown if type "Deer" or "deer" it will display "Red Deer" but on the Multi select component it displays nothing matched. 
- [ ] When I type "Calgary" -> arrow down -> press Enter -> I cannot use keyboard only to move up to the text field and continue to search (Shift + Tabs) doesn't work. We can refer to https://mantine.dev/core/multi-select/ this, Searchable example, where we search "Angular", arrow down, Enter to select, the text field clears the keyword "Angular", we can continue to type another one to continue without using a mouse. 
- [ ] I add the multi select inside the modal, I use a keyboard and press, select an option and then I press ESC to escape the multi select, however it closes the modal I am in. Expected to exit the multi select only, not the modal. 
- [ ] Filterable textfield should have a leading icon "search" similar to Dropdown.
- [ ] If I set Filerable true and selectAll true, and then I test: "Apple" and press Apple to choose it, then I click "Select All", the behaviour is very weird. We expect the search textfield cleared or update the label to reflect the choice "Select All", otherwise it behaves "conflicted"

https://github.com/user-attachments/assets/c42067fc-aed3-445a-b81e-a2ed1c9ee6b5

apps/prs/react/src/routes/docs/dropdown-multiselect/dropdown-multiselect.tsx
```diff
@@ -0,0 +1,119 @@
+import {
+  GoabDropdownItem,
+  GoabDropdownMultiselect,
+  GoabFormItem,
+  GoabText,
```
vanessatran-ddi: Unused import `GoabText`

docs/src/data/configurations/dropdown-multiselect.ts
```diff
@@ -0,0 +1,579 @@
...
+      <goab-dropdown-item value="banana" label="Banana"></goab-dropdown-item>
+      <goab-dropdown-item value="orange" label="Orange"></goab-dropdown-item>
+      <goab-dropdown-item value="pear" label="Pear"></goab-dropdown-item>
+    </goab-dropdown-multiselect>
+  </goab-form-item>
+</form>`,
+          },
+        ],
+        webComponents: `<goa-form-item version="2" label="Select fruits" mb="l">
+  <goa-dropdown-multiselect version="2" name="fruits" width="320px">
```
vanessatran-ddi: do we have `version` attribute for this new component?

docs/src/data/configurations/dropdown-multiselect.ts
```diff
@@ -0,0 +1,579 @@
...
+    width="320px"
+    truncateLabel={true}
+  >
+    <GoabDropdownItem value="apple" label="Apple" />
+    <GoabDropdownItem value="banana" label="Banana" />
+    <GoabDropdownItem value="orange" label="Orange" />
+    <GoabDropdownItem value="pear" label="Pear" />
+  </GoabDropdownMultiselect>
+</GoabFormItem>`,
+        angular: `<goa-form-item label="Select fruits" mb="l">
```
vanessatran-ddi: should be `goab-form-item` and `goab-dropdown-multiselect`

docs/src/data/configurations/dropdown-multiselect.ts
```diff
@@ -0,0 +1,579 @@
...
+    <goa-dropdown-item value="orange" label="Orange"></goa-dropdown-item>
+    <goa-dropdown-item value="pear" label="Pear"></goa-dropdown-item>
+  </goa-dropdown-multiselect>
+</goa-form-item>`,
+        webComponents: `<goa-form-item version="2" label="Select fruits" mb="l">
+  <goa-dropdown-multiselect
+    version="2"
+    name="fruits"
+    width="320px"
+    truncateLabel="true"
```
vanessatran-ddi: should be `truncate-label` right?

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,606 @@
...
+  customElement={{
+    tag: "goa-dropdown-multiselect",
+    props: {
+      testid: { type: "String", attribute: "testid", reflect: true },
```
vanessatran-ddi: we do not need to declare `tesid` here, unless we want to make it like `tesId`(camelCase) or `test-id` in the attribute. If we keep using `tesid` then we should remove this line. To make it similar to other components, I think we should keep `tesid` and remove this line

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,606 @@
...
+      truncateLabel: {
```
vanessatran-ddi: Can we change this `truncateLabel` to `labelDisplay` and instead of boolean, can we make it as `list | count`, the reason why I suggest so is if in the future we want to add more way to display this label (chips?) then this property will open more choices for us

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,606 @@
...
+        type: "String",
+        attribute: "truncate-label",
+        reflect: true,
+      },
+      selectAll: { type: "String", attribute: "select-all", reflect: true },
```
vanessatran-ddi: Can we rename this to be `allowSelectAll` or `withSelectAll`, `selectAll` starts with a verb so the type `boolean` makes me confused. Also can we follow `Drawer.svelte` with `open` property, so we can make `type` here as `Boolean` instead of `string`? Then we do not need `toBoolean` anymore

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,606 @@
...
+  } from "../dropdown/DropdownItem.svelte";
+
+  /** @required Identifier for the group. Used in change events. */
+  export let name: string;
+  /** Array of currently selected checkbox values. */
+  export let value: string[] = [];
+  /** Text shown when nothing is selected. */
+  export let placeholder: string = "";
+  /** Enables filtering of options by typing in the trigger. */
+  export let filterable: string = "false";
```
vanessatran-ddi: We can declare `filterable` on the top as a type `Boolean` like `open` inside `Drawer.svelte`

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,606 @@
...
+  /** Disables the component. */
+  export let disabled: string = "false";
+  /** Shows an error state. */
+  export let error: string = "false";
+  /** Sets a data-testid attribute for automated testing. */
+  export let testid: string = "";
+  /** Provides an accessible label when no visible label is associated. */
+  export let arialabel: string = "";
+  /** References an external element that labels this component. */
+  export let arialabelledby: string = "";
```
vanessatran-ddi: Like recently developed component `MenuButton`, I think we should declare `arialabelledby` as `ariaLabelledBy` on the options tags above and make attribute as `aria-labelledby`, also it would match the ARIA name

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,606 @@
...
+  /** Enables filtering of options by typing in the trigger. */
+  export let filterable: string = "false";
+  /** Disables the component. */
+  export let disabled: string = "false";
+  /** Shows an error state. */
+  export let error: string = "false";
+  /** Sets a data-testid attribute for automated testing. */
+  export let testid: string = "";
+  /** Provides an accessible label when no visible label is associated. */
+  export let arialabel: string = "";
```
vanessatran-ddi: rename as `ariaLabel` and define on options tag above, named attribute as `aria-label`

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,606 @@
...
+  /** References an external element that labels this component. */
+  export let arialabelledby: string = "";
+  /** Sets the maximum height of the dropdown content area. */
+  export let maxheight: string = "276px";
```
vanessatran-ddi: same here, make it `maxHeight` and attribute name `max-height`

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,606 @@
...
+      case "append":
+      case "reset":
+        _options = [..._options, detail];
+        break;
+    }
+    relay(detail.el, "dropdown:bind", { el: _rootEl });
+  }
+
+  function onChildDestroyed(detail: DropdownItemDestroyRelayDetail) {
+    _options = _options.filter((o) => o.value !== detail.value);
```
vanessatran-ddi, line 261: When this happen, can we reset the value, or at least keep track value to make sure it synced with available options, and also dispatch a _change event if it happens too

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,606 @@
...
+  }
+
+  function handleKeydown(e: KeyboardEvent) {
+    if (isDisabled) return;
+    switch (e.key) {
+      case "Enter":
+      case " ":
+        if (!_isOpen) {
+          e.preventDefault();
+          focusFirstCheckbox();
```
vanessatran-ddi: We dont need to call `focusFirstCheckbox` here, reason is: `_isOpen` we pass to `goa-popover` below, will be handled by `Popover.svelte` with function `syncPopoverOpenState`, and onMount above we already listen to `_open` from `goa-popover` and trigger this `focusFirsCheckbox` again, which means the current way will go through `focusFirstCheckbox` 2 times, with requestAnimation delay there

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,606 @@
...
+  function getCheckboxList(): HTMLElement[] {
+    const listItems = Array.from(
+      _checkboxListEl?.querySelectorAll("goa-checkbox") || [],
+    ) as HTMLElement[];
+    return isSelectAll && _selectAllEl
+      ? [_selectAllEl, ...listItems]
+      : listItems;
+  }
+
+  function focusFirstCheckbox() {
```
vanessatran-ddi, line 364: focusFirstCheckbox`, `focusNextCheckbox` and `focusPreviousCheckbox` have duplicated logic, we should merge 3 of them into `focusCheckboxAt(index)` for example, also, inside the function, we can make use of `findFirstFocusableNode([checkboxes[_focusedIndex]])?.focus()` from `utils` instead of querySelector with shadowRoot

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,606 @@
...
+  /** Provides an accessible label when no visible label is associated. */
+  export let arialabel: string = "";
+  /** References an external element that labels this component. */
+  export let arialabelledby: string = "";
+  /** Sets the maximum height of the dropdown content area. */
+  export let maxheight: string = "276px";
+  /** Sets a fixed width for the component and popover panel. */
+  export let width: string = "";
+  /** Sets the size variant. */
+  export let size: "default" | "compact" = "default";
```
vanessatran-ddi, line 120: for `size`, we should add `typeValidator` like other components and trigger at `onMount`

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,606 @@
...
+          type={_isOpen ? "chevron-up" : "chevron-down"}
+          size="medium"
+          aria-hidden="true"
+        />
+      </div>
+    </div>
+
+    <div
+      id={_contentId}
+      role="none"
```
vanessatran-ddi: perhaps we should make `role=\"dialog\"` at least, instead of `none` and we may add `aria-label=placeholder | \"Options\"` for example, otherwise the `aria-controls` at line 363 will point to this and screen reader will not be able to announce the relationship between the trigger above and this div

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,606 @@
...
+    overflow-y: auto;
+    padding: var(
+      --goa-dropdown-multiselect-padding,
+      var(--goa-space-m) var(--goa-space-s)
+    );
+  }
+
+  .select-all-divider {
+    border: none;
+    border-top: var(--goa-border-width-s) solid var(--goa-color-greyscale-200);
```
vanessatran-ddi: Is there a reason why this `border-top` does not have a token like the rest?

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,606 @@
...
+          aria-hidden="true"
+        />
+      </div>
+    </div>
+
+    <div
+      id={_contentId}
+      role="none"
+      class="content"
+      style="max-height: {maxheight}; overflow-y: auto;"
```
vanessatran-ddi: do we need inline style `overflow-y` here, I see we define under css `.content` line 594 too

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,606 @@
...
+    DropdownItemDestroyRelayDetail,
+    Option,
+  } from "../dropdown/DropdownItem.svelte";
+  import {
+    DropdownItemMountedMsg,
+    DropdownItemDestroyMsg,
+  } from "../dropdown/DropdownItem.svelte";
+
+  /** @required Identifier for the group. Used in change events. */
+  export let name: string;
```
vanessatran-ddi, line 96: consider to use `validateRequired` for this `name` property, under `onMount`

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,606 @@
...
+    const input = (checkboxes[_focusedIndex] as any).shadowRoot?.querySelector(
+      "input[type='checkbox']",
+    ) as HTMLElement | null;
+    if (input) input.focus();
+    else checkboxes[_focusedIndex].focus();
+  }
+
+  function handleFilterKeydown(e: KeyboardEvent) {
+    if (e.key === "ArrowDown") {
+      e.preventDefault();
```
vanessatran-ddi, line 401: Missing `e.stopPropagation()`, consequence is when we search `New` and we have 2 choices `New Brunswick` and `Newfoundland..`, we press ArrowDown, it instead of jumping to the first choice New Brunswick, it jumps to the second choice

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,606 @@
...
+  let _filterInputEl: HTMLInputElement | undefined;
+  let _selectAllEl: HTMLElement | undefined;
+  let _focusedIndex = -1;
+
+  // Reactive
+  $: isDisabled = toBoolean(disabled);
+  $: isError = toBoolean(error);
+  $: isFilterable = toBoolean(filterable);
+  $: isSelectAll = toBoolean(selectAll);
+  $: if (!Array.isArray(value)) value = [];
```
vanessatran-ddi: consider to do `Array.isArray(value) ? value : []`, reason is Angular/React also pass down the `value`, if we override it `value` internal with `[]`, it can become a loop of reactivity from React/Angular to our internal value hook here

docs/generated/component-apis/dropdown-multiselect.json
```diff
@@ -0,0 +1,444 @@
...
+          "type": "boolean",
+          "required": false,
+          "default": null,
+          "description": "Enables filtering of options by typing in the trigger."
+        },
+        {
+          "name": "maxHeight",
+          "type": "string",
+          "required": false,
+          "default": "276px",
```
vanessatran-ddi, line 60: I run `npm run build:prod` and this value becomes `null`

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
willcodeforcoffee: This is a great idea. I'm going to go with `labelFormat` instead though, since it affects the formatting of the label. I hope that is okay?

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
willcodeforcoffee: This is also a really good idea, but I went with `showSelectAll` since it is more consistent with other components like `showBanner`, `showMenu` (in AppHeader), `show` in Notification, etc.

docs/src/data/configurations/dropdown-multiselect.ts
willcodeforcoffee: No, it is only version 2. Removed!

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
willcodeforcoffee: This is updated and uses the `--goa-dropdown-multiselect-divider` from https://github.com/GovAlta/design-tokens/pull/161

willcodeforcoffee:
Updates completed, and all feedback has been addressed!

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,758 @@
+<svelte:options
```
vanessatran-ddi, line 1: Not on this specific line, but I recognize this weird issue: 
When you search "Cherry", and then you navigate up and down, select like this video, the last state is not correct: I see "Select All" checked but Cherry is not checked. and not all options are checked. 

Image

Image

https://github.com/user-attachments/assets/9d980b78-bfbf-47f8-b0ac-cd6b4ff19c72

docs/src/data/configurations/dropdown-multiselect.ts
```diff
@@ -0,0 +1,576 @@
...
+
+  configurations: [
+    {
+      id: "basic",
+      name: "Basic example",
+      description: "Simple dropdown multiselect with options",
+      code: {
+        react: `const [fruits, setFruits] = useState<string[]>([]);
+
+<GoabFormItem label="Select fruits" mb="l">
```
vanessatran-ddi: I notice that all examples for docs, we add `mb=l` for the `goab-form-item`, I dont think we need this extra `mb` for the code example, especially this is for developers to copy and paste. Can we remove `mb` on our code snippets?

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,758 @@
...
+    DropdownItemDestroyRelayDetail,
+    Option,
+  } from "../dropdown/DropdownItem.svelte";
+  import {
+    DropdownItemMountedMsg,
+    DropdownItemDestroyMsg,
+  } from "../dropdown/DropdownItem.svelte";
+  import { GoAIconType } from "../icon/Icon.svelte";
+  import { __classPrivateFieldGet } from "tslib";
+  import { GoabDropdownMultiselectLabelFormatOptions } from "@abgov/ui-components-common";
```
vanessatran-ddi: We don't import the type directly from `ui-components-common`, here we can do like the way we define `size` at line 70 and also we need to add typeValidator for labelFormat too, similar like what we do with size.

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,758 @@
...
+    DropdownItemMountedRelayDetail,
+    DropdownItemDestroyRelayDetail,
+    Option,
+  } from "../dropdown/DropdownItem.svelte";
+  import {
+    DropdownItemMountedMsg,
+    DropdownItemDestroyMsg,
+  } from "../dropdown/DropdownItem.svelte";
+  import { GoAIconType } from "../icon/Icon.svelte";
+  import { __classPrivateFieldGet } from "tslib";
```
vanessatran-ddi: Line 67 seems to be accidentally import

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,758 @@
...
+  customElement={{
+    tag: "goa-dropdown-multiselect",
+    props: {
+      labelFormat: {
+        type: "String",
+        attribute: "label-format",
+        reflect: true,
+      },
+      showSelectAll: {
+        type: "String",
```
vanessatran-ddi: Below we define `showSelectAll` as boolean: false, here we are safe to update type `String` to `Boolean`, as in React and Angular wrapper we already make this as `"true"` or `undefined` (never `"false"` passed in)

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,758 @@
...
+  const [Size, validateSize] = typeValidator("Size", ["default", "compact"]);
+
+  /** @required Identifier for the group. Used in change events. */
+  export let name: string;
+  /** Array of currently selected checkbox values. */
+  export let value: string[] = [];
+  /** Text shown when nothing is selected. */
+  export let placeholder: string = "";
+  /** Enables filtering of options by typing in the trigger. */
+  export let filterable: boolean = false;
```
vanessatran-ddi, line 102: We should declare filterable type `Boolean` on the options above, if we want to declare this as `boolean` and don't need to use `toBoolean` function

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,758 @@
...
+        if (_isOpen) {
+          _isOpen = false;
+        }
+        break;
+
+      case "ArrowDown":
+        if (!_isOpen) {
+          e.preventDefault();
+          focusFirstCheckbox();
+          _isOpen = true;
```
vanessatran-ddi, line 467: Missing from the previous code review so I re-comment here. so here at line 395 when we set `_isOpen=true` we will open the popover -> _open is true -> handlePopoverOpen is called -> focusFirstCheckbox again (2nd time). The fix here is to remove `focusFirstCheckbox()` at line 394 as it is redundant.

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
vanessatran-ddi: This is not resolved

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,758 @@
...
+  const _contentId = `goa-dropdown-multiselect-content-${generateRandomId()}`;
+  let _filterText = "";
+  let _filterInputEl: HTMLInputElement | undefined;
+  let _selectAllEl: HTMLElement | undefined;
+  let _focusedIndex = -1;
+
+  // Reactive
+  $: isDisabled = toBoolean(disabled);
+  $: isError = toBoolean(error);
+  $: Array.isArray(value) ? value : [];
```
vanessatran-ddi: It should be `$:value = Array.isArray(value) ? value : []` right? It is missing assining to value now

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,758 @@
...
+  $: _hasValue = value.length > 0;
+  $: _allSelected = _options.length > 0 && value.length === _options.length;
+  $: _someSelected = value.length > 0 && value.length < _options.length;
+  $: _labelMap = Object.fromEntries(
+    _options.map((o) => [o.value, o.label || o.value]),
+  );
+  $: _displayText = calculateDisplayText(value, _labelMap);
+  $: _visibleOptions =
```
vanessatran-ddi, line 168: Not this line, but we should dispatch `error::change` and `help-text::announce` otherwise FormItem.svelte cannot announce when there is an error or helptext (for screen readers)

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,758 @@
...
+        { name, value, labels: value.map((v) => _labelMap[v] || v) },
+        { bubbles: true },
+      );
+    }
+  }
+
+  function handlePopoverOpen() {
+    _isOpen = true;
+    if (filterable) {
+      requestAnimationFrame(focusFilterTextbox);
```
vanessatran-ddi: When I test, I don't see the focus focusing on the text box even with filterable true, it is always on the first checkbox.

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,758 @@
...
+
+  function focusCheckboxAt(index: number) {
+    const checkboxes = getCheckboxList();
+    if (checkboxes.length === 0) return;
+    const checkbox = checkboxes[index];
+    const input = findFirstFocusableNode([checkbox]) as HTMLElement | null;
+
+    input?.focus();
+
+    _focusedIndex = index;
```
vanessatran-ddi, line 361: First, we don't need extra line at line 253 and 255. Second: we should return earlier when `checkbox is null` at line 251, and when `input is null` at line 252, otherwise `_focusedIndex` will be off track

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,758 @@
...
+  }
+
+  function isFilterMatch(option: Option, filter: string, partialMatch = true) {
```
vanessatran-ddi: This function is exactly same with function isFilterMarch on Dropdown.svelte, can we make it under common/ maybe utils or if you want it more specific, we can put under common/utils.ts

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,758 @@
...
+    if (checkboxes.length === 0) return;
+
+    focusCheckboxAt(0);
+  }
+
+  function focusLastCheckbox() {
+    const checkboxes = getCheckboxList();
+    if (checkboxes.length === 0) return;
+
+    focusCheckboxAt(checkboxes.length - 1);
```
vanessatran-ddi, line 372: Inside `focusCheckboxAt` we already cover line 282 and 283, which means we repeat oursevles here. If we remove line 282 and 283, we don't need this function anymore, we can replace with focusCheckboxAt(getCheckboxList().lenth -1 ) for focusLastCheckbox, and focusNextCheckbox can be focusCheckboxAt(_focusIndex + 1)

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,758 @@
...
+
+  function handleSelectAllChange(e: Event) {
+    e.stopPropagation();
+    const detail = (e as CustomEvent<{ checked: boolean }>).detail;
+    if (!detail) return;
+    if (detail.checked) {
+      // Use visible options in the case of a filter
+      const allValues = _visibleOptions.map((o) => o.value);
+      const allLabels = _visibleOptions.map((o) => o.label || o.value);
+      value = allValues;
```
vanessatran-ddi: Seems not correct to me, I think we should make it union, not replace the whole value. So something like this: 
```
const visibleValues = _visibleOptions.map((o) => o.value);
value = detail.checked ? Array.from(new Set([...value, ...visibleValues])) : value.filter((v) => !visibleValues.includes(v)); 
dispatch (...) 
```
We don't need if else too.

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,758 @@
...
+      { name, value: [], labels: [] },
+      { bubbles: true },
+    );
+    if (filterable) {
+      requestAnimationFrame(focusFilterTextbox);
+    }
+    _isOpen = true;
+  }
+
+  function handleClearIconKeyDown(e: KeyboardEvent) {
```
vanessatran-ddi, line 501: looks like `e` is not used here.

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
willcodeforcoffee: You're right, I was able to shrink this quite a bit. I cut out labels as well.

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
willcodeforcoffee: This should be fixed now. Clearing the filter input actually cleared all the selections!

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
willcodeforcoffee: Yeah! I wonder what happened to this?

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
willcodeforcoffee: This works and is a lot more reliable, now. I ended up getting pretty frustrated trying to find a cleaner way and went with Claude's solution. The popover and focus-trap made focusing on the input/trigger after the popover challenging. 😞

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
willcodeforcoffee: Done! Good idea.

willcodeforcoffee:
Updates completed:
- [x] Fixed the strange filter + Select All behavior
- [x] focusing on the filter textbox when fitlerable
- [x] Removed all the `mb` attributes from the docs
- [x] removed unused imports and variables
- [x] fixed types like made `showSelectAll` and `filterable` are `boolean` instead of `string`
- [x] extracted and refactored the `isFilterMatch` method from Dropdown into common
- [x] extracted and refactored the `focusCheckbox` methods to remove duplicate code
- [x] restructured the `filteredValues`
- [x] implemented `fieldset` events like `help-text::announce` and `error::change`
- [x] added more examples in the PRs

libs/web-components/src/components/dropdown-multiselect/DropdownMultiselect.svelte
```diff
@@ -0,0 +1,823 @@
...
+    box-sizing: border-box;
+    font-family: var(--goa-font-family-sans);
+    display: block;
+  }
+
+  .root {
+    width: 100%;
+  }
+
+  .trigger {
```
bdfranck, line 670: I noticed that iOS Safari defaults to a system colour for the leading icon and the caret. 

Image

We can fix this by adding the text colour to `.trigger`:

```
    color: var(
      --goa-dropdown-multiselect-color-text,
      var(--goa-input-color-text-default)
    );
```

bdfranck (commented):


Item | Chrome | Safari | Firefox | iOS Safari | Android Chrome
-- | -- | -- | -- | -- | --
Open & close | ✅ | ✅ | ✅ | ✅ | ✅
Select an item | ✅ | ✅ | ✅ | ✅ | ✅
Deselect an item | ✅ | ✅ | ✅ | ✅ | ✅
Filter | ✅ | ✅ | ✅ | ✅ | ✅
Select all | ✅ | ✅ | ✅ | ✅ | ✅
Deselect all | ✅ | ✅ | ✅ | ✅ | ✅
Correct width | ✅ | ✅ | ✅ | ✅ | ✅
Works in modal | ✅ | ✅ | ✅ | ✅ | ✅

Looking good! I just have one comment about the icon colours on iOS.

bdfranck (approved):

- ✅ The leading icon and trailing caret are the correct colour in iOS Safari

Image

Looks good to me! 👍

willcodeforcoffee:
Fixed iOS Safari colours:
image

Dark mode:
image

Tested on:
- [x] BrowserStack iPhone 16
- [x] IRL iPad Mini 5

That was a good catch, @bdfranck ! I totally missed it.

I also updated the disabled state to match the text color:

image

image

Thank you!

vanessatran-ddi (approved):

PR #3953, opened 2026-05-19
fix(#3323): use anchor/fixed positioning for Tooltip

### LINKED ISSUE #3323: Tooltip inside Callout gets cut off

bdfranck:
This PR fixes the tooltip getting cropped by parent components with the following changes:
- Switch to CSS anchor positioning so the tooltip is always on top
- Use fixed positioning as a fallback for older browsers that don't support CSS anchor positioning
- Improves the alignment calculations so the tooltip doesn't overflow the viewport

## Using CSS Anchor Positioning

tooltip-in-callout

tooltips-anchor

## Using fixed position fallback

tooltips-fixed-fallback

## Note

- The original version of the PR used the [Popover API](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API). We removed it because it caused issues for desktop and mobile Safari

### DIFF
--- a/libs/react-components/specs/tooltip.browser.spec.tsx
+++ b/libs/react-components/specs/tooltip.browser.spec.tsx
@@ -1,4 +1,5 @@
 import { render } from "vitest-browser-react";
+import { page } from "@vitest/browser/context";
 
 import { GoabTooltip, GoabButton } from "../src";
 import { expect, describe, it, vi } from "vitest";
@@ -324,7 +325,9 @@ describe("Tooltip Browser Tests", () => {
     );
   });
 
-  it("should apply maxWidth using CSS custom property", async () => {
+  it("should apply maxWidth for long tooltip content", async () => {
+    await page.viewport(1280, 800);
+
     const Component = () => {
       return (
         <div data-testid="container">
@@ -338,8 +341,10 @@ describe("Tooltip Browser Tests", () => {
               </div>
             }
             maxWidth="300px"
+            hAlign="right"
+            testId="test-tooltip"
           >
-            <GoabButton>Constrained tooltip</GoabButton>
+            <GoabButton>Tooltip</GoabButton>
           </GoabTooltip>
         </div>
       );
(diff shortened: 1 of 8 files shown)


apps/prs/angular/src/routes/bugs/fix3323/fix3323.component.html
```diff
@@ -0,0 +1,54 @@
+<div style="display: grid; gap: 1rem;">
```
ArakTaiRoth: This is for all of the PR files here. Can you rename them to be the proper `bugXXXX`? Instead of `fixXXXX`? Can you also remove all extra CSS styling that's being used in all of them, having it makes it more complicated to test the bug in isolation. Also, maybe use GoabText everywhere?

libs/web-components/src/components/tooltip/Tooltip.svelte
```diff
@@ -113,24 +111,42 @@
     maxwidth = validateMaxWidth(maxwidth);
 
     _initialPosition = position;
+    _computedAlign = halign;
     _tooltipInstanceId = Math.random().toString(36);
 
-    window.addEventListener("resize", checkAndAdjustPosition);
-    checkAndAdjustPosition();
+    window.addEventListener("resize", onWindowResize);
+    reconcileTooltipLayout();
```
ArakTaiRoth: This is being called before the popover is being opened, which then calls `applyTooltipWidth`, which then writes the measured width back inline. The problem is, the popover isn't rendered while closed, so the initial `getBoundingClientRect` can be 0. Then later when `showPopover` is called, it measures that width. This all means that long tooltips will have problems. I've updated the React PR test file with an example at the bottom that shows off the issue.

libs/react-components/specs/tooltip.browser.spec.tsx
```diff
@@ -353,22 +447,19 @@ describe("Tooltip Browser Tests", () => {
         const tooltipEl = container
           .element()
           .querySelector("goa-tooltip") as HTMLElement | null;
-        expect(tooltipEl).toBeTruthy();
+        expect(tooltipEl?.tagName.toLowerCase()).toBe("goa-tooltip");
 
         tooltipEl?.dispatchEvent(new Event("mouseenter", { bubbles: true }));
```
ArakTaiRoth: This is dispatching `mouseenter` on `goa-tooltip`, which the listener isn't actually attached to. The listener is attached to the shadowRoot `.tooltip` element. Because of that, these tests don't actually test the proper width, as evidenced by this test passing while I wrote a manual test with roughly this same situation that fails. You could test the actual shadowRoot element (tested already in the test on line 169), wait for the class `show` to be attached, then test the width. Also you'd have to have a bounds on the width, instead of just testing equal or less than.

libs/web-components/src/components/tooltip/Tooltip.svelte
```diff
@@ -153,84 +200,201 @@
...
+      }
+    }
+  }
+
+  function updateTargetWidth() {
+    if (!_rootEl || !_tooltipEl || !_targetEl) return;
+
+    const targetHalfWidth = `${_targetEl.getBoundingClientRect().width / 2}px`;
+    _rootEl.style.setProperty("--target-width", targetHalfWidth);
+    _tooltipEl.style.setProperty("--target-width", targetHalfWidth);
```
willcodeforcoffee: Do you have to set this variable on both elements?

bdfranck: Good question! Since the value is inhereted by the element's children, I only need to set it on root. I've removed the second one.

libs/web-components/src/components/tooltip/Tooltip.svelte
bdfranck: Good catch! I've removed the call to `reconcileTooltipLayout` on mount. Then I added a guard to `reconcileTooltipLayout` to skip it the tooltip isn't open.

bdfranck:
I've addressed the above comments and updated the browser tests so they are faster and (hopefully) more reliable. 🤞

ArakTaiRoth (approved):

willcodeforcoffee (changes requested):
Testing worked really well on my computer, and on Safari 18 using Browserstack.
It also worked really well on Android devices when tapping the triggers.
Unfortunately on iOS devices like iPad and iPhone most of them didn't work. the "Long tooltip" mostly worked, and the button triggered tooltips sometimes showed on the second tap.
It seems the likely cause might be that Safari iOS cannot/will not focus on certain elements.

bdfranck:
~After chatting with @willcodeforcoffee, I found a fix for mobile Safari... but it broke support for Desktop Safari. Since tooltips are more important on desktop, I removed the commits for the mobile fix.~

~I'll make an issue for addressing the Mobile Safari tooltip issues later.~

We since learned that the [Popover API](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API) also didn't work well with Safari Desktop for tooltips. I've removed the Popover API portion but kept the rest of the changes.

willcodeforcoffee (approved):
Tested in BrowserStack on:
- iPhone 16 Pro Max
- iPad Pro 13 2025
- Sequoia / Safari 18.4
- Tahoe / Safari 26.4
- Samsung S23 Ultra / Samsung Browser

Test IRL on iPad Mini.
Chrome on Windows 11
Firefox Developer Edition on Windows 11

It is working successfully on _every_ browser!

PR #3955, opened 2026-05-19
(spike) Editable HTML in the config sandbox

### LINKED ISSUE #3958: Docs site: editable code snippets
## Description

After a retro discussion about making documentation sandbox configurable again, a quick exploration showed a lightweight approach is viable: making the code snippet under Web Components editable directly in the browser.

The implementation allows users to change content, change properties, and add other components inline — making it a flexible live editing experience.

## Acceptance Criteria

- [ ] Editable code snippets are available on component pages
- [ ] Users can change content and properties inline
- [ ] Changes are reflected live in the rendered output

twjeffery:
## Summary

You can now type into the code snippet for the web component of any config sandbox and the preview updates as you go. The idea is to let people try their own content or markup without leaving the page.

This is a quick spike for the team to give feedback on. 

## Why
A user cannot currently change the content of the config examples at the top of each component page. In our old design system website, we had a more fully featured sandbox where a team could edit properties and change content. We changed our approach to use a range of configured templates to show the component properties and variants so that it would be easier for us to maintain. Previously, there were also a lot of issues where certain properties and variants would not show up properly in the playground sandbox. 

This now gives the user the ability to edit the code and see the preview update. You can edit and add content, but you can also add more components, other content, change variants, and more within that sandbox preview now.  

## How to try it

1. Open any component page, e.g. `/components/badge` or `/components/accordion`
2. Scroll to the code area below the preview, click the Web Components tab
3. Click into the HTML. You should see the text cursor.
4. Type. After a brief pause the preview above updates.
5. Switch variants in the dropdown. Your edits clear, original HTML comes back.
6. Hard refresh resets everything.

## Only the Web Components tab is editable

The preview is the web component HTML you see in that tab, put straight onto the page. So editing the HTML changes what's shown. The React and Angular tabs are there to show how to use the same thing in those frameworks. Making those editable wouldn't reach the preview. 

## What's not in here

- The CSS and JS blocks of the Web Components tab (still read-only)
- React and Angular tabs (still read-only)
- Mirroring edits between framework tabs
- Edits sticking around after a page refresh
- Syntax highlighting while typing

## Open questions

- Is the cursor changing enough of a signal that you can edit, or does it need a visible hint?
- Should this also work on single-example pages, not just the configurable previews?
- Is being limited to the Web Components tab fine, or worth exploring how to mirror edits into the React/Angular code shown alongside?

## Test plan

- [ ] Component pages still load and previews still show
- [ ] Editing the HTML updates the preview after a brief pause
- [ ] Switching variants clears edits
- [ ] React and Angular tabs stay read-only
- [ ] Full page reload resets edits
- [ ] Focus outline matches the rest of the design system

image

### DIFF
--- a/docs/src/components/CodeSnippet.css
+++ b/docs/src/components/CodeSnippet.css
@@ -155,6 +155,32 @@ code {
   font-style: italic;
 }
 
+/* Brief 121: editable HTML block. Visually matches a code block but accepts
+   direct edits; the only affordance is the text-caret cursor on hover. */
+.editable-html {
+  display: block;
+  width: 100%;
+  box-sizing: border-box;
+  margin: 0;
+  padding: var(--goa-space-m, 1rem);
+  border: none;
+  background: var(--goa-color-greyscale-100, #f1f1f1);
+  color: var(--goa-color-text-default, #333);
+  font-family: "SF Mono", Monaco, "Courier New", monospace;
+  font-size: var(--goa-font-size-2);
+  line-height: 1.6;
+  resize: vertical;
+  outline: none;
+  white-space: pre;
+  overflow: auto;
+}
+
+/* Lift the focus outline up to the wrapper so its corners round with the
+   wrapper's border-radius instead of getting clipped by overflow:hidden. */
+.code-snippet:has(.editable-html:focus) {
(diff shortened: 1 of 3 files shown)


twjeffery:
Closing as this was a spike

PR #3956, opened 2026-05-19
chore(#3770): generate MCP data from docs site content

### LINKED ISSUE #3770: Generate the MCP from docs site content
Once design system guidelines live in the docs site content collections (#3769), we can generate the MCP and any other AI-facing docs from the same source. The docs site becomes the source of truth, and the generated outputs stay fresh whenever the content changes.

## What this looks like

- A build step that reads docs site content (component metadata, API data, guidance atoms, examples, foundations) and produces the MCP data
- Runs as part of the docs build so we don't have to remember to regenerate
- The same pipeline can output other AI-facing docs as we figure out what's useful

## Why now

Once we've moved the design system guidelines into the docs site (#3769), the docs site has everything the MCP needs and can be generated. This will mean we maintain it in one place and the MCP stays updated automatically as the component library grows and changes.

twjeffery:
## Summary

Generator pipeline at `docs/src/scripts/content-generators/` reads from the docs content collections (components, examples, guidance, foundations, get-started, productTypes) and emits JSON the design-system-mcp consumes. Output lands in `docs/generated/mcp/`, gitignored.

The docs site becomes the single source of truth for design system knowledge. The MCP no longer carries a separate set of hand-maintained data files.

### Pipeline shape

Four stages: `loaders/` read content, `transforms/` normalize and cross-link records, `checks/` cross-validate references, `outputs/` serialize to disk. Deterministic key order so reruns produce byte-identical output.

### Where each field comes from

- **Technical facts** (web component tag, React class name, Angular selector, props/events/slots): from the code in `libs/*` and `generated/component-apis/`. Never from frontmatter, so they can't drift.
- **Curation** (description, status, category, tags, related items): from frontmatter in docs collections.
- **Aliases**: derived automatically. Stale wrapper names like `GoabAppFooter` and `GoabTemporaryNotificationCtrl` land on the parent component's aliases array so old lookups still resolve. A small `COMPONENT_SLUG_ALIASES` map in `transforms/aliases.ts` captures legacy or sub-component names (text-area, notification, menu-button, table sub-components) that don't come from code.

### Cross-validation

The `checks/` stage walks all loaded records and verifies cross-references before output is written:

- 6 error-level checks for id-shaped references: `components.relatedComponents`, `components.relatedGuidance`, `examples.components`, `examples.relatedExamples`, `guidance.appliesTo.components`, `productTypes.components`. References resolve through aliases when available.
- 1 warn-level check for `guidance.relatedProps` against extracted prop names on the linked components (valid if the prop exists on any of them, since a single `relatedProps` entry usually describes a prop that lives on one of several related components).
- Broken references include a "did you mean..." hint when within 2 edits of a known id.
- Errors block output and preserve the previous `docs/generated/mcp/` on disk; warnings ship through.

Current state on dev: 0 errors, 0 warnings.

### Sync to dcp-monorepo

A sync script copies the output into the design-system-mcp's `data/` folder. Manual for now, automated later. The paired PR over in dcp-monorepo brings the corresponding data-loader changes.

The generator is also invoked manually for now (`npx tsx docs/src/scripts/content-generators/index.ts`); it's not part of `docs:build` yet. Both the generator run and the sync get wired up in a follow-up.

Paired with: GovAlta/dcp-monorepo#342

Fixes #3770

## Test plan

- [ ] `npx tsx docs/src/scripts/content-generators/index.ts` from the repo root produces output under `docs/generated/mcp/` with 6 folders.
- [ ] `footer` component's JSON includes `GoabAppFooter` and `app-footer` in its aliases.
- [ ] `question-page` example JSON includes the 9 old-slug aliases.
- [ ] `temporary-notification` component JSON shows `GoabTemporaryNotificationCtrl` as an alias.
- [ ] `docs/generated/mcp/` is gitignored.
- [ ] Introducing a broken `relatedComponents` ref in any component frontmatter blocks output (exit 1) and prints the error with a suggestion. Existing files in `docs/generated/mcp/` remain unchanged.
- [ ] Introducing a broken `relatedProps` ref in any guidance atom prints a warning but allows output to ship.

### DIFF
--- a/docs/generated/component-apis/temporary-notification.json
+++ b/docs/generated/component-apis/temporary-notification.json
@@ -1,6 +1,6 @@
 {
   "componentSlug": "temporary-notification",
-  "extractedFrom": "libs/web-components/src/components/temporary-notification/TemporaryNotification.svelte",
+  "extractedFrom": "libs/web-components/src/components/temporary-notification/TemporaryNotificationCtrl.svelte",
   "frameworks": {
     "react": {
       "props": [
@@ -77,6 +77,18 @@
           "description": "Direction the notification animates from when appearing or disappearing."
         },
         {
+          "name": "horizontal-position",
+          "type": "\"left\" | \"center\" | \"right\"",
+          "values": [
+            "left",
+            "center",
+            "right"
+          ],
+          "required": false,
+          "default": "center",
+          "description": "Horizontal position of the notification container."
+        },
+        {
           "name": "message",
           "type": "string",
           "required": false,
@@ -93,8 +105,8 @@
         {
           "name": "testid",
           "type": "string",
-          "required": false,
-          "default": "",
+          "required": true,
+          "default": null,
           "description": "Sets a data-testid attribute for automated testing."
         },
         {
@@ -110,6 +122,17 @@
           "required": false,
           "default": "basic",
           "description": "The notification type which determines the visual style and icon."
+        },
+        {
+          "name": "vertical-position",
+          "type": "\"top\" | \"bottom\"",
+          "values": [
+            "top",
+            "bottom"
+          ],
+          "required": false,
+          "default": "bottom",
+          "description": "Vertical position of the notification container."
         },
         {
           "name": "visible",
--- a/docs/generated/component-apis/temporary-notification.json
+++ b/docs/generated/component-apis/temporary-notification.json
@@ -59,24 +59,6 @@
     "webComponents": {
       "props": [
         {
-          "name": "action-text",
-          "type": "string",
-          "required": false,
-          "default": "",
-          "description": "Text for the optional action button. When provided, displays a clickable link button."
-        },
-        {
-          "name": "animation-direction",
-          "type": "\"up\" | \"down\"",
-          "values": [
-            "up",
-            "down"
-          ],
-          "required": false,
-          "default": "down",
-          "description": "Direction the notification animates from when appearing or disappearing."
-        },
-        {
           "name": "horizontal-position",
           "type": "\"left\" | \"center\" | \"right\"",
           "values": [
@@ -89,39 +71,11 @@
           "description": "Horizontal position of the notification container."
         },
         {
-          "name": "message",
-          "type": "string",
-          "required": false,
-          "default": "",
-          "description": "The notification message text to display."
-        },
-        {
-          "name": "progress",
-          "type": "number",
-          "required": false,
-          "default": "-1",
-          "description": "Progress value from 0-100. Use -1 to hide the progress bar. Only applies when type is \"progress\"."
-        },
-        {
           "name": "testid",
           "type": "string",
           "required": true,
           "default": null,
           "description": "Sets a data-testid attribute for automated testing."
-        },
-        {
-          "name": "type",
-          "type": "\"basic\" | \"success\" | \"failure\" | \"indeterminate\" | \"progress\"",
-          "values": [
-            "basic",
-            "success",
-            "failure",
-            "indeterminate",
-            "progress"
-          ],
-          "required": false,
-          "default": "basic",
-          "description": "The notification type which determines the visual style and icon."
         },
         {
           "name": "vertical-position",
@@ -133,13 +87,6 @@
           "required": false,
           "default": "bottom",
           "description": "Vertical position of the notification container."
-        },
-        {
-          "name": "visible",
-          "type": "boolean",
-          "required": false,
-          "default": "true",
-          "description": "Controls whether the notification is visible."
         }
       ],
       "events": [],
--- a/docs/src/scripts/content-generators/checks/prop-refs.ts
+++ b/docs/src/scripts/content-generators/checks/prop-refs.ts
@@ -1,0 +1,95 @@
+import type { AnyRecord, ComponentRecord, Finding } from "../types";
+import { computeHint } from "./lib";
+
+/**
+ * Validates guidance.relatedProps against the extracted prop names on the
+ * components named in guidance.appliesTo.components.
+ *
+ * Warn-level. A relatedProps entry is valid if it exists on AT LEAST ONE
+ * of the linked components, matching the semantic "this guidance is about
+ * this prop, which lives on one of these components."
+ *
+ * Skips guidance atoms whose appliesTo.components references components
+ * that do not exist; component-refs.ts reports those separately.
+ */
+export function checkPropRefs(records: AnyRecord[]): Finding[] {
+  const findings: Finding[] = [];
+  const componentMap = new Map<string, ComponentRecord>();
+  for (const r of records) {
+    if (r.collection === "components") componentMap.set(r.id, r);
+  }
+
+  for (const r of records) {
+    if (r.collection !== "guidance") continue;
+    if (!Array.isArray(r.relatedProps) || r.relatedProps.length === 0) continue;
+    const linkedComponents = r.appliesTo?.components;
+    if (!Array.isArray(linkedComponents) || linkedComponents.length === 0) {
+      continue;
+    }
+
+    // Build the union of props across linked components that exist.
+    // Missing components are reported by component-refs.ts; skip them here
+    // to avoid double-reports.
+    const allValidProps = new Set<string>();
+    const knownComponents: string[] = [];
+    for (const componentId of linkedComponents) {
+      const component = componentMap.get(componentId);
+      if (!component) continue;
+      knownComponents.push(componentId);
+      for (const propName of extractPropNames(component)) {
+        allValidProps.add(propName);
+      }
+    }
+    if (knownComponents.length === 0) continue;
+
+    for (const propRef of r.relatedProps) {
+      if (allValidProps.has(propRef)) continue;
+      findings.push({
+        severity: "warning",
+        source: { collection: "guidance", id: r.id },
+        field: "relatedProps",
+        brokenRef: propRef,
+        context:
+          knownComponents.length === 1
+            ? `on component "${knownComponents[0]}"`
+            : `on components ${knownComponents.map((c) => `"${c}"`).join(", ")}`,
+        notFoundMessage: "not in extracted props",
+        hint: computeHint(propRef, allValidProps),
+      });
+    }
+  }
+
+  return findings;
+}
+
+/**
+ * Walks the component's api blob (frameworks.{react,angular,webComponents}.props[])
+ * and returns the union of all prop names declared across frameworks.
+ */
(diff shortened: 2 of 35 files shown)


docs/src/scripts/content-generators/loaders/examples.ts
```diff
@@ -0,0 +1,96 @@
...
+  "section",
+  "page",
+  "task",
+  "product",
+]);
+
+export function loadExamples(): ExampleRecord[] {
+  const records: ExampleRecord[] = [];
+
+  for (const folder of listDirectories(paths.content.examples)) {
```
ArakTaiRoth: This doesn't scan directories inside directories, so it'll miss `examples/workspace/case-detail` for example, there are others.

docs/src/scripts/content-generators/transforms/link-guidance.ts
```diff
@@ -0,0 +1,26 @@
...
+    const targets = g.appliesTo?.components ?? [];
+    for (const slug of targets) {
+      const list = byComponent.get(slug) ?? [];
+      list.push(g.id);
+      byComponent.set(slug, list);
+    }
+  }
+
+  return components.map((c) => {
+    const linked = byComponent.get(c.id);
```
ArakTaiRoth: This will bypass adding relatedGuidance to components like `app-footer`, `text-area`, `notification`, and `menu-button`. Because All those components are seen as `footer`, `textarea`, `notification-banner`, and `multi-action-button`. For example, I assume this was all applied to dcp-monorepo PR that you created, in there, there's a file `footer-no-white-space-below.json` that applies to `app-footer`, but in `footer.json` there's no relatedGuidance attached. Or `textarea-max-width.json` applied to `textarea`, but in `text-area.json` there's no relatedGuidance.

docs/src/scripts/content-generators/checks/prop-refs.ts
```diff
@@ -0,0 +1,95 @@
...
+      continue;
+    }
+
+    // Build the union of props across linked components that exist.
+    // Missing components are reported by component-refs.ts; skip them here
+    // to avoid double-reports.
+    const allValidProps = new Set<string>();
+    const knownComponents: string[] = [];
+    for (const componentId of linkedComponents) {
+      const component = componentMap.get(componentId);
```
ArakTaiRoth: This will have the same issue as the above issue.

docs/src/scripts/content-generators/loaders/framework-identifiers.ts
```diff
@@ -0,0 +1,188 @@
...
+  // the canonical webComponentTag stays the first one found.
+  const wcRoot = path.join(paths.code.libs, WC_COMPONENTS_DIR);
+  const tagToWcFolder = new Map<string, string>();
+  if (fs.existsSync(wcRoot)) {
+    for (const folder of fs.readdirSync(wcRoot, { withFileTypes: true })) {
+      if (!folder.isDirectory()) continue;
+      const tags = readWebComponentTags(path.join(wcRoot, folder.name));
+      if (tags.length === 0) continue;
+      const rec = getOrInit(folder.name);
+      rec.webComponentTag = tags[0];
```
ArakTaiRoth: This sets webComponentTag to the first item found in a component folder, but in some situations, the first item found is not actually the parent component. Take a look at `menu-button.json` in the dcp-monorepo PR. It extracts from `MenuButton.svelte`, but all web component, react classname, and angular selectory use some combination of `goa-menu-action`.

docs/src/scripts/content-generators/outputs/mcp-json.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export function writeMcpJson(records: AnyRecord[]): { written: number } {
+  const byCollection = new Map<string, AnyRecord[]>();
+  for (const rec of records) {
+    const list = byCollection.get(rec.collection) ?? [];
+    list.push(rec);
+    byCollection.set(rec.collection, list);
+  }
+
+  fs.mkdirSync(paths.output.mcp, { recursive: true });
```
ArakTaiRoth, line 14: Should maybe consider clearing out the old collection folders/files first before doing this. If there's a deleted file, or renamed file, the old one will still exist, so there will be stale data.

docs/src/scripts/content-generators/loaders/examples.ts
twjeffery: Yep fixed. `workspace/case-detail`, `workspace/dashboard`, and `workspace/index-page` come through now.

docs/src/scripts/content-generators/transforms/link-guidance.ts
twjeffery: Fixed. `footer-no-white-space-below` shows up on `footer.json` now, and `textarea-max-width` on `text-area.json`. Same for the other components you mentioned.

docs/src/scripts/content-generators/loaders/framework-identifiers.ts
twjeffery: Fixed. `menu-button.json` is clean now: `angularSelector: goab-menu-button`, no `goa-menu-action` or `GoabMenuAction` mixed in.

docs/src/scripts/content-generators/outputs/mcp-json.ts
twjeffery: Yep, the writer clears out old files before writing new ones now. Runs after the cross-checks, so the previous output stays put if anything errors.

twjeffery:
@ArakTaiRoth I pushed changes to address each of your comments. Five commits, one per thread you opened, so they're easier to map back to your comments. I will squash before merge. It's good to review again now.

docs/generated/component-apis/temporary-notification.json
ArakTaiRoth: This file still includes properties that are only available on the internal component, not the Controller. This is likely an issue with `extract-api.ts`

docs/src/scripts/extract-api.ts
ArakTaiRoth: The issue above is likely caused by this. Preserving entries that don't exist between extractions. But I don't know what the knock on effects could be from removing this, so, be careful?

twjeffery: Yep, that was it. Before removing it I ran extract-api with this part off to see what would change. Only 6 entries dropped across all the components and they were all the old temporary-notification ones. Nothing else was affected, so it's gone. The description fill above stayed since that wasn't the problem.

docs/generated/component-apis/temporary-notification.json
twjeffery: You're right, the earlier fix has extract-api reading from TemporaryNotificationCtrl.svelte now. These old web component props were sticking around from before that change because extract-api was holding onto anything it had picked up earlier. Removed

twjeffery:
@ArakTaiRoth Both addressed. Good to review again. Also regenerated the data on dcp-monorepo#342.

ArakTaiRoth (approved):

docs/src/scripts/content-generators/loaders/frontmatter.ts
```diff
@@ -0,0 +1,164 @@
...
+      } else {
+        const [sub, consumed] = parseObject(lines, nextIdx, nextIndent);
+        obj[key] = sub;
+        i = consumed;
+      }
+    } else if (valueText.startsWith("[") && valueText.endsWith("]")) {
+      obj[key] = parseInlineArray(valueText);
+      i++;
+    } else {
+      obj[key] = parseScalar(valueText);
```
vanessatran-ddi, line 116: It will not work if we have `accessibilityNodes` with `|`
If you run like this:
```
npx tsx docs/src/scripts/content-generators/index.ts
```
It will generate like the below:

Image

The `accessibilityNotes` become "|" instead of 

Image

docs/src/scripts/content-generators/loaders/framework-identifiers.ts
```diff
@@ -0,0 +1,173 @@
...
+    const svelteFile = path.join(paths.workspaceRoot, api.extractedFrom);
+    const tag = readSvelteCustomElementTag(svelteFile);
+    if (!tag) continue;
+
+    const react = tagToReact.get(tag);
+    const angularSelector = tagToAngular.get(tag);
+
+    bySlug.set(slug, {
+      webComponentTag: tag,
+      reactClassName: react?.classNames[0],
```
vanessatran-ddi: This introduces a bug. If you checked `dropdown-item.json` the `reactClassName` will be `GoabDropdownOption` instead of `GoabDropdownItem`

Image

docs/src/scripts/content-generators/loaders/framework-identifiers.ts
```diff
@@ -0,0 +1,173 @@
...
+}
+
+function buildAngularWrapperIndex(): Map<string, string> {
+  const byTag = new Map<string, string>();
+  const root = path.join(paths.code.libs, ANGULAR_LIB_DIR);
+  if (!fs.existsSync(root)) return byTag;
+
+  for (const tsFile of walkSourceFiles(root, ".ts")) {
+    const content = fs.readFileSync(tsFile, "utf8");
+    const selectorMatch = content.match(/selector\s*:\s*["']goab-([a-z0-9-]+)["']/);
```
vanessatran-ddi, line 96: This assumes that angular selector is equal vs WC tag, but it is not true for `goab-temporary-notification-ctrl`
So the consequence is `temporary-notification.json` has the reactClassName correctly,  webComponentTag correctly, but missing angular selector. 

Image

twjeffery: Good catch. It was assuming the Angular selector matched the web component tag, which isn't true for temporary-notification-ctrl. It reads the actual tag from the wrapper template now. I also made sure the input variants still map to goab-input, since they share the goa-input tag.

docs/src/scripts/content-generators/loaders/framework-identifiers.ts
twjeffery: Yeah, it was grabbing the first export in the file, which is the deprecated GoabDropdownOption. It picks the class that matches the tag now, so dropdown-item gets GoabDropdownItem.

docs/src/scripts/content-generators/loaders/frontmatter.ts
twjeffery: Good catch. The frontmatter parser wasn't handling the | block style, so it just grabbed the "|" itself. It reads the indented lines now, so the multi-line accessibilityNotes come through.

twjeffery:
@vanessatran-ddi All three fixed. Good to review again.

vanessatran-ddi (approved):
Looks good!

Issue #3346, opened 2026-01-19
Add default horizontal scroll behaviour to the table component for overflow content


Spark450:
## Summary
Update to the table component to support horizontal overflow by default. When the table’s content exceeds the available horizontal space (e.g., many columns or wide cell content), the table should allow horizontal scrolling rather than clipping, wrapping unexpectedly, or breaking layout.

## Reference implementation
See [workspace template](https://workspace-demo-v4.netlify.app/cases#tab-0)

## Desired behaviour (default)
- `goa-table` should handle horizontal overflow out of the box.
- When content width > container width:
  - a horizontal scrollbar is available (within the table/container region)
  - Shadow appears on vertical edge of table where there is more content to show

- When content fits:
  - no horizontal scrollbar is shown
  - layout remains unchanged

## Acceptance criteria
- Table supports horizontal scrolling by default when content overflows horizontally.
- No additional consumer configuration (props) is required to enable horizontal scroll.
- Existing tables that do not overflow render unchanged.

PR #3966, opened 2026-05-20
feat(#3346): add horizontal scrolling to Table

bdfranck:
This PR adds horizontal scrolling with inner shadows to the Table when the content exceeds the available space.

table-horz-scroll

### DIFF
--- a/libs/web-components/src/components/table/Table.svelte
+++ b/libs/web-components/src/components/table/Table.svelte
@@ -11,7 +11,7 @@
 />
 
 <script lang="ts">
-  import { onMount } from "svelte";
+  import { onDestroy, onMount } from "svelte";
   import { calculateMargin } from "../../common/styling";
   import { typeValidator, toBoolean, dispatch } from "../../common/utils";
   import type { Spacing } from "../../common/styling";
@@ -67,9 +67,13 @@
   // Private
 
   let _rootEl: HTMLElement;
+  let _scrollEl: HTMLElement;
   let _isTableRoot: boolean = false;
   let _sorts: SortEntry[] = [];
   let _headings: NodeListOf<Element> | undefined;
+  let _horizontalScrollPos: "left" | "middle" | "right" | null = null;
+  let _initSortTimeout: ReturnType<typeof setTimeout>;
+  let _initScrollTimeout: ReturnType<typeof setTimeout>;
   const MAX_SORTS = 2;
 
   // Reactive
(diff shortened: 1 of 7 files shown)


libs/web-components/src/components/table/Table.svelte
```diff
@@ -258,11 +271,39 @@
...
+        _scrollEl.scrollWidth - _scrollEl.scrollLeft - _scrollEl.clientWidth,
+      ) < 1
+    ) {
+      _horizontalScrollPos = "right";
+    } else {
+      _horizontalScrollPos = "middle";
+    }
+  }
+
+  function handleHorizontalScroll() {
```
vanessatran-ddi, line 299: do we still need `handleHorizontalScroll`, should we use directly the function `updateHorizontalScrollShadows()`?

libs/web-components/src/components/table/Table.svelte
```diff
@@ -295,11 +342,55 @@
...
+    width: var(--goa-space-xs);
+    opacity: 0;
+    pointer-events: none;
+    z-index: 2;
+    transition: opacity 0.15s ease;
+  }
+
+  .goatable::before {
+    left: 0;
+    background: linear-gradient(to right, rgba(0, 0, 0, 0.08), transparent);
```
vanessatran-ddi, line 379: Do we have a token for those shadows? Also do you consider to handle dark mode?

vanessatran-ddi, line 379: Curious question: is there a reason we do not use `box-shadow: inset` like the way we do vs Modal

twjeffery (commented):
@bdfranck it looks good overall, but there's a bit of a UX question for us. 

Here's what we have on the design system website (as a wrapper that wraps the table to give it this shadow on scroll):

Image

https://github.com/user-attachments/assets/7569eee0-05b7-4336-a878-c9c9c5ce657f

This has the table itself scrolling to the edge of it's parent container and then showing a shadow. The one you made has the table static and scrolling into itself. Not sure both are valid at different points, or if we want to pick a direction here, but we can discuss.

### Other things:
- The scrollbar is always at the bottom of the table, when there is a table that's taller than the screen height, you don't see the horizontal scroll bar until you scroll to the bottom of the table. Not sure if it's reasonable, but it would be nice if that scrollbar stayed with the bottom of your screen in that case.
- When you tab through the page content, the whole table itself shows up as focused in the tab order, I don't think it should be

bdfranck:
@vanessatran-ddi Would it make sense to add a horizontal variant to your Scroll Panel to handle things like tables? 🤔 How hard would that be to implement?

vanessatran-ddi:
> @vanessatran-ddi Would it make sense to add a horizontal variant to your Scroll Panel to handle things like tables? 🤔 How hard would that be to implement?

Yes it makes sense. It should have a scrollable for horizontal direction too (now is vertical only), I can put it in another PR as now mine is too big, but I am pretty sure it will solve @twjeffery  comment on the level of workspace layout.

bdfranck:
@vanessatran-ddi Sounds good! I'll close this PR and remove it from the issue.

PR #3979, opened 2026-05-25
chore: regenerate stale generated docs file to clean baseline

twjeffery:
Running the docs generators surfaced one committed output (`temporary-notification.json`) that had drifted from its Svelte source. This catches it up so dev is a clean baseline.

### DIFF
--- a/docs/generated/component-apis/temporary-notification.json
+++ b/docs/generated/component-apis/temporary-notification.json
@@ -63,7 +63,7 @@
           "type": "string",
           "required": false,
           "default": "",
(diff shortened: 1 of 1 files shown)


twjeffery:
Closing this in favor of #4021.

This refreshed temporary-notification.json to a clean baseline, but the file was actually being generated from the wrong element (the positioning controller instead of the notification). #4021 fixes the extraction so it regenerates correctly, which covers what this PR was doing and stops it drifting back. So this one is no longer needed.

PR #3980, opened 2026-05-25
chore(#3868): add pre-commit hook to keep generated docs files in sync

### LINKED ISSUE #3868: Add pre-commit git hook to keep generated API files in sync with components
## Background

In #3513 we wired API extraction into `npm run build` in docs and kept the generated JSON files committed. The assumption was that anyone touching a component would also build docs and commit the refreshed files alongside their change.

In practice, most PRs touching components don't touch docs. The generated files on dev have fallen behind, and anyone building docs on any branch ends up with a bunch of file changes they didn't make.

## Proposed approach

A pre-commit git hook that runs the docs build when a commit touches a component, and blocks the commit if the generated files aren't up to date. That way any commit touching a component carries the refreshed files with it, and the diffs in PRs go back to being useful to review.

## Open questions for implementation

- Making sure the hook gets installed for everyone automatically when they clone the repo
- Whether to scope it to commits that touch components, so it doesn't slow down unrelated commits
- What happens if someone bypasses the hook, and whether GitHub should check the same thing as a backup

## Done when

- [ ] Hook installed automatically for everyone
- [ ] Hook blocks commits when generated files are out of sync
- [ ] Documented in ARCHITECTURE.md

## Context

- #3513 (original PR that set this up)

twjeffery:
## What this does

The docs site commits two generated files: the component API JSON (`docs/generated/component-apis/`) and the search index (`docs/public/search-index.json`). They only refresh when someone runs the docs build, and most component PRs never touch docs, so the committed files quietly fall behind the Svelte and MDX they come from. This is #3868.

This adds a pre-commit hook that regenerates those outputs when you stage a file that feeds them, and blocks the commit if the committed versions are stale. If nothing generator-related is staged, the hook does nothing.

## What's in here

1. **The hook** at `.githooks/pre-commit`. Watches the three libs source roots that feed `extract-api` (`web-components/src/components`, `react-components/src/lib`, `angular-components/src/lib/components`) and `docs/src/content/**` (MDX frontmatter, feeds `build:search-index`). Because the generators are deterministic, a file that is already fresh comes back byte-identical, so there is no friction when you have done the right thing.
2. **Determinism fix** in `build-search-index.ts`. The search index listed files in whatever order the filesystem returned, which differs between macOS and Linux. It now sorts them so the output is identical on every machine. Without this the hook would flip-flop between contributors.
3. **Docs** in `docs/ARCHITECTURE.md`: what the hook does, the one-time install, and how to bring a new generator under it.

## Install

Run `npm install` at the repo root. A `prepare` script points `core.hooksPath` at `.githooks` (the hook), and a `postinstall` runs `npm install --prefix docs` so you also get the docs dependencies the generators need. To set the hook by hand: `git config core.hooksPath .githooks`.

## How to test

1. `git config core.hooksPath .githooks`
2. Edit a JSDoc comment in any Svelte component, stage it, try to commit. The hook blocks and tells you to regenerate.
3. Run `npm run extract-api` in `docs/`, stage the JSON, commit again. It passes.
4. Same idea with an MDX frontmatter change and `npm run build:search-index`.
5. Edit a non-source file and commit. No hook, no friction.

## Deferred to follow-ups

A CI check that does the same thing on GitHub (a backstop for bypassed or uninstalled hooks), collapsing the generated files in PR diffs, and a merge driver for conflicts. All build on this and are out of scope here.

Closes #3868

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,0 +1,84 @@
+#!/usr/bin/env bash
+#
+# Pre-commit hook: keep committed docs generated files in sync with their sources.
+#
+# When a commit stages source that feeds a docs generator, this regenerates the
+# docs outputs and blocks the commit if the staged outputs are stale. The
+# generators are deterministic, so when the staged outputs are already fresh this
+# is a no-op and the commit proceeds with no change.
+#
+# Install once per clone:
+#   git config core.hooksPath .githooks
+#
+# Covers:
+#   docs/generated/component-apis/*.json   from libs/*/src/components/** (Svelte JSDoc)
+#   docs/public/search-index.json          from docs/src/content/** (MDX frontmatter)
+#
+# To add a new generator, extend SOURCE_REGEX and OUTPUTS below and run it in the
+# regenerate step. See docs/ARCHITECTURE.md.
+#
+# Limitation: this validates the working tree, not the staged snapshot. With
+# partial staging (git add -p, or staging some edits while leaving others
+# unstaged) the generators read working-tree source, which can differ from what
+# is being committed. Accepted for simplicity; see docs/ARCHITECTURE.md.
+
+set -euo pipefail
+
+cd "$(git rev-parse --show-toplevel)"
+
+# Sources that feed a generator, and the output paths those generators own.
+# extract-api scans three libs roots: Svelte components (web-components/src/components),
+# React wrappers (react-components/src/lib), and Angular wrappers
+# (angular-components/src/lib/components). build:search-index scans MDX in docs/src/content.
+SOURCE_REGEX='^(libs/(web-components/src/components|react-components/src/lib|angular-components/src/lib/components)/|docs/src/content/)'
+OUTPUTS=(docs/generated/component-apis docs/public/search-index.json)
+
+# 1. Only act when staged changes touch a generator source.
+# Capture to a variable rather than piping to grep: under `set -o pipefail`, grep -q
+# can close the pipe early, making git diff exit with SIGPIPE and silently skipping
+# the check on very large commits.
+staged="$(git diff --cached --name-only)"
+if ! grep -Eq "$SOURCE_REGEX" <<<"$staged"; then
+  exit 0
+fi
+
+# 2. The generators need docs dependencies (tsx). Without them we cannot verify
+#    that committed outputs match the source, so block the commit with a clear,
+#    actionable message instead of letting a potentially-stale commit through.
+if [ ! -x docs/node_modules/.bin/tsx ]; then
+  echo "pre-commit: cannot verify generated docs files: docs dependencies missing." >&2
+  echo "            run 'npm install' in docs/ and commit again." >&2
+  exit 1
+fi
+
+echo "pre-commit: checking generated docs files are up to date..." >&2
+
+# 3. Regenerate. If a generator cannot run (for example a missing dependency),
+#    skip with a clear warning instead of blocking the commit or printing a stack.
+genlog="$(mktemp)"
+if ! ( cd docs && npm --silent run extract-api && npm --silent run build:search-index ) >"$genlog" 2>&1; then
+  echo "pre-commit: skipping freshness check: a generator did not run." >&2
+  echo "            ensure dependencies are installed (npm install at the repo root and in docs/)." >&2
+  sed 's/^/            | /' "$genlog" >&2 || true
+  rm -f "$genlog"
+  exit 0
+fi
+rm -f "$genlog"
+
+# 4. Stale if regenerated outputs differ from what is staged (modified tracked
+#    files) or if a brand-new output was produced (untracked, e.g. a new
+#    component's JSON that was never generated).
+stale=0
+if ! git diff --quiet -- "${OUTPUTS[@]}"; then stale=1; fi
+if [ -n "$(git ls-files --others --exclude-standard -- "${OUTPUTS[@]}")" ]; then stale=1; fi
+
+if [ "$stale" -ne 0 ]; then
+  echo "pre-commit: generated docs files are out of date." >&2
+  echo "            in docs/, run 'npm run extract-api' and 'npm run build:search-index'," >&2
+  echo "            stage the result, and commit again." >&2
+  echo "" >&2
+  git status --short -- "${OUTPUTS[@]}" >&2
+  exit 1
+fi
+
+exit 0
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -13,6 +13,7 @@
 #
 # Covers:
 #   docs/generated/component-apis/*.json   from Svelte components + React/Angular wrappers
+#                                          + shared types and controllers (libs/common)
 #   docs/public/search-index.json          from docs/src/content/** (MDX frontmatter)
 #
 # To add a new generator, extend SOURCE_REGEX and OUTPUTS below and run it in the
@@ -28,10 +29,11 @@ set -euo pipefail
 cd "$(git rev-parse --show-toplevel)"
 
 # Sources that feed a generator, and the output paths those generators own.
-# extract-api scans three libs roots: Svelte components (web-components/src/components),
-# React wrappers (react-components/src/lib), and Angular wrappers
-# (angular-components/src/lib/components). build:search-index scans MDX in docs/src/content.
-SOURCE_REGEX='^(libs/(web-components/src/components|react-components/src/lib|angular-components/src/lib/components)/|docs/src/content/)'
+# extract-api scans four libs roots: Svelte components (web-components/src/components),
+# React wrappers (react-components/src/lib), Angular wrappers
+# (angular-components/src/lib/components), and shared types plus imperative-API
+# controllers (common/src/lib). build:search-index scans MDX in docs/src/content.
+SOURCE_REGEX='^(libs/(web-components/src/components|react-components/src/lib|angular-components/src/lib/components|common/src/lib)/|docs/src/content/)'
 OUTPUTS=(docs/generated/component-apis docs/public/search-index.json)
 
 # 1. Only act when staged changes touch a generator source.
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -4,14 +4,15 @@
 #
 # When a commit stages source that feeds a docs generator, this regenerates the
 # docs outputs and blocks the commit if the staged outputs are stale. The
-# generators are deterministic, so when the staged outputs are already fresh this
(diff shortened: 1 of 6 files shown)


.githooks/pre-commit
```diff
@@ -0,0 +1,81 @@
...
+# partial staging (git add -p, or staging some edits while leaving others
+# unstaged) the generators read working-tree source, which can differ from what
+# is being committed. Accepted for simplicity; see docs/ARCHITECTURE.md.
+
+set -euo pipefail
+
+cd "$(git rev-parse --show-toplevel)"
+
+# Sources that feed a generator, and the output paths those generators own.
+SOURCE_REGEX='^(libs/[^/]+/src/components/|docs/src/content/)'
```
ArakTaiRoth: `extract-api.ts` also reads from the react-components and angular-components libraries, should probably watch those as well.

.githooks/pre-commit
```diff
@@ -0,0 +1,81 @@
...
+# the check on very large commits.
+staged="$(git diff --cached --name-only)"
+if ! grep -Eq "$SOURCE_REGEX" <<<"$staged"; then
+  exit 0
+fi
+
+# 2. The generators need docs dependencies (tsx). A contributor who only works in
+#    libs/ may not have them installed, which is exactly when this hook fires.
+#    Skip with guidance rather than crashing the commit over a tooling gap.
+if [ ! -x docs/node_modules/.bin/tsx ]; then
```
ArakTaiRoth, line 53: I don't think we should skip this. It's entirely possible that someone works in docs, but never runs `npm i`. It doesn't technically need to be run in order to make a change and commit that change. I think keep the error message that they need to run it, but I think we should still block the commit.

.githooks/pre-commit
twjeffery: Broadened to all three and updated the watched-paths list in ARCHITECTURE.md.

.githooks/pre-commit
twjeffery: Switched it to block instead of skip

.githooks/pre-commit
```diff
@@ -0,0 +1,84 @@
...
+  echo "pre-commit: cannot verify generated docs files: docs dependencies missing." >&2
+  echo "            run 'npm install' in docs/ and commit again." >&2
+  exit 1
+fi
+
+echo "pre-commit: checking generated docs files are up to date..." >&2
+
+# 3. Regenerate. If a generator cannot run (for example a missing dependency),
+#    skip with a clear warning instead of blocking the commit or printing a stack.
+genlog="$(mktemp)"
```
vanessatran-ddi, line 63: This should `exit 1` when the script fail, otherwise it will pass silently

.githooks/pre-commit
```diff
@@ -0,0 +1,84 @@
...
+# 4. Stale if regenerated outputs differ from what is staged (modified tracked
+#    files) or if a brand-new output was produced (untracked, e.g. a new
+#    component's JSON that was never generated).
+stale=0
+if ! git diff --quiet -- "${OUTPUTS[@]}"; then stale=1; fi
+if [ -n "$(git ls-files --others --exclude-standard -- "${OUTPUTS[@]}")" ]; then stale=1; fi
+
+if [ "$stale" -ne 0 ]; then
+  echo "pre-commit: generated docs files are out of date." >&2
+  echo "            in docs/, run 'npm run extract-api' and 'npm run build:search-index'," >&2
```
vanessatran-ddi: at line 59 above we already run and generate files, so at this step when it failed, we can suggest to `git add` those stale files instead of asking them to re-run the script. Should we update this message to be `the regenerated files are already in your working tree, run git add docs/generated/component-apis docs/public/search-index.json and commit again`

.githooks/pre-commit
```diff
@@ -0,0 +1,84 @@
...
+# When a commit stages source that feeds a docs generator, this regenerates the
+# docs outputs and blocks the commit if the staged outputs are stale. The
+# generators are deterministic, so when the staged outputs are already fresh this
+# is a no-op and the commit proceeds with no change.
+#
+# Install once per clone:
+#   git config core.hooksPath .githooks
+#
+# Covers:
+#   docs/generated/component-apis/*.json   from libs/*/src/components/** (Svelte JSDoc)
```
vanessatran-ddi: the script reads from Svelte components + React/Angular wrappers, so we should update this comment. Suggestion `docs/generated/component-apis/*.json   from Svelte components + React/Angular wrappers`

docs/ARCHITECTURE.md
```diff
@@ -71,6 +71,55 @@ The site resolves monorepo packages directly via aliases in `astro.config.mjs`:
...
+
+When a commit stages a file that feeds a generator, the hook regenerates the outputs and blocks the commit if the committed versions are stale.
+
+- Sources it watches:
+  - `libs/web-components/src/components/**` (Svelte JSDoc, feeds `extract-api`)
+  - `libs/react-components/src/lib/**` (React wrappers, feed `extract-api`)
+  - `libs/angular-components/src/lib/components/**` (Angular wrappers, feed `extract-api`)
+  - `docs/src/content/**` (MDX frontmatter, feeds `build:search-index`)
+- If none of those are staged, the hook does nothing.
+- If the regenerated output matches what you staged, the commit goes through. The generators are deterministic, so a file that is already fresh comes back byte-identical and nothing is slowed down.
```
vanessatran-ddi: \"nothing is slowed down\" may not be correct, everytime we git commit the script still runs no matter output changed or not, so this statement is a bit confusing. I would suggest to change it like this \"..comes back byte-identical, so the commit goes through with no files changes to stage.\"

docs/ARCHITECTURE.md
```diff
@@ -71,6 +71,55 @@ The site resolves monorepo packages directly via aliases in `astro.config.mjs`:
...
+- If they differ, the commit is blocked, naming the file that drifted and what to run.
+
+### Install it (one time per clone)
+
+```bash
+git config core.hooksPath .githooks
+```
+
+That points git at the tracked `.githooks/` directory. There is no automatic installer. It is one command, run once, worth doing right after cloning.
```
vanessatran-ddi: actually when I read about it, I think there is a way to automatic install. We can add `prepare` step on this project package.json, like this `prepare: \"git config core.hooksPath .githooks || true\"`, then when ppl run `npm install`, this will be automatically installed at the final step

.githooks/pre-commit
twjeffery: Updated to your wording: `from Svelte components + React/Angular wrappers`.

.githooks/pre-commit
twjeffery: Right, it runs on every commit regardless. Reworded to say the commit goes through with no new files to stage, and that the generators still run but come back byte-identical when the file is already fresh.

.githooks/pre-commit
twjeffery: Right, the hook already regenerates them at that point, so re-running was redundant. Updated the message to point you at `git add` for the files already in your working tree, and wired it to the OUTPUTS list so the paths stay correct if we add a generator later.

docs/ARCHITECTURE.md
twjeffery: Changed it to `exit 1`. If a generator can't run we can't verify the outputs are fresh, so blocking makes more sense than passing silently. It also lines up with the missing-deps check above, which already blocks.

docs/ARCHITECTURE.md
twjeffery: Added the `prepare` to `docs/package.json` like you suggested, so it sets up on `npm install` in docs/.

vanessatran-ddi: Sorry I do not clarify it clearer. We should add this under root folder, package.json file, so when anyone under root folder and run `npm install` the hook will be installed automatically and quietly for them. under `docs` will require developers `cd docs` and run `npm install` there, which I do not think anyone would do.

twjeffery: Moved it to the root package.json.

twjeffery: @vanessatran-ddi One thing I want to check with you: a root npm install doesn't include the docs deps, so the hook would block anyone working only in libs/ who doesn't have docs set up. To avoid that, I had it skip with a warning on libs-only commits and only block when you're staging a docs/src/content change. Does that sound right?

willcodeforcoffee (approved):
I've tested a few scenarios. I have a few positive comments:
- I appreciate that it runs fairly quickly
- I also like that it leaves the updated docs behind so you can amend your commit
- I'm glad it runs in Bash so doesn't require more scripting languages, this also helps it run quickly
- It blocked my commit when I added a param, or changes doc comments
- It did not block commits when I updated README, or added comments in the middle of a file

Looks great! Thanks @twjeffery !

docs/ARCHITECTURE.md
vanessatran-ddi: Yes Tom. `root` is the right place for `prepare`. Instead of skip vs block on missing docs deps, can we add a root `postinstall: npm install --prefix docs` so one root install gives everyone `tsx` and the hook always runs?

.githooks/pre-commit
```diff
@@ -0,0 +1,93 @@
...
+#    freshness check is the backstop for that case.
+if [ ! -x docs/node_modules/.bin/tsx ]; then
+  if grep -Eq '^docs/src/content/' <<<"$staged"; then
+    echo "pre-commit: cannot verify generated docs files: docs dependencies missing." >&2
+    echo "            run 'npm install' in docs/ and commit again." >&2
+    exit 1
+  fi
+  echo "pre-commit: skipping generated-docs check: docs dependencies not installed." >&2
+  echo "            run 'npm install' in docs/ to check component changes too." >&2
+  exit 0
```
vanessatran-ddi: so if we add `postinstall` inside root package.json, here we can `exit 1`, and block it instead of skipping.

docs/ARCHITECTURE.md
twjeffery: Done. Added `postinstall: npm install --prefix docs` to the root package.json, so a single root install gives everyone the docs deps and the hook always has what it needs. I also removed the skipping, so now it just blocks if the deps are missing.

.githooks/pre-commit
twjeffery: Yep, changed it to block (`exit 1`) now that the postinstall makes sure the deps are always there. No more skipping.

twjeffery:
While addressing review on #4021 I noticed the hook wasn't watching libs/common, which extract-api now reads for the temporary notification method docs. Added it to the watched paths.

vanessatran-ddi (approved):
Looks good!

PR #3993, opened 2026-05-29
Add a Markdown bundle output for AI tools

### LINKED ISSUE #3960: Automate the content generator and add a Markdown bundle output
Follow-up to #3770. With the generator producing MCP data from the docs site, the next step is making the MCP stay current automatically as content changes. The same generator can also produce a Markdown bundle for AI tools that don't speak MCP.

## Build automation

The generator runs manually today (`npx tsx docs/src/scripts/content-generators/index.ts`) and the sync to dcp-monorepo is a manual bash script. To deliver on "MCP stays accurate as content changes":

- Generator runs as part of `docs:build`
- Sync becomes a CI job that opens a PR against dcp-monorepo when content changes on `dev`
- Freshness / CI gates on the generated outputs so stale artifacts can't ship

## Markdown bundle

A second output target alongside the MCP data: a Markdown bundle for AI tools that don't speak MCP. Same source, different distribution.

- Per-component Markdown file mirroring the website's component pages
- Per-foundation Markdown file
- An entry-point document for tools loading the bundle as a whole
- Output at `docs/generated/md-bundle/`, gitignored
- Distribution path decided. Options to weigh: served from the docs site at a known URL, published as a GitHub release artifact, or both.

twjeffery:
A second output from the docs content generator: a downloadable Markdown bundle for AI tools that don't use the MCP. It covers components, examples, foundations, get started, and product types, in per-framework sets (React, Angular, Web Components).

Part of #3960.

## Why

The MCP covers tools that can connect to it. This covers the rest, plus the case where someone wants to paste design system context in up front. An AI gets the same picture either way.

## How it works

- Same source as the MCP output (the docs content), so it stays in step with the components. No hand-maintained Markdown.
- Per framework, so a tool loads only what it uses. Properties and example code are framework-specific; guidance and foundations are copied into each set so it stands on its own.
- Each set opens with an index and a rough size, so a tool can load the index plus the files it needs instead of everything.
- Output goes to `docs/generated/md-bundle/` and is gitignored, like the MCP output.

## In this PR

- The Markdown output target (`outputs/md-bundle.ts`) and the generator wiring.
- A test suite, and an output validator the generator runs after writing, so a malformed bundle fails the run instead of shipping quietly.

## Not in this PR

- Where the bundle is hosted and how teams download it. Separate decision.
- Wiring the generator into the build and CI. Follow-up. Two things ride along with that:
  - The output validator runs as part of the generator, so it catches problems when you run it, but it isn't wired into CI yet. For now it doesn't automatically guard the branch.
  - The validator uses the `yaml` library to check frontmatter, which currently resolves as a transitive dependency. I'll make it a direct dependency when the gate gets wired into the build.

## Try it

```
npx tsx docs/src/scripts/content-generators/index.ts
```

Output lands in `docs/generated/md-bundle/{react,angular,web-components}/`. Worth opening: `react/components/button.md`, `react/get-started/developers__dark-mode-theme.md`, `react/examples/error-pages.md`, `react/index.md`.

```
npm run --prefix docs test
npm run --prefix docs verify-bundle
```

## Notes for reviewers

A couple of things the bundle surfaced in the docs content, worth fixing separately:

- Button guidance says use tertiary for "cancel", but the address example uses secondary.
- The "Form" page is titled "Form" but documents the public form container.

One known gap: the bundle names types like `Spacing` and the icon type but not their allowed values, so an AI only sees the values that show up in examples. Worth a follow-up.

### DIFF
--- a/docs/src/scripts/content-generators/outputs/md-bundle.ts
+++ b/docs/src/scripts/content-generators/outputs/md-bundle.ts
@@ -1,6 +1,6 @@
 import * as fs from "fs";
 import * as path from "path";
-import { paths } from "../config";
+import { paths, WEB_COMPONENT_ONLY } from "../config";
 import type {
   AnyRecord,
   ComponentRecord,
@@ -27,6 +27,13 @@ export interface FrameworkTarget {
   apiKey: string;
   /** Human label used in headings. */
   label: string;
+  /**
+   * Heading for the content-insertion section. Web components have native
+   * slots; React passes content as ReactNode, Angular via Template Ref. Naming
+   * it per framework stops an AI reaching for a `slot` attribute where the
+   * framework uses a different mechanism.
+   */
+  slotsLabel: string;
   /** Frontmatter key for this framework's component identifier. */
   frontmatterKey: string;
   /** ComponentRecord field holding the identifier (e.g. `GoabButton`). */
@@ -47,6 +54,7 @@ const FRAMEWORKS: FrameworkTarget[] = [
     id: "react",
     apiKey: "react",
     label: "React",
+    slotsLabel: "ReactNode",
     frontmatterKey: "react",
     identifierKey: "reactClassName",
     sources: [{ file: "react.tsx", lang: "tsx" }],
@@ -56,6 +64,7 @@ const FRAMEWORKS: FrameworkTarget[] = [
     id: "angular",
     apiKey: "angular",
     label: "Angular",
+    slotsLabel: "Template Ref",
     frontmatterKey: "angular",
     identifierKey: "angularSelector",
     sources: [
@@ -68,6 +77,7 @@ const FRAMEWORKS: FrameworkTarget[] = [
     id: "web-components",
     apiKey: "webComponents",
     label: "Web component",
+    slotsLabel: "Slots",
     frontmatterKey: "webComponent",
     identifierKey: "webComponentTag",
     sources: [{ file: "web-components.html", lang: "html" }],
@@ -102,6 +112,19 @@ interface FrameworkApi {
 }
 interface ApiBlob {
   frameworks?: Record<string, FrameworkApi>;
+}
+
+/**
+ * Components for a framework's set. Web components carry the full set; React and
+ * Angular drop the web-component-only utilities (no wrapper there), so an AI is
+ * not shown a component it cannot use in that framework.
+ */
+export function componentsForTarget(
+  components: ComponentRecord[],
+  target: FrameworkTarget,
+): ComponentRecord[] {
+  if (target.id === "web-components") return components;
+  return components.filter((c) => !WEB_COMPONENT_ONLY.has(c.id));
 }
 
 /** Generate the Markdown bundle: one self-contained set per framework. */
@@ -155,6 +178,7 @@ export function writeMdBundle(records: AnyRecord[]): { written: number } {
   let written = 0;
   for (const target of targets) {
     const root = path.join(paths.output.mdBundle, target.id);
+    const targetComponents = componentsForTarget(components, target);
     let chars = 0;
     let fileCount = 0;
     // Write a file and tally its size, so the index can report the set's weight.
@@ -165,7 +189,7 @@ export function writeMdBundle(records: AnyRecord[]): { written: number } {
       written++;
     };
 
-    for (const c of components) {
+    for (const c of targetComponents) {
       emit(
         path.join(root, "components", filename(c.id)),
         renderComponent(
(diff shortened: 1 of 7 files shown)


twjeffery:
**Quick flag before review:** CodeQL shows one high-severity alert, on the table-cell helper in `md-bundle.ts` where we strip stray HTML tags before writing Markdown. From what I understand it's a false positive: it only runs over our own docs content at build time, and the output is Markdown, not a web page, so there's nothing to inject into. If you agree, I'll dismiss it with that reasoning.

docs/src/scripts/content-generators/outputs/md-bundle.ts
```diff
@@ -0,0 +1,1082 @@
...
+  // Properties
+  const api = (c.api as ApiBlob | undefined)?.frameworks?.[target.apiKey];
+  const propsTable = api?.props?.length ? apiTable(api.props, true) : "";
+  const eventsTable = api?.events?.length ? apiTable(api.events, false) : "";
+  const slotsTable = api?.slots?.length ? apiTable(api.slots, false) : "";
+  if (propsTable || eventsTable || slotsTable) {
+    out.push("## Properties");
+    if (propsTable) out.push(propsTable);
+    if (eventsTable) out.push("### Events", eventsTable);
+    if (slotsTable) out.push("### Slots", slotsTable);
```
vanessatran-ddi: This should depend on framework/library. For example Angular, on our website we called it Template Ref, and React is ReactNode, call it `slots` will make the Agent misunderstand using `slot` attribute instead. 

Image

Image

docs/src/scripts/content-generators/outputs/md-bundle.ts
```diff
@@ -0,0 +1,1082 @@
...
+    let fileCount = 0;
+    // Write a file and tally its size, so the index can report the set's weight.
+    const emit = (file: string, md: string): void => {
+      writeFile(file, md);
+      chars += md.length;
+      fileCount++;
+      written++;
+    };
+
+    for (const c of components) {
```
vanessatran-ddi: we should filter out some components such as `FocusTrap` that doesnt have react and angular wrapper, right now I see focus-trap.md file generated

docs/src/scripts/content-generators/outputs/md-bundle.ts
```diff
@@ -0,0 +1,1082 @@
...
+  componentNameById: Map<string, string>,
+): string {
+  const fm: [string, unknown][] = [
+    ["id", pt.id],
+    ["title", pt.title],
+    ["status", pt.status],
+    ["tags", pt.tags],
+  ];
+  const out = [frontmatter(fm), `# ${pt.title}`];
+  if (pt.summary) out.push(mdxBodyProse(pt.summary));
```
vanessatran-ddi: Should we guard `mdxBodyProse(pt.summary)` not empty before pushing, then it will be aligned with others functions, such as line 480 above

docs/src/scripts/content-generators/verify-bundle.ts
```diff
@@ -0,0 +1,193 @@
...
+
+export interface Violation {
+  file: string;
+  line?: number;
+  message: string;
+}
+
+const SENTINEL = "@@CODEMASK@@";
+const TAG =
+  /<\/?(goa-[a-z]+|[A-Z][A-Za-z0-9]+|div|span|p|h[1-6]|ul|ol|li|table|thead|tbody|tr|td|th|br|img|pre)\b/;
```
vanessatran-ddi: missing other tag such as `a`, `strong`, `em`

docs/src/scripts/content-generators/outputs/md-bundle.ts
twjeffery: Done. React shows ReactNode, Angular shows Template Ref, Web Components stays Slots

docs/src/scripts/content-generators/outputs/md-bundle.ts
twjeffery: Done. Dropped FocusTrap and Scrollable from the React and Angular sets (turned out they're the only two with no wrapper), kept them in Web Components

docs/src/scripts/content-generators/outputs/md-bundle.ts
twjeffery: Good catch, fixed

docs/src/scripts/content-generators/verify-bundle.ts
twjeffery: Added those plus `b`, `i`, `code`, `blockquote`, `hr`

vanessatran-ddi (approved):
Looks good!

willcodeforcoffee (approved):
Since CodeQL doesn't seem to be a real issue for us and won't break `dev` or `main` everything else LGTM!

PR #3996, opened 2026-05-29
chore: prevent PR check from hanging on playwright browser download

vanessatran-ddi:
# Before (the change)

# After (the change)

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -30,11 +30,8 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Prepare git base ref
-        run: |
-          git config --global --add safe.directory "$GITHUB_WORKSPACE"
-          git fetch --no-tags --depth=1 origin \
-            "+refs/heads/${{ github.base_ref }}:refs/remotes/origin/${{ github.base_ref }}"
+      - name: Allow git to use the container workspace
+        run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
 
       - uses: actions/setup-node@v6
         with:
@@ -53,15 +50,8 @@ jobs:
           npx nx affected --target=build --base=origin/${{ github.base_ref }}
           npx nx affected --target=post --base=origin/${{ github.base_ref }}
 
-      - name: Check for build output
-        id: dist
+      - name: Test
         run: |
           if [ -d "./dist" ]; then
-            echo "exists=true" >> "$GITHUB_OUTPUT"
-          else
-            echo "exists=false" >> "$GITHUB_OUTPUT"
+            npm run test:pr
           fi
-
-      - name: Test
-        if: steps.dist.outputs.exists == 'true'
-        run: npm run test:pr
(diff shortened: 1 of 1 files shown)


.github/workflows/pull-request.yml
```diff
@@ -13,12 +13,29 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    timeout-minutes: 30
+    container:
+      image: mcr.microsoft.com/playwright:v1.59.0-noble
+
+    env:
+      # the playwright npm postinstall not to download them during `npm ci`.
+      PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD: "1"
```
ArakTaiRoth, line 22: By skipping this download, this means that if we update playwright in the `package.json`, we also have to update the version here. Not necessarily bad, but does introduce a future problem.

.github/workflows/pull-request.yml
```diff
@@ -36,9 +53,15 @@ jobs:
...
-            npx playwright install --with-deps
-            npm run test:pr
+            echo "exists=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "exists=false" >> "$GITHUB_OUTPUT"
           fi
+
+      - name: Test
+        if: steps.dist.outputs.exists == 'true'
+        run: npm run test:pr
```
ArakTaiRoth: Why not just move this running of the test to inside the if statement above? And in the else, just fail the build?

Instead of setting another variable above, and essentially checking the same if statement here that we already checked above?

.github/workflows/pull-request.yml
```diff
@@ -13,12 +13,29 @@ on:
...
+      # image owns as `pwuser`. Firefox refuses to launch when $HOME is owned by a
+      # different user, so point HOME at root's own home.
+      HOME: /root
 
     steps:
       - uses: actions/checkout@v5
         with:
           fetch-depth: 0
 
+      - name: Prepare git base ref
```
ArakTaiRoth: I don't understand why this additional step is necessary? Above, we already checkout everything on line 29 - 31, so this shouldn't be explicitly necessary. Could we remove it and see if it still works as expected?

ArakTaiRoth: Actually, now that I look at it more, we might need the line on line 35, but the `git fetch` line shouldn't be necessary at all.

.github/workflows/pull-request.yml
vanessatran-ddi: This is an intentionally tradeoff. I pinned the image manually instead of deriving it automatically so the CI browser version stays an explicit (same reason the silent `ubuntu-latest` just bit us). If someone bumps Playwright and forgets the tag, it only fails that bump PR.

.github/workflows/pull-request.yml
vanessatran-ddi: I combined it into the Test step to match how dev did it before.

.github/workflows/pull-request.yml
vanessatran-ddi: You are right. the `git fetch` was redundant. I removed it

ArakTaiRoth (approved):

willcodeforcoffee (approved):
LGTM! Thanks so much, Vanessa!

PR #4000, opened 2026-05-29
feat(#3346): add horizontal scrolling to Scroll Panel

bdfranck:
This PR adds a `direction` property to the Scroll Panel to allow for horizontal scrolling. This PR was branched off of https://github.com/GovAlta/ui-components/pull/3933 and will need to be updated as that work evolves.

scrolling


twjeffery (commented):
@bdfranck Maybe I'm not pulling in the right branch, but I think this needs a rebase onto the current #3933 before we can review it visually. It looks like it branched off an earlier point of that branch, before Vanessa's updates on the scroll panel's shadow styles.

Image

PR #4002, opened 2026-05-29
fix(#4001): scope PreviewLayout's full-viewport reset to its own pages

### LINKED ISSUE #4001: Page background cuts off below the fold on long pages
On the documentation site, the page background only covers the first screen. On any page taller than the window, scrolling down reveals a white strip where the background should continue. It happens in both light and dark mode, and is most obvious in dark mode where the dark background gives way to white.

**Where:** Get Started, Components, Examples, and Foundations pages (including their landing pages). The home page is not affected.

**Steps to reproduce:**
1. Open a docs page with enough content to scroll (for example, a Get Started page).
2. Scroll down past one screen height.
3. The grey (or dark) page background stops and the area below it turns white.

**Expected:** The background extends the full height of the page, regardless of content length.

**Likely cause:** A global style meant only for the example preview pages is applying site-wide. It pins `html, body` to `height: 100%`, locking the body to one screen so the background can't grow with the content. A PR to scope that style to the preview pages only will follow.

Image

twjeffery:
## What was happening
On the docs site, the page background only covered the first screen. On any page taller than the window, scrolling down revealed a white strip where the background should continue. It showed on Get Started, Components, Examples, and Foundations pages, in both light and dark mode, and was most obvious in dark mode. The home page was unaffected because it fits in one screen.

## Why
PreviewLayout (the shell for full page example previews) carried a global reset that set `html, body { height: 100% }` and made the body a flex column. Because it was global, it applied to every page, not just the preview routes. `height: 100%` locks the body to exactly one screen, so on longer pages the body could not grow to fit its content. The page background sits on the body, so it stopped at the fold and the white underneath showed through.

## The fix
Scope that reset to the preview pages. PreviewLayout now adds a `preview-doc` class to its own `body`, and the reset targets `body.preview-doc` instead of every `html`/`body`. Preview pages render exactly as before. Every other page keeps a body that can grow with its content, so the background fills the full height.

The `goa-*` layout rule in the same block stays global on purpose: inline examples on guidance and docs pages render those layout components outside this layout and depend on it.

## Also in this PR
Set the docs `html` background to the same grey token the page wrapper uses, so the scroll overflow area (the elastic overscroll space past the top and bottom of the page on macOS and iOS) matches the page instead of showing white. It uses the same `--goa-color-greyscale-50` token as `.layout-wrapper`, so it tracks light, dark, and forced dark automatically. Below 624px, where the layouts switch the wrapper to white, the overscroll matches that too.

## Verified
- Doc pages: the body and grey background now grow to the full document height instead of stopping at one screen.
- Preview pages: unchanged. Full-height shell and layout components still render correctly (checked the 404 preview).

Closes #4001.

### DIFF
--- a/docs/src/layouts/PreviewLayout.astro
+++ b/docs/src/layouts/PreviewLayout.astro
@@ -48,7 +48,7 @@ const resolvedBackUrl = withBase(backUrl);
     <script type="module" src="https://unpkg.com/ionicons@5.5.2/dist/ionicons/ionicons.esm.js"></script>
     <script nomodule src="https://unpkg.com/ionicons@5.5.2/dist/ionicons/ionicons.js"></script>
   </head>
-  <body>
+  <body class="preview-doc">
     <header class="preview-bar">
       <goa-link size="small" leadingicon="arrow-back">
         <a href={resolvedBackUrl} data-back-link>
@@ -103,14 +103,11 @@ const resolvedBackUrl = withBase(backUrl);
 </html>
 
 <style is:global>
-  html,
-  body {
+  /* Scoped to .preview-doc (set on this layout's own body) so this
+     full-viewport reset cannot leak onto pages that use BaseLayout. */
+  body.preview-doc {
     margin: 0;
     padding: 0;
-    height: 100%;
-  }
-
-  body {
     font-family: var(--goa-font-family-sans, -apple-system, BlinkMacSystemFont, "Segoe UI", system-ui, sans-serif);
     color: var(--goa-color-text-default, #333);
     background: var(--goa-color-surface-default, #fff);
@@ -145,7 +142,9 @@ const resolvedBackUrl = withBase(backUrl);
 
   /* GoA layout web components default to display:inline because their
      :host rules don't set display:block. Force them to block layout so
-     they fill their parent container in preview routes. */
+     they fill their parent container. Intentionally global: inline
+     <Preview> examples on guidance/docs pages render these components
+     outside this layout and rely on it too. */
   goa-one-column-layout,
   goa-page-block,
   goa-app-header,
--- a/docs/src/layouts/BaseLayout.astro
+++ b/docs/src/layouts/BaseLayout.astro
@@ -107,10 +107,22 @@ const ogUrl = new URL(withBase(Astro.url.pathname), Astro.site).href;
 
       html {
         color-scheme: light;
+        /* Match the page's grey wrapper so the scroll overflow / overscroll
+           area (visible on elastic scroll) blends instead of showing white.
+           Same token as .layout-wrapper, so it matches in light/dark/forced-dark. */
+        background: var(--goa-color-greyscale-50);
       }
 
       html[data-theme="dark"] {
         color-scheme: dark;
+      }
+
+      /* Below 624px the layouts flip the page wrapper to white, so match the
+         overscroll area there too instead of leaving it grey. */
+      @media (max-width: 623px) {
+        html {
+          background: var(--goa-color-greyscale-white);
+        }
       }
 
       body {


docs/src/layouts/PreviewLayout.astro
```diff
@@ -37,7 +37,7 @@ const resolvedBackUrl = withBase(backUrl);
 ---
 
 <!DOCTYPE html>
-<html lang="en">
+<html lang="en" class="preview-doc">
```
ArakTaiRoth: You don't need an HTML class for this

docs/src/layouts/PreviewLayout.astro
```diff
@@ -103,14 +103,16 @@ const resolvedBackUrl = withBase(backUrl);
 </html>
 
 <style is:global>
-  html,
-  body {
+  /* Scoped to .preview-doc (set on this layout's own html/body) so this
+     full-viewport reset cannot leak onto pages that use BaseLayout. */
+  html.preview-doc,
```
ArakTaiRoth: If you don't have the HTML class, then this can be removed

docs/src/layouts/PreviewLayout.astro
```diff
@@ -103,14 +103,16 @@ const resolvedBackUrl = withBase(backUrl);
...
+  body.preview-doc {
     margin: 0;
     padding: 0;
     height: 100%;
   }
 
-  body {
+  body.preview-doc {
     font-family: var(--goa-font-family-sans, -apple-system, BlinkMacSystemFont, "Segoe UI", system-ui, sans-serif);
```
ArakTaiRoth, line 111: All of this CSS can be moved to be under the `body.preview-doc` above on line 109 currently

docs/src/layouts/PreviewLayout.astro
```diff
@@ -103,14 +103,16 @@ const resolvedBackUrl = withBase(backUrl);
...
 <style is:global>
-  html,
-  body {
+  /* Scoped to .preview-doc (set on this layout's own html/body) so this
+     full-viewport reset cannot leak onto pages that use BaseLayout. */
+  html.preview-doc,
+  body.preview-doc {
     margin: 0;
     padding: 0;
     height: 100%;
```
ArakTaiRoth: If you do the above, then this line isn't necessary either because we're setting min-height to 100vh

docs/src/layouts/PreviewLayout.astro
twjeffery: Removed. It's only on `<body>` now.

docs/src/layouts/PreviewLayout.astro
twjeffery: Removed. The rule is just `body.preview-doc` now.

docs/src/layouts/PreviewLayout.astro
twjeffery: Merged into the single `body.preview-doc` rule.

docs/src/layouts/PreviewLayout.astro
twjeffery: Dropped. `min-height: 100vh` covers it, and I confirmed the preview shells still render full-height without it.

docs/src/layouts/PreviewLayout.astro
```diff
@@ -145,7 +142,9 @@ const resolvedBackUrl = withBase(backUrl);
 
   /* GoA layout web components default to display:inline because their
      :host rules don't set display:block. Force them to block layout so
-     they fill their parent container in preview routes. */
+     they fill their parent container. Intentionally global: inline
+     <Preview> examples on guidance/docs pages render these components
+     outside this layout and rely on it too. */
```
vanessatran-ddi, line 147: Should we move this style for `goa-*` components into `<style is:global>` is inside `BaseLayout.astro` so all pages will inherit this style

vanessatran-ddi (approved):
Looks good!

bdfranck (approved):
- ✅ I see the base desktop layout bg colour is `greyscale-50`
- ✅ I see the base mobile layout bg colour is `greyscale-white`
- ✅ I only see the preview styling changes on the preview template

Image

Looks good to me! 👍

PR #4006, opened 2026-06-01
chore(#3943): release docs from main and add dev branch preview

### LINKED ISSUE #3943: Add Dev preview link to PR comments and stop deploying docs from dev branch
## Description

One changes to the docs deployment workflow:

- The docs site is currently deployed when code merges to `dev`. This should be moved to trigger on `main` instead, so the published docs site reflects released code rather than in-progress dev work.

One addition to our #design-system-team slack channel:

- A "Dev" link should be added pointing to the dev branch deployment so reviewers can see changes in that environment (particularly the docs site).

## Acceptance Criteria

- [ ] A "Dev" link is included in our #design-system-team channel
- [ ] The docs site is no longer deployed when merging to `dev`
- [ ] The docs site deploys when merging to `main`

vanessatran-ddi:
# Before (the change)

# After (the change)

### DIFF
--- a/.github/workflows/dev-release.yml
+++ b/.github/workflows/dev-release.yml
@@ -1,0 +1,126 @@
+name: Dev Release & Preview
+
+on:
+  push:
+    branches: [dev]
+  workflow_dispatch:
+
+concurrency:
+  group: dev-release
+  cancel-in-progress: false
+
+permissions:
+  contents: write
+  id-token: write # Required for OIDC
+
+jobs:
+  dev-release:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
+
+      - uses: actions/setup-node@v6
+        with:
+          node-version: "24"
+          cache: "npm"
+          cache-dependency-path: |
+            package-lock.json
+            docs/package-lock.json
+
+      - name: Install npm 11
+        run: npm install -g npm@11
+
+      - name: Install dependencies
+        run: |
+          npm ci
+          cd docs && npm ci --legacy-peer-deps
+
+      # ---- Release libs to npm (dev prerelease channel) ----
+      - name: Get last successful commit
+        uses: nrwl/nx-set-shas@v4
+        id: last_successful_commit
+        with:
+          main-branch-name: ${{ github.ref_name }}
+
+      - name: Build libs
+        run: npm run build:prod
+
+      - name: Fix web-component lib README
+        run: cp libs/web-components/README.md dist/libs/web-components
+
+      - name: Release libs
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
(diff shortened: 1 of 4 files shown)


ArakTaiRoth:
@vanessatran-ddi Just a suggestion here, right now both .yml files run dev, and both are duplicating effort (doing some of the same thing), which means publishing to dev will take longer than it needs to. I would suggest instead making a dev only release file that does everything you have in the new file, and adds in the publishing to npm into it as well from the other file. Then the other file, dev gets removed from (as well as the other branches, we only need main), and it handles exactly what it's handling now. Bonus, you no longer need to check the if statement about what branch is running it, because it will only be run in main.

willcodeforcoffee (approved):

vanessatran-ddi:
> @vanessatran-ddi Just a suggestion here, right now both .yml files run dev, and both are duplicating effort (doing some of the same thing), which means publishing to dev will take longer than it needs to. I would suggest instead making a dev only release file that does everything you have in the new file, and adds in the publishing to npm into it as well from the other file. Then the other file, dev gets removed from (as well as the other branches, we only need main), and it handles exactly what it's handling now. Bonus, you no longer need to check the if statement about what branch is running it, because it will only be run in main.

Thank you @ArakTaiRoth, that is a good call on the duplication and the branch `if` check. I did an investigation and the 2 workflows actually build different targets. `release-ci` builds the libs, and `dev-preview` buids docs plus the playgrounds to deploy to Pages. So the real overlap on a dev push is mostly `npm ci` and `release-ci`.  The win from consolidating is real but smaller than it looks. `next` is still configured as a prerelease channel in `.releaserc.json` alongside `dev`. Dropping `next` can be done in another PR (out of scope. 
So I would like to land it as is and open a follow up to consolidate the dev release plus docs preview into 1 workflow and remove the `next`. Does it sound good to you?

.github/workflows/dev-release.yml
```diff
@@ -0,0 +1,126 @@
...
+        run: cp libs/web-components/README.md dist/libs/web-components
+
+      - name: Release libs
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: npx nx affected --target release --parallel=false --base=${{ steps.last_successful_commit.outputs.base }}
+
+      # ---- Build + deploy docs & playground previews to Pages ----
+      - name: Build docs
```
ArakTaiRoth, line 61: Correct me if I'm wrong. But I don't think this step is needed actually. On line 49 we run `npm run build:prod` which as I understand it, will also build docs.

vanessatran-ddi: Line 49 will build docs, but build with the default base `/`. Dev preview deploy will need to be inside a sub folder `ui-components/dev/` so we need to build again with PREVIEW_BASE at line 61. So honestly the `build docs` on step `build-prod` is a bit redundant, but it can be a good sanity check before release libs and make sure everything compile and build correctly, so we can keep like the way it is now.

ArakTaiRoth (approved):

PR #4010, opened 2026-06-03
fix(#4009): align checkbox and radio item label weight and color

### LINKED ISSUE #4009: Checkbox and radio item labels have inconsistent weight and color
Spotted while reviewing #3992 (vertical alignment). This is a separate typography consistency issue, not alignment.

## Current behavior

The checkbox label and the radio item label render differently even though they are the same kind of form-control label, side by side:

- Checkbox label is **regular weight (400)** but the correct color (it sets the `--goa-checkbox-color-label` component token).
- Radio item label is **medium weight (500)** but sets **no color at all**, so it inherits the host default (near-black) instead of the secondary text color the checkbox uses.

So each is wrong on a different axis: the checkbox has the right color but the wrong weight, and the radio has the right weight but the wrong color.

## Expected behavior

Checkbox and radio item labels should match, and both should get their color the same way: each through its own component token that points at the shared secondary input text color, not a hardcoded value.

- Default size: 18px / 24px line height, medium weight
- Compact size: keeps its smaller font size, same medium weight
- Color: both resolve to `--goa-input-color-text-secondary` through their own component tokens (`--goa-checkbox-color-label` and a new `--goa-radio-color-label`)

## Visual or code evidence

[Attach the checkbox vs radio side-by-side screenshots.]

Checkbox (`libs/web-components/src/components/checkbox/Checkbox.svelte`, `.text`):
- `font: var(--goa-checkbox-label-font-size)` which resolves to `--goa-typography-body-m` (regular)
- `color: var(--goa-checkbox-color-label)` (= #353535)

Radio item (`libs/web-components/src/components/radio-item/RadioItem.svelte`, `.label`):
- `font: var(--goa-radio-label)` (medium)
- no `color` set, so it inherits `--goa-color-text-default` (greyscale-black)

## Suggested fix or root cause

The two labels were authored independently: the radio never set a label color token (so it inherits the page default), and the checkbox label font token points at the regular body token rather than a medium one.

Color (through tokens, no hardcoded hex):
- The checkbox already does this correctly: `color: var(--goa-checkbox-color-label)`, where that component token maps to `--goa-input-color-text-secondary`. No color change needed on the checkbox.
- The radio has no equivalent. Add a component token `--goa-radio-color-label`, defined the same way as `--goa-checkbox-color-label` (light: `var(--goa-input-color-text-secondary)`; dark theme matching the checkbox), and use `color: var(--goa-radio-color-label)` in the radio CSS. Both labels then point at the same shared token and stay matched across themes.

Weight:
- Checkbox: set `font-weight: var(--goa-font-weight-medium)` on both the base label and the compact label (the compact `font` shorthand resets the weight, so it needs it too). The radio is already medium.

## Target values

- Size: 18px, default (`--goa-font-size-4`, 1.125rem)
- Line height: 24px (`--goa-line-height-3`, 1.5rem)
- Weight: medium, 500 (`--goa-font-weight-medium`)
- Color: through component tokens (`--goa-checkbox-color-label`, new `--goa-radio-color-label`) that reference `--goa-input-color-text-secondary`, which resolves to `--goa-color-greyscale-800` (#353535) in light and tracks the theme in dark. Don't hardcode the hex.

## Open questions

None blocking for checkbox + radio. Possible follow-up as a separate issue: extend the same label spec to the other form-control labels (toggle, form-item) so all input labels share one definition.

twjeffery:
# Before (the change)

The checkbox label and the radio item label rendered differently even though they are the same kind of form control label, side by side:

- Checkbox label: correct color (`--goa-checkbox-color-label`) but regular weight (400). The `font` shorthand from `--goa-checkbox-label-font-size` resolves to the regular body token, so the label was regular.
- Radio item label: correct medium weight (500) but no color set, so it inherited the host default (near black) instead of the secondary input text color the checkbox uses.

So each was wrong on a different axis.

# After (the change)

Both labels now render in medium weight (500) and resolve their color the same way, through their own component token pointing at the shared secondary input text color.

- Checkbox (`Checkbox.svelte`, `.text` and `.v2.compact .text`): added `font-weight: var(--goa-font-weight-medium)`. The `font` shorthand resets weight to the token default, so it has to be restated, including on the compact label whose shorthand also resets it.
- Radio item (`RadioItem.svelte`, `.label`): added `color: var(--goa-radio-color-label, var(--goa-input-color-text-secondary))`. This introduces a new `--goa-radio-color-label` component token (to be defined the same way as `--goa-checkbox-color-label` in the tokens package) and falls back to the shared `--goa-input-color-text-secondary` so the color resolves correctly today, matching the repo's V2 fallback convention. No color change was needed on the checkbox; it already resolves correctly.

This is a CSS only change in the Svelte source, so the React and Angular wrappers did not need updates (wrappers pass attributes through and own no styling).

## Steps needed to test

1. Run the React playground: `npm run serve:prs:react` and open the `4009 Checkbox and radio label consistency` page from the side menu.
2. In Test 1, confirm the checkbox label and the radio item label match: both medium weight and both the secondary input text color (#353535 in light theme).
3. In Test 2, confirm the compact checkbox label keeps its smaller font size but is still medium weight.
4. In Test 3, confirm disabled labels still use their disabled color and are unaffected.
5. Repeat in the Angular playground: `npm run serve:prs:angular`, same `4009` page.

## Tests

- Extended `checkbox.browser.spec.tsx`: asserts the base and compact labels consume the medium font-weight token.
- Extended `radio.browser.spec.tsx`: asserts the item label resolves its color through the shared secondary input text token via the new component token's fallback.
- Verified both new tests fail without the source change and pass with it (Chromium and Firefox).

Fixes #4009

---
_Generated by [Claude Code](https://claude.ai/code/session_01X1pkKbVgv5UTAn3am3dPSA)_

### DIFF
--- a/libs/web-components/src/components/radio-item/RadioItem.svelte
+++ b/libs/web-components/src/components/radio-item/RadioItem.svelte
@@ -343,6 +343,7 @@
 
   .label {
     font: var(--goa-radio-label);
+    color: var(--goa-radio-color-label, var(--goa-input-color-text-secondary));
     margin-top: -3px; /* V1: Optical centering - move text up */
     padding-left: var(--goa-radio-gap-label, var(--goa-space-s));
   }
--- a/apps/prs/angular/src/routes/bugs/4009/bug4009.component.html
+++ b/apps/prs/angular/src/routes/bugs/4009/bug4009.component.html
@@ -1,0 +1,58 @@
+<goab-block gap="m" direction="column">
+  <goab-text tag="h1" mt="m">
+    Bug #4009: Checkbox and radio item labels have inconsistent weight and color
+  </goab-text>
+
+  <goab-text tag="p">
+    The checkbox label and the radio item label rendered differently even though
+    they are the same kind of form control label. The checkbox had the right color
+    but regular weight, and the radio had the right medium weight but no color, so it
+    inherited the near black page default. After the fix both render in medium weight
+    and resolve their color through component tokens that point at the shared
+    secondary input text color.
+  </goab-text>
+
+  <goab-divider mt="l" mb="l"></goab-divider>
+
+  <goab-text tag="h2">Test Cases</goab-text>
+
+  <goab-text tag="h3">Test 1: Default size, side by side</goab-text>
+  <goab-text tag="p">
+    The checkbox and radio item labels should be identical in weight (medium) and
+    color (secondary input text color).
+  </goab-text>
+  <goab-block gap="m" direction="column" mb="l">
+    <goab-checkbox name="default-checkbox" text="Checkbox label"></goab-checkbox>
+    <goab-radio-group name="default-radio">
+      <goab-radio-item value="option1" label="Radio item label"></goab-radio-item>
+    </goab-radio-group>
+  </goab-block>
+
+  <goab-text tag="h3">Test 2: Compact size</goab-text>
+  <goab-text tag="p">
+    The compact checkbox label keeps its smaller font size but stays medium weight.
+  </goab-text>
+  <goab-block gap="m" direction="column" mb="l">
+    <goab-checkbox
+      name="compact-checkbox"
+      size="compact"
+      text="Compact checkbox label"
+    ></goab-checkbox>
+  </goab-block>
+
+  <goab-text tag="h3">Test 3: Disabled state</goab-text>
+  <goab-text tag="p">
+    Disabled labels use their own disabled color token and are unaffected by this
+    change.
+  </goab-text>
+  <goab-block gap="m" direction="column">
+    <goab-checkbox
+      name="disabled-checkbox"
+      text="Disabled checkbox label"
+      [disabled]="true"
+    ></goab-checkbox>
+    <goab-radio-group name="disabled-radio" [disabled]="true">
+      <goab-radio-item value="option1" label="Disabled radio item label"></goab-radio-item>
+    </goab-radio-group>
+  </goab-block>
+</goab-block>
--- a/apps/prs/angular/src/routes/bugs/4009/bug4009.component.ts
+++ b/apps/prs/angular/src/routes/bugs/4009/bug4009.component.ts
@@ -1,0 +1,24 @@
+import { Component } from "@angular/core";
+import {
+  GoabBlock,
+  GoabCheckbox,
+  GoabDivider,
+  GoabRadioGroup,
+  GoabRadioItem,
+  GoabText,
+} from "@abgov/angular-components";
+
+@Component({
+  standalone: true,
+  selector: "abgov-bug4009",
+  templateUrl: "./bug4009.component.html",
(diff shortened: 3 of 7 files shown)


libs/react-components/specs/checkbox.browser.spec.tsx
```diff
@@ -134,6 +134,49 @@ describe("Checkbox", () => {
     });
   });
 
+  it("applies the medium font-weight token to the label", async () => {
```
bdfranck: Are we sure we want to add these browser tests? We normally use these tests for regressions in functionality, not styling. I don't know how I'd feel if someone's PR was blocked because of a font weight.

bdfranck (commented):
I reviewed the changes...

- ✅ I see the checkbox label is the correct font weight

Image
 
- ✅ I see the radio item label is the correct colour

Image


Looks good to me! 👍  I have one comment about the tests.

libs/react-components/specs/checkbox.browser.spec.tsx
twjeffery: Good call. I've removed the browser style assertions from both `checkbox.browser.spec.tsx` and `radio.browser.spec.tsx` so this fix won't gate PRs on a computed font weight or color. The change is verified visually via the `4009` playground pages in React and Angular instead. (The web-component unit tests for both components still pass.)

---
_Generated by [Claude Code](https://claude.ai/code)_

bdfranck (approved):
I looked at the latest changes and see the browser tests are gone. Looks good to me! 👍

libs/web-components/src/components/radio-item/RadioItem.svelte
```diff
@@ -343,6 +343,9 @@
 
   .label {
     font: var(--goa-radio-label);
+    /* Match the checkbox label: resolve color through a component token that
+       points at the shared secondary input text color. */
+    color: var(--goa-radio-color-label, var(--goa-input-color-text-secondary));
```
vanessatran-ddi, line 346: We can remove the comment line 346. We need a PR on design-tokens repo to add this `--goa-radio-color-label`

twjeffery: Done, removed the comment. The `color: var(--goa-radio-color-label, var(--goa-input-color-text-secondary))` line stays, so it falls back to the secondary input text color until the `--goa-radio-color-label` token lands in the design-tokens repo. I'll leave that separate tokens PR to the team since it's outside this repo.

---
_Generated by [Claude Code](https://claude.ai/code)_

twjeffery: this has been done, please re-review

twjeffery: @vanessatran-ddi

vanessatran-ddi (approved):

PR #4021, opened 2026-06-04
fix(#4020): show the right properties on the Temporary Notification page

### LINKED ISSUE #4020: Temporary Notification: Web Components props show the controller, not the notification
## Current behavior

The Temporary Notification page's Web Components Properties tab shows the controller's positioning props (`horizontal-position`, `vertical-position`, `testid`) instead of the notification's API. Regressed with the extraction change in #3770.

## Expected behavior

The Web Components tab shows the notification props: `message`, `type`, `progress`, `action-text`, `animation-direction`, `visible`, `testid`.

## Cause and fix

`temporary-notification/` has two custom elements: the notification (`goa-temp-notification`, no React wrapper) and the controller (`goa-temp-notification-ctrl`, which has one). Since #3770, `extract-api` resolves the element through the React wrapper, so it lands on the controller. Fix: prefer the Svelte file matching the slug. Scanned all 69 component JSONs; this is the only one affected.

Image

twjeffery:
Closes #4020.

The Temporary Notification page documented the wrong thing. `temporary-notification/` has two custom elements: the notification (`goa-temp-notification`) and the controller (`goa-temp-notification-ctrl`). Only the controller has React and Angular wrappers, so it's the only element that exists across all three frameworks. And the notification's content (message, type, duration, and so on) isn't set through element props at all. It's set imperatively through `TemporaryNotification.show(message, options)`, which a props table can't express.

This PR does two things:

1. **Keeps all three element tabs on the controller** so React, Angular, and Web Components stay consistent. (An earlier attempt surfaced the notification element's props on the Web Components tab, which made it diverge from the other two. That's been reverted.)

2. **Documents the imperative API** in a new section under Properties: `show(message, options)`, `dismiss(uuid)`, and `setProgress(uuid, progress)`. `extract-api` now parses the controller source and emits a `staticMethods` block, rendered as a signature plus a params table per method. The method list, signatures, and descriptions all come from the controller itself (JSDoc on the functions and on `GoabNotificationOptions`, with `uuid` and `visible` tagged `@internal` to keep them out of the docs). `show()`'s options table covers `type`, `duration`, `actionText`, `action`, and `cancelUUID`.

Regenerating all 69 component JSONs confirms `temporary-notification` is the only one affected. The docs build passes and the new section renders on the preview.

### DIFF
--- a/docs/generated/component-apis/temporary-notification.json
+++ b/docs/generated/component-apis/temporary-notification.json
@@ -92,5 +92,92 @@
       "events": [],
       "slots": []
     }
-  }
+  },
+  "staticMethods": [
+    {
+      "name": "show",
+      "signature": "TemporaryNotification.show(message, options)",
+      "returnType": "string",
+      "description": "Displays a temporary notification from your component. Returns the notification's UUID, which you can use to dismiss it or update its progress.",
+      "params": [
+        {
+          "name": "message",
+          "type": "string",
+          "required": true,
+          "description": "The message to display in the notification."
+        },
+        {
+          "name": "options.type",
+          "type": "GoabTemporaryNotificationType",
+          "values": [
+            "basic",
+            "success",
+            "failure",
+            "indeterminate",
+            "progress"
+          ],
+          "required": false,
+          "description": "The type of notification, which determines its styling and icon. Use \"indeterminate\" to show an animated progress bar while work of unknown length runs, or \"progress\" to show a progress bar you update with setProgress(). Defaults to \"basic\"."
+        },
+        {
+          "name": "options.duration",
+          "type": "\"long\" | \"medium\" | \"short\" | number",
+          "required": false,
+          "description": "How long the notification stays before it auto-dismisses: \"short\" (about 3 seconds), \"medium\" (about 4 seconds), \"long\" (about 6 seconds), or a number of milliseconds. Only \"basic\", \"success\", and \"failure\" notifications auto-dismiss (default \"short\"). \"indeterminate\" and \"progress\" notifications have no default duration and stay until you dismiss them."
+        },
+        {
+          "name": "options.actionText",
+          "type": "string",
+          "required": false,
+          "description": "Text for an action button. When set, the notification shows a button the user can select."
+        },
+        {
+          "name": "options.action",
+          "type": "() => void",
+          "required": false,
+          "description": "Function to run when the action button is selected."
+        },
+        {
+          "name": "options.cancelUUID",
+          "type": "string",
+          "required": false,
+          "description": "UUID of an existing notification to cancel when this one is shown."
+        }
+      ]
+    },
+    {
+      "name": "dismiss",
+      "signature": "TemporaryNotification.dismiss(uuid)",
+      "returnType": "void",
+      "description": "Hides a notification, using the UUID that show() returns.",
+      "params": [
+        {
+          "name": "uuid",
+          "type": "string",
+          "required": true,
+          "description": "The UUID of the notification to dismiss. This is the value that show() returns."
+        }
+      ]
+    },
+    {
+      "name": "setProgress",
+      "signature": "TemporaryNotification.setProgress(uuid, progress)",
+      "returnType": "void",
+      "description": "Updates the progress shown on a progress notification, using the UUID that show() returns.",
+      "params": [
+        {
+          "name": "uuid",
+          "type": "string",
+          "required": true,
+          "description": "The UUID of the progress notification to update. This is the value that show() returns."
+        },
+        {
+          "name": "progress",
+          "type": "number",
+          "required": true,
+          "description": "The progress to display, from 0 to 100."
+        }
+      ]
+    }
+  ]
 }
--- a/docs/src/components/StaticMethods.astro
+++ b/docs/src/components/StaticMethods.astro
@@ -1,0 +1,154 @@
+---
+/**
+ * StaticMethods - Documents imperative helper methods (e.g. TemporaryNotification.show)
+ * that are exposed as a static API rather than element props.
+ *
+ * Framework-agnostic: a single signature snippet plus a params table, rendered the
+ * same way regardless of the selected framework. Visual language matches PropsTable.
+ */
+import type { StaticMethod } from "../lib/content-queries";
+
+interface Props {
+  methods: StaticMethod[];
+}
+
+const { methods } = Astro.props;
+---
+
+{
+  methods.map((method) => (
+    <div class="api-section">
+      <div class="api-heading">
+        <h3 id={`method-${method.name}`}>{method.name}()</h3>
+      </div>
+
+      {method.description && <p class="static-method-description">{method.description}</p>}
+
+      <pre class="static-method-signature"><code>{method.signature}{method.returnType ? `: ${method.returnType}` : ""}</code></pre>
+
+      <goa-container version="2" type="interactive">
+        <goa-data-grid keyboard-nav="layout" keyboard-icon-position="right">
+          {method.params.map((param) => (
+            <div class="prop-row" data-grid="row">
+              <div class="prop-cell prop-cell-name" data-grid="cell">
+                <span class="prop-name">{param.name}</span>
+                {param.required && (
+                  <goa-badge version="2" type="important" icon="false" content="Required" />
+                )}
+              </div>
+              <div class="prop-cell prop-cell-type" data-grid="cell">
+                <span class="prop-type">{param.type}</span>
+                {param.values && param.values.length > 0 && (
+                  <span class="prop-type-details">{param.values.join(" | ")}</span>
+                )}
+              </div>
+              {param.description && (
+                <div class="prop-cell prop-cell-description" data-grid="cell">
+                  {param.description}
+                </div>
+              )}
+            </div>
+          ))}
+        </goa-data-grid>
+      </goa-container>
+    </div>
+  ))
+}
+
+<style>
+  .api-section {
+    margin-bottom: var(--goa-space-xl);
+  }
+
+  .api-heading {
+    display: flex;
+    align-items: center;
+    gap: var(--goa-space-m);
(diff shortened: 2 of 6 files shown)


vanessatran-ddi:
React and Angular only have a wrapper for the controller (`temporary-notification-ctrl`) not for the inner `goa-temp-notification`. So the only element that exists across all 3 frameworks is the controller. This PR makes the web components tab show the inner notification's props while React and Angular shows the controller's props, so the 3 tabs no longer match. 

image

The root issue is that this component is a controller plus an `.show()` API, and a generic props table cannot express that. My suggestion:
1. Narrow this PR: restore all three tabs to the controller so they are consistent again
2. Change this to adding a section documenting `TemporaryNotification.show(message, options)` with a prams table sourced from `GoabNotification` interfave and extend `extract-api`. Reference v1 docs https://v1.design.alberta.ca/components/temporary-notification#components-temporarynotificationctrl-properties

What do you think? We could fold all of it into this PR.

twjeffery:
@vanessatran-ddi I Went with both suggestions and folded them into this PR.

1. Reverted that first change so all three tabs show the controller again and stay consistent.
2. Added an imperative API section under Properties. `extract-api` now reads `GoabNotificationOptions` from the controller source to build the params table for `show(message, options)`, with descriptions from the v1 docs.

I documented all three helpers (`show`, `dismiss`, `setProgress`) since they work as a set: `show` hands back a UUID that `dismiss` and `setProgress` take. For `show()`'s options I kept the table to the ones you'd actually pass (`type`, `duration`, `actionText`, `action`, `cancelUUID`) and left out `uuid` and `visible` since `show()` manages those itself.

docs/generated/component-apis/temporary-notification.json
```diff
@@ -92,5 +92,93 @@
...
+        },
+        {
+          "name": "options.type",
+          "type": "GoabTemporaryNotificationType",
+          "typeLabel": "GoabTemporaryNotificationType",
+          "values": [
+            "basic",
+            "success",
+            "failure",
+            "indeterminate",
```
vanessatran-ddi, line 116: I think we need to add description to explain what is `indeterminate` and what is `progress`, They are quite simila, also for `indeterminate` and `progess` it is not autodismiss, which may need to have more clarification under line 127

docs/src/components/StaticMethods.astro
```diff
@@ -0,0 +1,154 @@
...
+    font: var(--goa-typography-number-s);
+    color: var(--goa-color-text-default);
+  }
+
+  .prop-row {
+    display: flex;
+    flex-wrap: wrap;
+    align-items: baseline;
+    gap: 0;
+    border-bottom: 1px solid var(--goa-color-greyscale-200);
```
vanessatran-ddi: picky: can we use `var(--goa-border-width-s)`

docs/src/lib/content-queries.ts
```diff
@@ -45,11 +45,29 @@ export interface SubComponentApi {
   frameworks: FrameworkApiMap;
 }
 
+export interface StaticMethodParam {
+  name: string;
+  type: string;
+  typeLabel?: string;
```
vanessatran-ddi: `type` and `typeLabel` are set to have the same value here (`rawType`) so `typeLabel` is redundant in this path. normally we will have `type` as an interface for example `GoabTemporaryNotificationType` and `values` carrying the expansion, so I think we can remove the `typeLabel`

docs/src/lib/content-queries.ts
```diff
@@ -45,11 +45,29 @@ export interface SubComponentApi {
   frameworks: FrameworkApiMap;
 }
 
+export interface StaticMethodParam {
```
vanessatran-ddi: `StaticMethodParam` is similar to `PropDefinition` line 4 above, instead of duplicating them, how about `export type StaticMethodParam = Omit<PropDefinition, \"default\" | \"deprecated\";`

docs/src/scripts/extract-api.ts
```diff
@@ -2076,6 +2094,218 @@ function findSvelteDeclaringTag(tag: string): string | undefined {
   return undefined;
 }
 
+// =============================================================================
+// Static helper methods (imperative APIs that are not element props)
+// =============================================================================
+
+// A few components expose an imperative helper (e.g. TemporaryNotification.show)
+// rather than, or alongside, element props. The param *structure* (names, types,
```
vanessatran-ddi: update this js doc comment to be `option fields are parsed from source so they cannot drift; leading params and return types are documented here and not verified against the function signature`

docs/src/scripts/extract-api.ts
```diff
@@ -2076,6 +2094,218 @@ function findSvelteDeclaringTag(tag: string): string | undefined {
...
+        console.warn(
+          `  static method ${method.name}: option "${option.field}" not found on ${config.optionsType}`,
+        );
+        continue;
+      }
+      const rawType = cleanType(member.type?.getText(sourceFile)?.trim() || "unknown");
+      const values = parseTypeAliasUnionValues(rawType, sourceFile);
+      optionParams.push({
+        name: `${method.optionsParamName}.${option.field}`,
+        type: rawType,
```
vanessatran-ddi, line 2326: We can remove `typeLabel` under interface `ExtractedStaticMethodParam: typeLabel?` according to the logic written here.

docs/src/scripts/extract-api.ts
```diff
@@ -2076,6 +2094,218 @@ function findSvelteDeclaringTag(tag: string): string | undefined {
...
+  options?: StaticMethodOptionDoc[]; // public option fields, in display order
+}
+
+interface StaticMethodSourceDoc {
+  optionsSource: string; // path to the options type's source file, relative to the workspace root
+  optionsType: string; // the type alias to read option field structure from
+  methods: StaticMethodDoc[];
+}
+
+const STATIC_METHOD_DOCS: Record<string, StaticMethodSourceDoc> = {
```
vanessatran-ddi: Is there a way to extract directly from common `temporary-notification-controller.ts`? I think we can add JS Docs per methods there and then add a small logic on this script to generate this static method docs instead of hardcoding them here right?

docs/generated/component-apis/temporary-notification.json
twjeffery: Added descriptions for both. The type description now explains what indeterminate and progress are for, and the duration description says only basic, success, and failure auto-dismiss while indeterminate and progress have no default duration and stay until dismissed. Since I went with your JSDoc suggestion, these now live on the options type in the controller.

docs/src/components/StaticMethods.astro
twjeffery: Swapped to `var(--goa-border-width-s)`. PropsTable has the same hardcoded 1px from before this PR, I left that one alone to keep this PR small.

docs/src/lib/content-queries.ts
twjeffery: Removed. For these params `type` already carries the name and `values` carries the list, so `typeLabel` was always the same value. Props keep their `typeLabel` since `type` works differently there.

docs/src/lib/content-queries.ts
twjeffery: Done. It's `Omit<PropDefinition, "default" | "deprecated" | "typeLabel">` now, with `typeLabel` in the list too since it's gone from the static method params.

docs/src/scripts/extract-api.ts
twjeffery: Reworded, though it says something different now. After the change from your other comment, the params and return types come from the function signatures too, so the caveat about them not being verified no longer applies. The comment now says everything is parsed from the controller source.

docs/src/scripts/extract-api.ts
twjeffery: Removed, along with the matching field in content-queries.

docs/src/scripts/extract-api.ts
twjeffery: Went with this. Everything the docs show is now read from the controller file:

1. The method list comes from the exported `TemporaryNotification` object
2. Signatures, params, and return types come from the functions themselves
3. Descriptions come from new JSDoc on the functions and on the options type fields. `uuid` and `visible` are tagged `@internal`, which is what hides them from the docs now.

I also renamed `opts` to `options` in show() so the signature reads the way the docs do, and reordered the options fields so the table order follows the source. After the change the regenerated JSON only differs in the ways you asked for in the other comments, so nothing else moved.

One gap this surfaced: the pre-commit hook in #3980 wasn't watching libs/common, so editing the controller wouldn't trigger a regeneration. I added it to the watched paths on that PR.

docs/src/scripts/extract-api.ts
```diff
@@ -2076,6 +2093,250 @@ function findSvelteDeclaringTag(tag: string): string | undefined {
...
+// the function declarations, descriptions from their JSDoc (@param tags for
+// positional params), and option fields from the options type, where members
+// tagged @internal are omitted.
+
+interface StaticMethodSource {
+  source: string; // path to the controller source file, relative to the workspace root
+  namespace: string; // exported const whose object members are the public methods
+}
+
+const STATIC_METHOD_SOURCES: Record<string, StaticMethodSource> = {
```
vanessatran-ddi: Can we move this const to line 35 and above, where we define the path from others

docs/generated/component-apis/temporary-notification.json
vanessatran-ddi: Great work!

vanessatran-ddi (approved):
Looks good!

docs/src/scripts/extract-api.ts
twjeffery: Done. Moved the const and its `StaticMethodSource` interface up into the configuration block with the other path constants, and added a short comment pointing to the static-methods section so the parsing logic is still easy to find.

willcodeforcoffee (approved):
LGTM, thanks Tom!

PR #3981, opened 2026-05-26
feat(#3789): add AI tools and resources section to get-started

twjeffery:
## What this is

A first version of the AI tools and resources area in Get Started: an overview of the AI tools for working with the design system, plus a full MCP setup guide for Claude Code, Claude Desktop, Cursor, and ChatGPT.

## Outstanding questions
The overview page leaves two "how to access it" spots open on purpose:
1. **Where the MD files live** and how teams download them.
2. **How teams install the skill** (the using-goa-design-system workflow): clone, download, drop in a folder?

Closes #3789.

PR #4023, opened 2026-06-05
feat(#4022): add skills/ folder

### LINKED ISSUE #4022: share our AI skills for teams to consume
We have two AI skills for the design system, and we need a way for teams to get them. This outlines what they are, how we'll share them, and why.

## The skills

Two skills, and they're different kinds of thing:

- **using-goa-design-system** (the navigator): from "I'm building X for users", it walks to the right product type, templates, and components. Tied to the component library and the MCP.
- **content-design** (the writer): designs user-facing copy for its reader, a citizen or a worker. Self-contained, no component dependency.

More will follow (a public-form skill, and others), but these two are ready now.

## The problem

Teams install our packages, not our repo files, so a skill sitting in our repo isn't reachable to them. We need a deliberate way to share skills.

## How we'll share them

- A public `skills/` folder in this repo, one folder per skill (a `SKILL.md` plus any supporting files).
- Teams install a skill with one command, pointed at this repo: `npx skills add GovAlta/ui-components --skill using-goa-design-system`
- It loads on its own when their work matches, and `npx skills update` keeps it current.
- Skills track `dev`, so updates reach teams on the skill's own cadence, separate from component releases.

## Why this way

- **It's a standard.** Skills are an open format now (SKILL.md), read by Claude Code, Cursor, Copilot, and more. The same files work everywhere, no custom setup.
- **No outside dependency.** Teams install them the same way they already install our package, and we don't rely on any third-party service.
- **One home.** Skills sit beside the MCP, and eventually the design tokens, so there's one place to maintain and find them.

twjeffery:
## What this does

Adds a public skills/ folder so consuming teams can install our AI skills the same way they install our package. Skills are an open standard now (SKILL.md), so these work across Claude Code, Cursor, Copilot, and other tools that read the format.

## What's in it

A skills/ folder at the repo root (modelled on apps/), with two skills:

- using-goa-design-system: the navigator. Turns "I'm building X for users" into the right product type, templates, and components, pulling live facts from the design system MCP.
- content-design: the writer. Designs user-facing copy for its reader, a citizen or a worker.

Each folder has a SKILL.md (what the AI reads) and a README.md (what it is, for people). A top-level README has the install commands and the skill index. No build wiring; it's a static folder teams fetch from.

## How teams use it

    npx skills add GovAlta/ui-components --skill using-goa-design-system
    npx skills add GovAlta/ui-components --skill content-design

The skill loads on its own when the work matches. `npx skills update` keeps it current.

## How to test

- Read the three READMEs (top-level plus one per skill) and check the framing lands.
- Optionally run an install command above in a scratch project and confirm the skill lands in your tool's skills folder.

## Notes

- Skills track dev, so updates reach teams on the skill's own cadence, separate from component releases.
- Pairs with the AI tools and resources page (#3981). Its skill install commands go live once this merges.
- Opening as a draft; I'll mark it ready after review.

Closes #4022

### DIFF
--- a/skills/content-design/SKILL.md
+++ b/skills/content-design/SKILL.md
@@ -1,0 +1,131 @@
+---
+name: content-design
+description: "Designs user-facing service content (service names, descriptions, guidance, microcopy, button labels, status messages, errors, empty states) for its specific reader: a citizen (public, plain language, grade-9) or a worker (internal, dense, domain-expert). Use when writing or editing any citizen-facing or worker-facing copy for a Government of Alberta service, or when checking whether existing copy fits the reader it is actually for."
+---
+
+# Designing content for its reader: citizen or worker
+
+Given a piece of content (or a brief to write one), design it for whoever reads it. **Name the reader first.** It is the master switch: the same state said to a citizen and to a worker is two different designs, not one lightly reworded, because "good" means opposite things to them.
+
+## When to use it
+
+- Writing or editing any user-facing words in a Government of Alberta service: service names, descriptions, guidance, button and link labels, status messages, empty states, errors, confirmations, notifications.
+- Stress-testing existing copy against its actual reader. The two recurring failures: citizen copy written at a policy reading level, and worker copy padded with hand-holding that slows an expert down.
+
+Not for your team's own internal docs in your own voice (those follow your team's voice guide), though the worker profile is a fair lens for any fluent internal audience.
+
+## Step 1 — Name the reader (first, every time)
+
+Fix three things before writing a word:
+
+1. **Audience: citizen or worker?** Grounded in the design system's own user-type model (the **User Types** foundation, also served through the MCP):
+   - **Citizen** — accessing a service from outside government, maybe once in their life, once a year, or rarely. They didn't choose to be here; they need something. The experience should feel like a clear, considered path: slow on purpose, one step at a time, hard to make a mistake. Plain language, ~grade 9, minimal density. Values clarity over speed, confidence over efficiency.
+   - **Worker** — internal staff using a tool to deliver or administer a service from inside, every day; it's their job and they know it. The experience should feel like a power tool: fast, dense, information-rich, built for repetition and scale. Density fine, efficiency-focused, trusts expertise. Splits into an intake worker (deciding on a submission) and an ongoing case worker ("what needs me on this file").
+   - **Frequency can outrank role.** The mode is the reader's relationship to the task, not their title. A contractor filing daily permits is a citizen but wants worker-terse content; a caseworker handling a once-in-a-career exception wants citizen-style guidance. When the two disagree, frequency usually wins for content too.
+   - **Can't tell? Ask. Do not guess.** A wrong audience invalidates everything downstream.
+2. **Service task and lifecycle stage.** A service is a verb; *apply*, *check status*, and *report a change* read differently. The same stage often has a citizen-facing and a worker-facing track, said two different ways.
+3. **Surface and moment** — a button, page intro, error, empty state, confirmation, notification. The moment sets the job the words do.
+
+## Step 2 — Apply the busy-reader base, dialed to the reader
+
+The six principles (Rogers & Lasky-Fink, *Writing for Busy Readers*) hold for everyone: enough formatting, design for navigation, less is more, make reading easy, tell readers why they should care, make responding easy.
+
+What changes is their settings. **"Make reading easy" and "use enough" are defined relative to the reader's expertise and frequency.**
+
+| Dimension | Citizen (public, occasional, novice) | Worker (internal, frequent, expert) |
+|---|---|---|
+| Reading level | Plain language, ~grade 9 | Domain-fluent; precise terms are precision, not jargon |
+| Density | Minimal; one idea at a time; whitespace | Dense is fine; completeness and scannability over hand-holding |
+| Vocabulary | Everyday words; **verbs, not program names** ("report a change", not "Income Reporting Module"); expand acronyms, unless one is more familiar to this reader than its expansion (SIN, AISH on a recipient's own portal) | Real domain terms, program names, case states, stage labels |
+| Orientation | High — what is this, what's next, why it matters to me | Low — orient to *this file or task*, not the system |
+| Tone | Calm, reassuring, human | Efficient, direct, action-first |
+| Job | Help one person do one thing, confidently | Help an expert decide and move work |
+| "Less is more" → | Cut everything non-essential | Cut noise, not data; the bar for "needed" is higher |
+| "Responding easy" → | One obvious action, no dead ends | Quick actions, batch where it fits; don't over-confirm |
+
+Stakes amplify the tone dial: the higher the consequence to the reader (money, health, legal standing), the more a citizen needs calm and reassurance, and the less a worker wants anything slowing them down.
+
+Both readers, non-negotiable (not dials): WCAG 2.2 AA, government voice, and **no citizen PII** — roles, never people.
+
+**Cutting wordiness has a lookup.** For the "less is more" and "everyday words" dials, a companion reference lists ~860 common wordy or formal phrases and their plain replacements (`content-design-succinct-alternatives.md`). Use it reader-awarely: most entries cut padding ("due to the fact that" → "because") and help every reader; some swap a formal word for a plain one ("terminate" → "stop"), which suits a citizen but can wrongly strip a worker's precise vocabulary. Match the substitution to the reader; never blanket-replace.
+
+### Surface-specific moves
+
+A few patterns the dial table doesn't spell out, surfaced by real use:
+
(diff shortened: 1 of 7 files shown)


willcodeforcoffee (approved):
I like this a lot. It makes me think maybe we should have skills or rules on how to layout pages with Block and Container? What kind of padding or colours to use when? Especially with tokens?
That might be getting a little too focused, but it got me thinking.
LGTM for these though!

twjeffery:
> I like this a lot. It makes me think maybe we should have skills or rules on how to layout pages with Block and Container? What kind of padding or colours to use when? Especially with tokens? That might be getting a little too focused, but it got me thinking. LGTM for these though!

@willcodeforcoffee Yeah, I think those are all good ideas! I think there's a balance around what's already part of and baked into the UI components themselves in terms of the padding and type and everything else that is a property, and design tokens used. 

**But there's again an interesting difference that comes up between:** 
A) skills for us as design system maintainers and builders 
B) skills for product teams to use the design system and build their products

Either way, worth exploring both sides further!

skills/using-goa-design-system/SKILL.md
```diff
@@ -0,0 +1,85 @@
+---
+name: using-goa-design-system
+description: Use when building or scoping a screen, page, or feature with the Government of Alberta Design System starting from a user-facing intent ("worker tool for case management", "public form for licence renewal", "error page for payment failure") rather than from a specific component or token.
+---
+
+# Using the GoA Design System
+
+## Overview
+
```
vanessatran-ddi, line 9: This will be read by agent, so we need to inform that the prequisites to run this skill is to have GoA design system MCP running, write this only in README.md will be ignored by agent

skills/using-goa-design-system/SKILL.md
```diff
@@ -0,0 +1,85 @@
...
+| Example by size | Scale of artifact | `interaction` → `section` → `page` → `task` → `product` |
+| Components | Atomic UI building blocks | `goa-button`, `goa-form-item`, `goa-table` |
+| Guidance atoms | Do/don't/tip/warning/info linked to component + topic | "use goa-block for spacing, not goa-container" |
+| Tokens | Sizing, color, spacing values | `--goa-space-m`, `--goa-color-text-default` |
+
+Service type is the vocabulary layer (Kate Tarling), not a schema layer yet. The skill leads with it when recognizable, then narrates the mapping to product type. See `taxonomy.md` for the full framework.
+
+Public services typically have two sides: one to **provide and manage** (worker), and one to **receive** (citizen). When the intent names one side, name it. When the intent could apply to either, name both. The product type follows from which side the developer is building for.
+
+User type (citizen vs worker) is **derived from product type**, not asked. `workspace` = worker. `public-form` = citizen.
```
vanessatran-ddi: to save the usage of tokens, I recommend to shorten this and reduce the duplication, I can see at line 38 repeated again at line 77 below, and also repeated again from line 40 to 50 on `taxonomy.md`, I think we can keep taxonomy.md and remove this on this line 38, keep line 77

skills/using-goa-design-system/SKILL.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Common mistakes
+
+- **Skipping the service type.** Service language ("intake service", "permit application service") deserves explicit acknowledgment, not silent translation to product type. Name the service type, then narrate the mapping to product type.
+- **Skipping the product type layer.** Going straight to components produces correct-looking code that ignores the system's layered conventions.
+- **Asking for user type.** Don't. Derive from product type.
+- **Inventing tokens.** Always reference the actual token; never hardcode a value. If a value doesn't fit, use a component-level override; tokens themselves are referenced, not modified by teams.
+- **Treating guidance atoms as optional.** They encode known failure modes; surface them before code, not after review.
+- **Returning "not found" on an old slug.** Old slugs live in `aliases` on entries (e.g., `confirm-that-an-application-was-submitted` aliases to `result-page`). Treat aliases as additional lookup keys.
```
vanessatran-ddi: I think we can remove examples, shorten this line, point to taxomony.md line 80

skills/using-goa-design-system/taxonomy.md
```diff
@@ -0,0 +1,112 @@
...
+Don't promote service types into a structured field; use them as intent vocabulary until the framework is formally adopted in the schema.
+
+## Aliases (slugs)
+
+Old slugs are preserved in the `aliases` array on entries for search continuity and URL redirects. When a developer references a slug that doesn't resolve directly, look up entries whose `aliases` contains that slug. Common cases include:
+
+- `confirm-that-an-application-was-submitted` → `result-page`
+- `ask-a-user-one-question-at-a-time` → `question-page` (one of the six inline variants)
+- `give-more-information-before-asking-a-question-a` → `question-page`
+- 9 question-page variants now live as inline sections inside `/examples/question-page/`
```
vanessatran-ddi, line 87: Here we said 9 variants but on line 85 we said `one of six`, are they different?

skills/using-goa-design-system/taxonomy.md
```diff
@@ -0,0 +1,112 @@
...
+
+**Don't ask the developer for user type. Derive it.**
+
+| Product type | User type |
+|---|---|
+| `workspace` | worker |
+| `public-form` | citizen |
+| (none / interaction-only) | both |
+
+This derivation replaces the previous `userType` field on examples (removed in PR 3888).
```
vanessatran-ddi: Do we need `PR 3888`? perhaps we can remove it

skills/content-design/SKILL.md
```diff
@@ -0,0 +1,132 @@
...
+```
+
+**Citizen** (recipient's portal, "where is my file"):
+
+```
+We're reviewing your application
+
+Right now we're checking your medical eligibility. You don't need to do anything.
+
+We'll text and email you as soon as there's a decision. Most take about 8 weeks.
```
vanessatran-ddi, line 96: I think you need to add the information `8 weeks` or something like that in the undesigned draft message on line 82 above, otherwise it looks like this information comes from nowhere, and will be conflicted with our rule defined on line 72 above

skills/content-design/SKILL.md
```diff
@@ -0,0 +1,132 @@
...
+```
+
+Why: leads with status and "you don't need to do anything" (the anxious reader's real question); grade-9, dropped "in due course", "determination", and the legal program name; four lines, one idea each; ends on what's next and when.
+
+**Worker** (case worker's file view, "what needs me on this file"):
+
+```
+Stage 3 · Awaiting medical eligibility · 6 days in queue
```
vanessatran-ddi, line 104: Same here, we can add a clue or information about timing on the undesigned draft above

skills/content-design/SKILL.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Extending it (a first foundation)
+
+The base is one set of principles; audiences are profiles, so it grows without restructuring:
+
+- More **profiles** as they earn it: a delegated third party (trustee, POA); a screen-reader-first reader named explicitly rather than treated as an afterthought.
+- More **principles**: deepen the surface-specific moves (errors, empty states, notifications), plus service-language conventions.
+- Stay **aligned with the code side**: this is the prose counterpart to the design system's User Types foundation. When one moves, check the other.
+
+To harden it the best-practice way: run it on a few real pieces of content with a fresh agent and refine where the output misses (evaluation-driven iteration), rather than trusting it on first write.
```
vanessatran-ddi: When I read this, looks like this section is for human being, not for agent right? especially `run it on .. with a fresh agent`, if yes we can remove it

vanessatran-ddi (approved):
Looks good!

skills/using-goa-design-system/SKILL.md
twjeffery: Good call, added the MCP prerequisite to `SKILL.md` so the agent picks it up.

skills/using-goa-design-system/SKILL.md
twjeffery: Good point, removed it. The mapping stays in `taxonomy.md` and the rule stays in common mistakes.

skills/using-goa-design-system/SKILL.md
twjeffery: Good point, shortened it and pointed to `taxonomy.md` for the examples.

skills/using-goa-design-system/taxonomy.md
twjeffery: Good catch, they should match. It's nine, I fixed the two spots that said six.

skills/using-goa-design-system/taxonomy.md
twjeffery: Good point, removed it.

skills/content-design/SKILL.md
twjeffery: Good catch, added the timing and contact info to the draft so it's not coming from nowhere.

skills/content-design/SKILL.md
twjeffery: Same here, the draft now includes the queue age so the worker version pulls it from there.

skills/content-design/SKILL.md
twjeffery: Yeah, that one was for whoever maintains the skill, not the agent. Removed it.

PR #4029, opened 2026-06-08
fix(#3763): size dropdown and date picker popover to measured width for percentage widths

### LINKED ISSUE #3763: Width 100% on dropdown has container expanding too much
# Info

If you use any percentage width (not just `width="100%"`) with the Dropdown right now, it causes the actual open state of the Dropdown to expand far past the viewport in width.

This is happening because the Dropdown passes the same width it's assigned to the Popover (so 100% for example). That width is then applied to the Popover container, where the percentage is not guaranteed to resolve against the same parent container as the Dropdown.

## More Info

It was reported by Vidit that this issue also exists for Datepicker

## Acceptance Criteria:

1. When using percentage width on Dropdown, the open width should match the closed width
2. When using percentage width on Datepicker, the open width should still not exceed the maximum width

vanessatran-ddi:
# Before (the change)
image

# After (the change)

image

## Steps needed to test
Angular link: https://govalta.github.io/ui-components/pr-preview-angular/pr-4029/bugs/3763
React link: https://govalta.github.io/ui-components/pr-preview-react/pr-4029/bugs/3763

### DIFF
--- a/libs/web-components/src/components/date-picker/DatePicker.svelte
+++ b/libs/web-components/src/components/date-picker/DatePicker.svelte
@@ -12,6 +12,7 @@
   import type { Spacing } from "../../common/styling";
   import { toBoolean } from "../../common/utils";
   import { receive, dispatch, relay } from "../../common/utils";
+  import { isValidDimension } from "../../common/validators";
   import {
     FieldsetSetValueMsg,
     FieldsetSetValueRelayDetail,
@@ -82,6 +83,12 @@
     addRelayListener();
     sendMountedMessage();
     showDeprecationWarnings();
+
+    if (width && !isValidDimension(width)) {
+      console.error(
+        "DatePicker width must be a valid CSS dimension (e.g. 50%, 320px, 16ch)",
+      );
+    }
   });
 
   function showDeprecationWarnings() {
@@ -266,7 +273,7 @@
         bind:this={_rootEl}
         tabindex="-1"
         data-testid="calendar-popover"
-        data-content-fits-width="true"
+        fitcontent="true"
         {testid}
         {mt}
         {mb}
--- a/libs/web-components/src/components/dropdown/Dropdown.svelte
+++ b/libs/web-components/src/components/dropdown/Dropdown.svelte
@@ -28,6 +28,7 @@
     toBoolean,
   } from "../../common/utils";
   import { calculateMargin } from "../../common/styling";
+  import { isValidDimension } from "../../common/validators";
   import {
     FieldsetErrorRelayDetail,
     FieldsetResetErrorsMsg,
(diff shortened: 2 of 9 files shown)


libs/web-components/src/components/date-picker/DatePicker.svelte
```diff
@@ -82,8 +84,26 @@
...
+    }
   });
 
+  function setupPercentageWidth() {
+    // cannot use anchor size because chrome/edge < 124, firefox < 146, safari, safari iOS < 26 not supported
+    _popoverWidth = getRenderedCalendarWidth();
+
+    _rootEl?.addEventListener("_open", () => {
+      _popoverWidth = getRenderedCalendarWidth();
+    });
```
willcodeforcoffee: If we don't `removeEventListener` we have a memory leak here.

willcodeforcoffee (changes requested):
I've only got one issue, I'm a little worried about this `eventListener` not getting removed.

libs/web-components/src/components/date-picker/DatePicker.svelte
vanessatran-ddi: the listener is on `_rootEl` not on `document/window` so it is together with the component DOM subtree on teardown. Also in `.claude/rules/component-authoring.md` line 51 said we `do not add removeAddEventListener`

willcodeforcoffee: @vanessatran-ddi my mistake! I didn't realize that about `_rootEl`.
Thank you!

willcodeforcoffee (approved):
LGTM, thanks Vanessa!

libs/web-components/src/components/date-picker/DatePicker.svelte
```diff
@@ -82,8 +84,26 @@
     addRelayListener();
     sendMountedMessage();
     showDeprecationWarnings();
+
+    if (type === "calendar" && width.includes("%")) {
+      setupPercentageWidth();
```
bdfranck: @vanessatran-ddi Would it be helpful to add a console warning if someone accidentally enters an invalid width? (ex: `5 0%`)

I noticed that we do it in PageBlock:
https://github.com/GovAlta/ui-components/blob/e33682cbe2fac9ca8a0925dad1d82fcfaeb5d6f4/libs/web-components/src/components/page-block/PageBlock.svelte#L33-L35

This may help us troubleshoot future width issues more quickly.

libs/web-components/src/components/dropdown/Dropdown.svelte
```diff
@@ -140,6 +140,12 @@
 
   $: calculateWidths(width, _options, _inputEl);
 
+  // Re-measure on open so the popover tracks the current rendered width after
+  // container resizes or layout that settled after mount.
+  $: if (_isMenuVisible && (width?.includes("%") || maxwidth?.includes("%"))) {
```
bdfranck, line 146: We could throw a warning for an invalid width property here as well.

vanessatran-ddi: We have a console message at line 195 before, I refactor to use the common `isValidDimensions`

bdfranck (approved):

- ✅ The calendar is sized correctly with a small date picker with a percentage width

Image

- ✅ The calendar is sized correctly with a large date picker with a percentage width

Image

- ✅ I see an console error when I set an invalid width on a Date Picker

Image

- ✅ I see an console error when I set an invalid width on a Dropdown

Image

- ✅ Both Angular and React work as expected

Looks good to me! 👍

PR #4034, opened 2026-06-09
fix(#3662): remove unintentional MenuButton popover toggle

### LINKED ISSUE #3662: Menu button: closing the menu swallows the next click event
When closing a menu button dropdown, the next click event on the page gets swallowed. For example, if you click outside the menu to close it, clicking a button immediately after doesn't register.

https://github.com/user-attachments/assets/5d0274d2-7e7c-4594-958c-9e06ac2b80dd

## Acceptance Criteria
1. Clicking somewhere off the menu button should activate any other interactive element, and should close the menu button

willcodeforcoffee:
# Before (the change)

Opening a MenuButton when another was open would cause both MenuButtons to be closed. See https://github.com/GovAlta/ui-components/issues/3662 for details.

The Popover is receiving a toggle message where the `newState` and `oldState` are both `'closed'`.

# After (the change)

When a Popover toggle message has both states `'closed'` it will open.
The closing Popover has `oldState === 'open'` so this is valid.

## Steps needed to test

Have two `MenuButton` controls on a page. Open one, then open the second. You will see one of two things:

- The open MenuButton will close
- In some situations you may see a brief flash of the second MenuButton toggle open, but then immediately both will close.

## Story

This PR closes https://github.com/GovAlta/ui-components/issues/3662

### DIFF
--- a/docs/src/scripts/extract-api.ts
+++ b/docs/src/scripts/extract-api.ts
@@ -154,7 +154,11 @@ interface WrapperExtraction {
   slotRequired: Record<string, boolean>;
 }
 
-const INTERNAL_PROP_NAMES = new Set(["publicformsummaryorder", "filterablecontext", "version"]);
+const INTERNAL_PROP_NAMES = new Set([
+  "publicformsummaryorder",
+  "filterablecontext",
+  "version",
+]);
 const INTERNAL_SLOT_NAMES = new Set(["version"]);
 const INTERNAL_EVENT_NAMES = new Set(["_revealChange", "_update"]);
 
@@ -176,6 +180,7 @@ const WEB_COMPONENT_INTERNAL_EVENT_OVERRIDES: Record<string, Set<string>> = {
     "_mountItem",
     "_navigate",
   ]),
+  "menu-button": new Set(["close"]),
 };
 const WEB_COMPONENT_EVENT_TYPE_OVERRIDES: Record<string, Record<string, string>> = {
   dropdown: {
@@ -224,7 +229,9 @@ function shouldSkipInternalProp(componentName: string, propName: string): boolea
 function shouldAllowInternalProp(componentName: string, propName: string): boolean {
   const normalizedComponentName = toKebabCase(componentName);
   return Boolean(
-    ALLOW_INTERNAL_PROP_BY_COMPONENT[normalizedComponentName]?.has(propName.toLowerCase()),
+    ALLOW_INTERNAL_PROP_BY_COMPONENT[normalizedComponentName]?.has(
+      propName.toLowerCase(),
+    ),
   );
 }
 
@@ -235,7 +242,9 @@ function isInternalPropName(propName: string): boolean {
 function isStandardV1V2VersionProp(prop: ExtractedProp | undefined): boolean {
   if (!prop || prop.name.toLowerCase() !== "version") return false;
 
-  const values = (prop.values || []).map((value) => String(value).replace(/['"]/g, "").trim());
+  const values = (prop.values || []).map((value) =>
+    String(value).replace(/['"]/g, "").trim(),
+  );
   if (values.length === 2 && values.includes("1") && values.includes("2")) {
     return true;
   }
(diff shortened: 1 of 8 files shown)


libs/web-components/src/components/popover/Popover.svelte
```diff
@@ -298,6 +298,14 @@
   function handleNativeToggle(toggleEvent: ToggleEvent) {
     if (toggleEvent.newState === "open") {
       _isOpen = true;
+    } else if (
```
vanessatran-ddi: I think this fix will make difficult to understand later. 

My suggestion, the fix is not on `Popover.svelte`, it is on the component using this Popover.svelte. 

Image

So right now, MenuButton wants to close, but close globally instead of closing itself only, makes the this function on Popover not work as expected

Image

Same with Datepicker where we set the `dispatch` with global instead of the component.

vanessatran-ddi: For this fix, can we review this browser test and see why it didn't fail? 

Image

vanessatran-ddi (changes requested):

libs/web-components/src/components/popover/Popover.svelte
willcodeforcoffee: @vanessatran-ddi I'm pretty sure this spec passed because the `Popover` worked. I am pretty sure the problem was in the `MenuButton` and how it listened for `goa:closePopover`.

willcodeforcoffee: @vanessatran-ddi brilliant! You are correct. That solved it and did it better.

vanessatran-ddi (approved):
I forgot to add another comment, can we add a browser test for MenuButton to test this bug, similar like what we have for Popover. 
I also approve.

willcodeforcoffee:
I have added a browser spec. That was a good idea, @vanessatran-ddi . 👍

libs/web-components/src/components/menu-button/MenuButton.svelte
```diff
@@ -119,7 +119,8 @@
 
   function close() {
     _icon = "chevron-down";
-    dispatch(document.body, "goa:closePopover", { target: _rootEl });
+    // dispatch(document.body, "goa:closePopover", { target: _rootEl });
```
ArakTaiRoth: No need to comment this out. Just remove it if it's wrong, if we want to go back to it that's why we have version control.

libs/react-components/specs/menu-button.browser.spec.tsx
```diff
@@ -221,4 +230,38 @@ describe("GoabMenuButton", () => {
...
+      );
+    };
+
+    const result = render(<Component />);
+    const menu1 = result.getByTestId("menu-1");
+    const menu2 = result.getByTestId("menu-2");
+
+    // Open first menu only
+    await menu1.click();
+    expect(result.getByTestId("action-1")).toBeVisible();
```
ArakTaiRoth: I could be wrong, but I think all these expect statements after click events should be wrapped in a `await vi.waitFor()`. Just because the menus can open asynchronously, so these could happen out of order.

docs/generated/component-apis/menu-button.json
```diff
@@ -236,6 +236,14 @@
           "frameworks": [
             "webComponents"
           ]
+        },
+        {
```
ArakTaiRoth: I don't think we want to document this close event. The close event in `MenuButton.svelte` is exclusively used for Popover use, it's not for public use.

libs/web-components/src/components/menu-button/MenuButton.svelte
willcodeforcoffee: 😦 Sorry, commenting out code is a bad habit.

willcodeforcoffee:
Addressed @ArakTaiRoth 's feedback:
- removed commented code 😬 
- wrapped tests in `waitFor`
- removed event documentation for `close` which is only emitted for Popover

docs/src/scripts/extract-api.ts
```diff
@@ -154,7 +154,11 @@ interface WrapperExtraction {
   slotRequired: Record<string, boolean>;
```
ArakTaiRoth, line 154: I haven't reviewed the changes made in this file yet. I'm just wondering, why are there so many changes to this file in this PR? As far as I can tell, this PR shouldn't have to make any changes at all to this extract script.

ArakTaiRoth: Oh, correction, were all these changes necessary to remove `close` event documentation? Roughly 300 lines of code needed to be changed for that one thing?

willcodeforcoffee: It looks like it was committed before anyone ran it through Prettier... All those changes are because VS Code does Prettier formatting on save. I didn't notice it happening at the time.
And you're correct, it was to remove the `close` event in the documentation..
¯\\\_(ツ)\_/¯

ArakTaiRoth (approved):

PR #4035, opened 2026-06-09
fix(#3683): vertically center text values for date and time input types

### LINKED ISSUE #3683: Input refinements
## Overview

## Acceptance Criteria:
- Vertically align text values for Date and Time "types"

Spark450:
## Overview

For `type="date"`, `type="time"`, `type="datetime-local"`, `type="month"`, and `type="week"` inputs in V2, the displayed value was visually lower than the value in a `type="text"` input. The browser's native date/time widget centers itself within the full 56px input height, while regular text follows the asymmetric padding (9px top / 11px bottom) and sits higher.

## Fix

Added `align-items: center` to the `.goa-input` container for all date/time types in V2. This lets the input size naturally (45px) and centers it within the 56px container so the browser widget renders at the correct vertical position.

CSS-only change. React and Angular wrappers are unchanged.

## Before / After

**Before:** Date/time values rendered lower in the input box than text values.

**After:** Date/time values vertically aligned with text values.

## Steps to test

1. Run `npm run serve:prs:react` and navigate to **Bugs > 3683 Input date/time vertical alignment**
2. Compare the reference `type=text` input against each date/time type
3. Confirm the values sit at the same vertical position in the input box
4. Run `npm run serve:prs:angular` and repeat at the same route

## Checklist

- [x] Setup steps read
- [x] Unit tests pass (65/65)
- [x] Tested in React playground
- [x] Angular playground page included

Fixes #3683

### DIFF
--- a/apps/prs/angular/src/routes/bugs/3683/bug3683.component.html
+++ b/apps/prs/angular/src/routes/bugs/3683/bug3683.component.html
@@ -1,0 +1,32 @@
+<goab-block direction="column" gap="l" mt="l">
+  <goab-text tag="h1" mt="none">Bug #3683: Input date/time vertical alignment</goab-text>
+  <goab-text tag="p">
+    The text value in date and time input types should be vertically centered in the
+    input box, matching how type=text renders. Compare each type below against the
+    reference text input.
+  </goab-text>
+
+  <goab-form-item label="Reference: type=text">
+    <goab-input name="text" type="text" [value]="textVal"></goab-input>
+  </goab-form-item>
+
+  <goab-form-item label="type=date">
+    <goab-input name="date" type="date" [value]="dateVal"></goab-input>
+  </goab-form-item>
+
+  <goab-form-item label="type=time">
+    <goab-input name="time" type="time" [value]="timeVal"></goab-input>
+  </goab-form-item>
+
+  <goab-form-item label="type=datetime-local">
+    <goab-input name="datetime-local" type="datetime-local" [value]="datetimeVal"></goab-input>
+  </goab-form-item>
+
+  <goab-form-item label="type=month">
+    <goab-input name="month" type="month" [value]="monthVal"></goab-input>
+  </goab-form-item>
+
+  <goab-form-item label="type=week">
+    <goab-input name="week" type="week" [value]="weekVal"></goab-input>
+  </goab-form-item>
+</goab-block>
--- a/apps/prs/angular/src/routes/bugs/3683/bug3683.component.ts
+++ b/apps/prs/angular/src/routes/bugs/3683/bug3683.component.ts
@@ -1,0 +1,17 @@
+import { Component } from "@angular/core";
+import { GoabBlock, GoabFormItem, GoabInput, GoabText } from "@abgov/angular-components";
+
+@Component({
+  standalone: true,
+  selector: "abgov-bug3683",
+  templateUrl: "./bug3683.component.html",
+  imports: [GoabBlock, GoabFormItem, GoabInput, GoabText],
+})
+export class Bug3683Component {
+  dateVal = "2025-06-09";
+  timeVal = "09:30";
+  datetimeVal = "2025-06-09T09:30";
+  monthVal = "2025-06";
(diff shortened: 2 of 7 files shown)


bdfranck (commented):
Thanks for tackling this, @Spark450! I was struggling to replicate the issue until I realized that it was specific to Safari. Your approach looks good but introduces a tiny "dead zone" on the top and bottom edges of the input. In these areas, the input shows a hover state but isn't interactive.

~I've added a commit with an alternative approach that specifically targets Safari. It doesn't resize the input so the label is aligned but the input has no dead zones. Does that work for you?~

bdfranck:
D'oh! My above approach misaligned the labels in Chrome. 🤪 Never mind, I've removed the commit.

I'll start by showing the issue to help us make a decision. There are 2px - 6px at the top and bottom of these inputs that show a hover state but don't open the input. This is because the input is shorter than the container. Do we want to try to resolve this issue?

image

date input dead zone

Spark450:
> D'oh! My above approach misaligned the labels in Chrome. 🤪 Never mind, I've removed the commit.
> 
> I'll start by showing the issue to help us make a decision. There are 2px - 6px at the top and bottom of these inputs that show a hover state but don't open the input. This is because the input is shorter than the container. Do we want to try to resolve this issue?
> 
> image date input dead zone

No worries @bdfranck I worked with Claude on another solution, which seemed to be working better. Just a heads up they asked specifically to check how it was working in Firefox. On my end (in Firefox) it looked like it was working.

bdfranck:
@Spark450 Looks great! I'm testing it now. I pushed a change with a couple small cleanups:

- I shortened the comment a bit. Comments are most helpful when they explain the "why" behind the code rather than restating what the code already does.
- I simplified the CSS selector to a single line since `::-webkit-datetime-edit` only shows on date/time fields
- I squashed the commits to a single one and rebased it against the latest `dev` branch.

bdfranck (approved):

- ✅ I can see that input fills the height of the container

Image

- ✅ The labels are vertically centred in the latest version of Firefox

Image

- ✅ The labels are vertically centred in the latest version of Safari

Image

- ✅ The labels are vertically centred in the latest version of Chrome

Image

Looks good to me! 👍

willcodeforcoffee (changes requested):
I did some mobile testing using BrowserStack. Things look really nice on desktop, but iPhone Safari is looking like it has too much bottom spacing:

Image
(iPhone 14 Safari)

Image
(iPhone 17e Chrome)

I saw the same alignment issues on iPhone 14, iPhone 17 Pro Max and iPhone 17e. 😞 

Ditto, iPad Pro 12 2021 Safari - same problem:

Image

Android Chrome looked good though:

Image

Really sorry, I'm going to request changes. It is broken in Safari on all iOS devices. 😭

Spark450:
Hi @willcodeforcoffee Thanks for the review 🙏 
image

I played around a bit and I think I came to a good solution. Just a heads up with this (hover: none) solution there could be an edge case (iPad with Magic Keyboard) where the text alignment issue is still present.

I'm not sure if that's a combo you can check, but if you can test it that would be great 👍

willcodeforcoffee (approved):
Unfortunately I cannot test with a real iPad, but it does correct now on iPad and iPhone!
LGTM

bdfranck:
@Spark450 I cleaned up the PR with the following changes:

- Removed the media queries. We don't need them since the CSS does the same thing for both selectors.
- Simplified the comment.
- Squashed the commits and rebased against latest `dev`.

I'll test against desktop and mobile browsers.

bdfranck (approved):

### Desktop browsers

- ✅ It is aligned in the latest desktop Chrome

Image

- ✅ It is aligned in the latest desktop Firefox

Image

- ✅ It is aligned in the latest desktop Safari

Image

### Mobile browsers

- ✅ It is aligned in the latest mobile Chrome

Image

- ✅ It is aligned in the latest mobile Safari

Image

Looks good to me! 👍

Spark450:
Thanks for the improvements @bdfranck 🙏

PR #4036, opened 2026-06-09
fix(#3837): prevent tab label horizontal shift on selection change

### LINKED ISSUE #3837: Tabs: Tab label should not shift horizontally when selection changes
## Description

When switching between tabs, the tab labels shift horizontally. This is typically caused by a bold font weight being applied to the active tab, causing the text to take up more space and nudging adjacent tabs out of position.

## Steps to Reproduce

1. Render the tabs component with multiple tabs
2. Click between tabs and observe the tab label positions shifting

## Suggested Approach

Use a `::after` pseudo-element on each tab label that mirrors the tab text content with the bold font weight applied but set to invisible (`visibility: hidden; height: 0; overflow: hidden`). This reserves the space the bold text would occupy without rendering it, keeping the layout stable regardless of which tab is active.

```css
.tab-label::after {
  content: attr(data-label);
  font-weight: var(--goa-font-weight-bold);
  visibility: hidden;
  height: 0;
  overflow: hidden;
  display: block;
}
```

## Acceptance Criteria

- [ ] Tab labels do not shift horizontally when switching between tabs
- [ ] The active tab bold treatment is preserved

Spark450:
## Summary

- Extends the existing ghost text technique (already used by the segmented variant) to V2 default tabs
- Sets `data-text` on tab links when `version === "2"`, then uses a `::before` pseudo-element at bold weight with `height: 0; visibility: hidden; overflow: hidden` to pre-reserve the bold text width
- No changes to React or Angular wrappers. CSS-only fix in the Svelte component

## Before / After

**Before:** Switching tabs causes a ~3px horizontal shift in adjacent tab labels as the selected tab grows to accommodate bold text.

**After:** All tab labels stay at the same position. The bold treatment on the active tab is preserved.

## Known limitation

Slot headings that mix text and a badge (or other inline element) may still exhibit a small shift. The ghost text only captures the text content of the heading, not the full slot width. This is documented in Test 4 of the playground page.

## Steps to test

1. `npm run serve:prs:react`
2. Navigate to **Bugs > 3837 Tabs label shift on selection**
3. Click between tabs in each test case and verify labels do not shift
4. Repeat in the Angular playground: `npm run serve:prs:angular`

- [x] Setup steps read
- [x] Unit tests pass (`npx vitest run libs/web-components/src/components/tabs/`)
- [x] Tested in React playground
- [x] Angular playground page included

Fixes #3837

### DIFF
--- a/apps/prs/angular/src/routes/bugs/3837/bug3837.component.html
+++ b/apps/prs/angular/src/routes/bugs/3837/bug3837.component.html
@@ -1,0 +1,19 @@
+<goab-text tag="h1" mt="m" mb="s">Bug 3837 - Tabs: label should not shift horizontally when selection changes</goab-text>
+<goab-text mb="m">
+  Click between tabs and observe that labels do not shift horizontally. The active tab bold treatment should be preserved.
+</goab-text>
+
+<goab-text tag="h3">Test 1: Default variant</goab-text>
+<goab-tabs>
+  <goab-tab heading="Short">Short tab content</goab-tab>
+  <goab-tab heading="Much longer label">Longer tab content</goab-tab>
+  <goab-tab heading="Medium label">Medium tab content</goab-tab>
+  <goab-tab heading="Another tab">Another tab content</goab-tab>
+</goab-tabs>
+
+<goab-text tag="h3">Test 2: Similar-length labels</goab-text>
+<goab-tabs>
+  <goab-tab heading="Information">Information content</goab-tab>
+  <goab-tab heading="Documents">Documents content</goab-tab>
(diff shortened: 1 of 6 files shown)


willcodeforcoffee (approved):
Too bad it doesn't work with the badge. LGTM.

bdfranck (approved):
I looked at the changes...

tab-test-2

- ✅ The tabs work as expected in the latest Firefox
- ✅ The tabs work as expected in the latest Chrome
- ✅ The tabs work as expected in the latest Safari
- ✅ The tabs work both in Angular and React

Looks good to me! 👍

PR #4039, opened 2026-06-11
chore(#4038): add CI gate to keep generated docs files in sync

### LINKED ISSUE #4038: Add CI check to keep generated docs files in sync with source
Add a check to the PR workflow that regenerates the docs API files and search index, and fails if the committed files don't match source. Catches stale or missing generated files before they merge.

twjeffery:
## What

Adds a "Docs Freshness" check to the PR workflow. On every PR it regenerates the docs API files and search index from source and fails if the committed files don't match. It's the CI counterpart to the pre-commit hook from #3980, so local and CI enforce the same thing and what you review is what the docs site ships.

It also makes extract-api exit non-zero when a component fails to extract. Without that, a failed extraction leaves the old committed JSON untouched, so the check (and the hook, and the build) would pass while extraction is actually broken.

## Why

The docs build used to regenerate these files during the build, so the deployed site stayed fresh while the committed files could quietly drift. The hook gives fast local feedback; this check is the unskippable backstop. Together they make the committed files the single source of truth. Removing the build-time regeneration is a small separate follow-up, held until this check is green and required.

## Testing

Locally against dev: passes on clean dev, fails on a stale committed output, fails on a new untracked output (a component whose JSON was never committed), and fails when a component can't extract. Generators sort their output, so no cross-machine false-fails. actionlint clean.

Closes #4038

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -113,7 +113,9 @@ jobs:
             echo "source and are overwritten on regeneration, so don't hand-edit these files."
             echo ""
             git status --short -- "${outputs[@]}"
+            git add --intent-to-add -- "${outputs[@]}"
             git diff --stat -- "${outputs[@]}"
+            git diff -- "${outputs[@]}"
             exit 1
           fi
 
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -55,3 +55,66 @@ jobs:
           if [ -d "./dist" ]; then
             npm run test:pr
           fi
+
+  docs-freshness:
+    name: Docs Freshness
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    permissions:
+      contents: read
+
+    env:
+      # The generators don't need browsers; skip the Playwright download in npm ci.
+      PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD: "1"
+
+    steps:
+      - uses: actions/checkout@v5
+
+      - uses: actions/setup-node@v6
+        with:
+          node-version: "24"
+          cache: npm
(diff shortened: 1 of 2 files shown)


.github/workflows/pull-request.yml
```diff
@@ -55,3 +55,66 @@ jobs:
...
+            echo "To fix, from the repo root run:"
+            echo "    npm install"
+            echo "    cd docs && npm run extract-api && npm run build:search-index"
+            echo "then commit the updated files under:"
+            echo "    docs/generated/component-apis/   docs/public/search-index.json"
+            echo ""
+            echo "Note: extracted fields (types, JSDoc descriptions) come from the component"
+            echo "source and are overwritten on regeneration, so don't hand-edit these files."
+            echo ""
+            git status --short -- "${outputs[@]}"
```
vanessatran-ddi, line 115: Can we make `git diff` and list down the full difference? it will be easier to debug instead of hide it and show only file name, another reason is if we are using Mac it will have a similar env vs ubuntu, but if we use windows, sometimes we can generate with a hidden space or new line, or format change, so even json file has the same content but `git diff` will think it is different. the more information we can print out the easier to debug later

vanessatran-ddi (approved):
Looks good!

.github/workflows/pull-request.yml
twjeffery: Good call. It now prints the full `git diff` on failure instead of just the file names, so you can see exactly what changed, including whitespace or line-ending differences (you'll see markers like "no newline at end of file"). Used `git add -N` first so brand-new files show in the diff too.

willcodeforcoffee (approved):
Cool stuff! 👍LGTM

PR #4041, opened 2026-06-12
fix(#3665): prevent page jump when switching tabs

### LINKED ISSUE #3665: Tabs: page jumps when switching tabs programmatically
When switching between tabs (e.g. framework tabs on the docs site), the page scrolls/jumps to the tab content. This happens because `setCurrentTab()` in `Tabs.svelte` calls `el.focus()`, which causes the browser to scroll the focused element into view.

### Steps to reproduce

1. Go to any component page on the docs site (e.g. /components/button)
2. Scroll down so the code snippet tabs are partially visible
3. Switch the framework tab (React / Angular / Web Components)
4. Page jumps

### Expected

The page should stay in place when switching tabs.

### Root cause

`setCurrentTab()` calls `el.focus()` on the tab panel, which triggers the browser's default scroll-into-view behavior. This is especially noticeable when tabs are used for framework switching where the user is reading code and doesn't want the viewport to move.

## Acceptance Criteria

1. Set `navigation` property to "none", and see if that fixes the issue
    * Otherwise fix in component as described above

Spark450:
# Before (the change)

Switching tabs moved the viewport. `setCurrentTab()` in `Tabs.svelte` called `el.focus()` on the selected tab, which triggers the browser's default scroll-into-view behavior. On the docs site this is especially noticeable with the framework switcher tabs: switching React / Angular / Web Components while reading code jumped the page.

# After (the change)

`el.focus({ preventScroll: true })` keeps focus on the selected tab for keyboard accessibility, but the browser no longer scrolls the tab into view. The viewport stays put when switching tabs.

One line change in the Svelte source. No prop, event, or type changes, so the React and Angular wrappers did not need updates.

## Steps needed to test

1. Run the React playground: `npm run serve:prs:react`
2. Go to `/bugs/3665`
3. Scroll down so the framework tabs are partially in view
4. Switch between React / Angular / Web Components
5. The page should stay in place (no jump). The tab content should still update.
6. Repeat in the Angular playground: `npm run serve:prs:angular`, then `/bugs/3665`

Fixes #3665

### DIFF
--- a/libs/web-components/src/components/tabs/tabs.spec.ts
+++ b/libs/web-components/src/components/tabs/tabs.spec.ts
@@ -109,6 +109,24 @@ describe("Tabs", () => {
     });
   });
 
+  it("should focus the selected tab with preventScroll to avoid page jump", async () => {
+    const { container } = render(Tabs);
+
+    const tab2Link = (await waitFor(() => {
+      const link = container.querySelector("a#tab-2") as HTMLElement;
+      expect(link).toBeTruthy();
+      return link;
+    })) as HTMLElement;
+
+    const focusSpy = vi.spyOn(tab2Link, "focus");
+
+    await fireEvent.click(tab2Link);
+
+    await waitFor(() => {
+      expect(focusSpy).toHaveBeenCalledWith({ preventScroll: true });
+    });
+  });
+
   it("should select the last tab if the tab exceeds the number of tabs", async () => {
     const result = render(Tabs, { initialtab: 3 });
 
--- a/apps/prs/angular/src/routes/bugs/3665/bug3665.component.html
+++ b/apps/prs/angular/src/routes/bugs/3665/bug3665.component.html
@@ -1,0 +1,30 @@
+<goab-text tag="h1" mt="m" mb="m">Bug 3665 - Tabs: page jumps when switching tabs</goab-text>
+<goab-text mb="l">
+  Switching tabs should not move the viewport. Previously setCurrentTab() called el.focus()
+  on the selected tab, which triggered the browser's scroll-into-view behavior and jumped the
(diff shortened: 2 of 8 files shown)


Spark450:
You may have to run locally to test right now since the previews are not working.

libs/web-components/src/components/tabs/tabs.spec.ts
```diff
@@ -109,6 +109,24 @@ describe("Tabs", () => {
     });
```
ArakTaiRoth, line 109: Testing should be expanded. This test accurately reports on preventScroll being called. But a proper browser test in `tabs.browser.spec.tsx` would be optimal to test this exact case. Build it with tall content both above and below the actual tabs, scroll so the tabs are just in view, then click a tab and verify that `window.scrollY` is unchanged.

Spark450: Good call. Added a browser test in `tabs.browser.spec.tsx` (`bug-3665`) that does exactly this: tall content above and below, scroll the tabs just into view, click a tab, and assert `window.scrollY` is unchanged. I also confirmed it is meaningful by reverting the fix and rebuilding: it fails in both Chromium and Firefox (scrollY jumps ~482px) and passes with `preventScroll` in place. Kept the original unit test too since it pinpoints the `preventScroll` call directly.

ArakTaiRoth (approved):

twjeffery (approved):
Looks great to me, thanks @Spark450 

I tested this in our website because that was where I first found that bug was the examples tab would auto scroll down to the middle of the examples for another tab, which was really strange. Looks to all be fixed now with your changes.

https://github.com/user-attachments/assets/2e7d89b1-252d-4aa7-9f61-0f08f6d22188

PR #4060, opened 2026-06-15
fix(#3741): make pagination example reproducible in Angular

### LINKED ISSUE #3741: Docs: Pagination example not reproducible
# Info

For the Pagination component, the only example it shows, it appears like this in documentation:

Image

But when reproduced in a larger viewport, exactly with the code given, it looks like this:

Image

The Pagination part of the component is not on the right side

## Acceptance Criteria:

1. Should be able to reproduce the Pagination example exactly as visually displayed

Spark450:
Fixes #3741

## What and why

The Pagination documentation example ("Show number of results per page") could not be reproduced as shown. The example uses a spacer with `hSpacing="fill"` to push the pagination controls to the right edge of the row. In Angular the controls instead bunched on the left.

### Root cause

React and web components render `<goa-spacer>` directly, so the spacer (which sets `width: 100%` on its `:host` for the "fill" value) is the direct flex child of the row and fills the free space.

The Angular wrapper renders an extra `<goab-spacer>` host element around `<goa-spacer>`. That host becomes the actual flex child, collapses to 0 width, and the inner `goa-spacer`'s `width: 100%` then resolves against a 0 width parent. The "fill" never takes effect.

### Fix

Add `:host { display: contents }` to the Angular `goab-spacer` wrapper. The host becomes transparent to layout, so `<goa-spacer>` participates directly in the parent flex container, exactly like React and web components. This mirrors the existing `goab-badge` wrapper, which already uses the same approach.

This is an Angular wrapper only issue. React and web components were already correct. There is no prop or type change, so no cross framework wrapper updates were required.

### Also included

The example (all three frameworks) now uses the compact dropdown (`size="compact"`) for the per page selector, which suits the dense inline layout.

## Before / After

Before: in Angular, "Show 10 per page" and the pagination controls sat together on the left (see the second screenshot in #3741).

After: "Show 10 per page" stays on the left and "Page X of N / Previous / Next" is pushed flush to the right edge, matching the documented example at all viewport widths.

Verified in the Angular playground at 1280px and 1500px: spacer fills (about 426px), pagination is flush right, dropdown renders compact. Verified the docs React example still renders the compact dropdown and right aligned pagination.

## Testing notes

No unit test was added. The fix depends on `display: contents` flex behaviour, which jsdom (the Angular Jest environment) does not compute, so a unit assertion would not exercise it. It is covered by the playground page instead, consistent with the untested `goab-badge` precedent. The existing `spacer.spec.ts` still passes.

## Steps to test

1. `npm run serve:prs:angular` and open `/bugs/3741`.
2. Confirm "Show 10 per page" is on the left and the pagination controls are flush to the right edge.
3. Resize the viewport wider and confirm the pagination stays pinned to the right.
4. `npm run serve:prs:react` and open `/bugs/3741` to confirm the React control page (already correct) matches.

## Checklist

- [x] Read the setup steps
- [x] Tested in React AND Angular
- [ ] Unit tests created (see Testing notes for why not)

### DIFF
--- a/apps/prs/angular/src/routes/bugs/3741/bug3741.component.html
+++ b/apps/prs/angular/src/routes/bugs/3741/bug3741.component.html
@@ -1,3 +1,11 @@
+<goab-text tag="h1" mt="m" mb="s">Bug 3741 - Pagination example not reproducible</goab-text>
+<goab-text mb="l">
+  The "Show number of results per page" example uses a spacer with
+  hSpacing="fill" to push the pagination controls to the right edge. In Angular
+  the spacer collapsed to zero width and the controls bunched on the left. After
+  the fix the spacer fills and the pagination sits flush right.
+</goab-text>
+
 <goab-table width="100%" mb="xl">
   <thead>
     <tr>
@@ -7,12 +15,12 @@
     </tr>
   </thead>
   <tbody>
-    @for (user of pageUsers; track $index) {
-    <tr>
-      <td>{{ user.firstName }}</td>
-      <td>{{ user.lastName }}</td>
-      <td>{{ user.age }}</td>
-    </tr>
+    @for (user of pageUsers; track user.id) {
+      <tr>
+        <td>{{ user.firstName }}</td>
+        <td>{{ user.lastName }}</td>
+        <td>{{ user.age }}</td>
+      </tr>
     }
   </tbody>
 </goab-table>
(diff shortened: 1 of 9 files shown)


willcodeforcoffee (approved):
Switching branches and looking on `http://localhost:4200/docs/pagination` I was able to replicate the bug.
I had a small concern because on that page (and _only_) that page the page selector dropdown is not `compact`:
image
Since it is only on that page (and not on the docs site, like http://localhost:4203/components/pagination#angular I assume there is some one-off issue? No further concerns. ✅

bdfranck (approved):
- ✅ I see the spacer works in the Angular pagination example

Image

- ✅ I see the Angular, React, and web examples use a compact dropdown

Looks good to me! 👍 

@willcodeforcoffee I added a small commit that fixes the PR Docs issue that you noticed.

Spark450:
Thanks for the review and assist @willcodeforcoffee and @bdfranck

PR #4068, opened 2026-06-16
chore: Updating release build to use built in GITHUB_TOKEN

ArakTaiRoth:
This is to change our token usage to instead use the built in GITHUB_TOKEN. This token gets created every job and only exists as long as the job is running, and is then deleted. The token inherently only has create and read permissions on the repo it's running from, and if run from a fork only has read permissions, which should be enough I think.

I don't know if this will fix the issue, but it's worth a shot and gets us away from using yet another personal access token.

### DIFF
--- a/.github/workflows/release-ci.yml
+++ b/.github/workflows/release-ci.yml
@@ -48,7 +48,7 @@ jobs:
 
       - name: Release libs
         env:
-          GITHUB_TOKEN: ${{ secrets.RELEASE_TOKEN }}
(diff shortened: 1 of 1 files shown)


bdfranck (approved):
I looked at the changes...

- ✅ I see the token is changed to `GITHUB_TOKEN`
- ✅ The usage aligns with [GitHub's documentation](https://docs.github.com/en/actions/tutorials/authenticate-with-github_token)

Looks good to me! 👍

vanessatran-ddi (approved):

PR #4074, opened 2026-06-18
Docs nav alternate: land on section pages and route "All" to the home page

twjeffery:
## Summary

This is an alternate version of the docs side navigation, put up for comparison against current production. Use this PR's preview link as "version B" and production as "version A". It is not necessarily meant to be merged as is.

It addresses a mismatch between how the side menu looks and how it behaves: the top sections look clickable but only expand an in menu view, and the "All" item looks like a back/navigation control but only collapses the menu.

## What changes

**1. Sections navigate to their page**

Clicking a top level section now lands on that section's index page instead of only swapping the in menu view:

- Components goes to `/components` (All Components)
- Get started goes to `/get-started`
- Foundations goes to `/foundations`

The destination page still shows the matching submenu (it is derived from the URL). Tokens and Examples are unchanged.

**2. "All" goes back to the home page**

The "All" item at the top of each submenu now navigates to the home page instead of silently collapsing back to the parent menu. The home page's hero search field is front and center, so search stays emphasized without anything popping open.

## How it works

- Section items are plain links. The destination page's `SiteNav` derives which submenu to show from the URL, so no in menu click state is needed.
- "All" navigates to the home page (`/`). The shared "All" item is extracted into `AllHomeItem`, and the now unused in menu state handlers (`onSelectSection`, `onBack`) were removed.

## Before / After

Before: Components, Get started, and Foundations expand a secondary in menu view without navigating. "All" (arrow back icon) collapses the submenu with no page change.

After: those sections navigate to their index page. "All" navigates back to the home page.

Compare the preview link against production for the visual before/after.

## Steps to test

1. Open the preview link.
2. From the home page, click Components in the side menu. You should land on All Components with the Components submenu shown.
3. Repeat for Get started and Foundations.
4. In any submenu, click "All" (the arrow back item at the top). You should land on the home page, with the hero search field front and center. No modal should pop open.
5. Confirm Tokens and Examples still navigate as before.
6. On mobile width, open the menu with the hamburger, click a section, confirm it navigates, then click "All" and confirm it returns home.

## Notes

- Draft on purpose. This is an A vs B comparison, not a request to merge. The session opener will verify the diff and decide whether to mark it ready.
- An earlier revision auto opened the global search modal on "All". That was removed as too intrusive: "All" now just returns home, where search is already prominent.
- No issue number attached (exploratory navigation change).
- Type checked with `astro check`. The changed files introduce no new type errors. Remaining check errors are pre existing artifacts of the unbuilt web components dist, on lines not touched here.

https://claude.ai/code/session_014KqafCHdPSm7GsgVWwAVHN

### DIFF
--- a/docs/src/components/SiteNav.tsx
+++ b/docs/src/components/SiteNav.tsx
@@ -1,16 +1,14 @@
 /**
  * SiteNav.tsx
  *
- * Unified site navigation with parent/sub-menu state switching.
+ * Unified site navigation. Every top-level section links directly to its index
+ * page; which menu renders (parent vs. a submenu) is derived from the current
+ * URL, not from in-menu click state.
  *
- * Navigation behavior:
- * - Components, Get started, Foundations: has submenu (many pages)
- * - Tokens: Direct link, stays on parent menu, highlights "Tokens"
- * - Examples: Direct link, stays on parent menu, highlights "Examples"
- *
- * Menu level is managed via React state:
- * - Parent view: Shows all sections
- * - Sub-menu view: Shows section-specific navigation (components, get-started, foundations)
+ * - Components, Get started, Foundations: land on their index page, which shows
+ *   the matching submenu.
+ * - Tokens, Examples: single pages, stay on the parent menu and highlight.
+ * - A submenu's "All" item navigates home and opens global search.
  */
 
 import { useState, useEffect, useCallback } from "react";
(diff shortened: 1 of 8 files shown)


twjeffery:
Closing this one. It was a side by side to compare against production, not a change meant to land. The description stays here in case the section navigation idea gets picked up later.

PR #4079, opened 2026-06-19
fix(#3964): mark Angular input/textarea as touched on blur only

### LINKED ISSUE #3964: Input being marked as touch (Angular) when user tabs into it
# Info

Reported by Walter and Alejandro

If a person is using `.touched` on a Reactive FormControl for validation, this will trigger as soon as a user tabs into the Input component. In the Angular wrapper, we trigger `markAsTouched` on `onFocus`, when it should be triggered on `onBlur` instead. `markAsTouched` should also be removed from `onChange` and `onKeyPress` because the strict definition of touched is "user entered/interacted and then left". This should also all be changed in the TextArea component.

## Code to Replicate

```html
<form [formGroup]="form">
  <goab-form-item label="Last name" [error]="lastNameError">
    <goab-input
      formControlName="lastName"
      name="lastName"
      width="100%"
      [error]="!!lastNameError"
    />
  </goab-form-item>
</form>
```

```typescript
form: FormGroup = this.fb.group({
  lastName: ["", Validators.required],
});

get lastNameError() {
  const lastName = this.form.get("lastName") as FormControl;
  return !lastName.valid && lastName.touched && lastName.hasError("required")
    ? "Enter a last name."
    : undefined;
}
```

## Workaround

There is a workaround available that does work. But it's a bit of a pain to implement:

```html
<form [formGroup]="form">
  <goab-form-item label="Last name" [error]="getError('lastName')">
    <goab-input
      formControlName="lastName"
      name="lastName"
      width="100%"
      [error]="getError('lastName')"
      (onBlur)="onBlur('lastName')"
    />
  </goab-form-item>
</form>
```

```typescript
blurred: Record<string, boolean> = {};

form: FormGroup = this.fb.group({
  lastName: ["", Validators.required],
});

onBlur(controlName: string): void {
  this.blurred[controlName] = true;
}

getError(controlName: string) {
  const lastName = this.form.get(controlName) as FormControl;
  return !lastName.valid && this.blurred[controlName] && lastName.hasError("required")
    ? "Enter a last name."
    : undefined;
}
```

The above can be expanded to work with any number of FormControls

## Acceptance Criteria:

1. Input Angular wrapper is changed to remove `markAsTouched` from `onChange`, `onKeyPress`, and `onFocus`, and instead added to run `onBlur`.
2. Textarea Angular wrapper is changed to remove `markAsTouched` from `onChange`, and `onKeyPress`

Spark450:
# Before (the change)

Using `.touched` on a reactive `FormControl` for validation triggered as soon as a user tabbed into a `goab-input` (or `goab-textarea`). The Angular wrappers called `markAsTouched` on focus, change, and keypress, so the error state appeared before the user had interacted and left the field.

# After (the change)

`markAsTouched` runs only on blur, matching the definition of touched: entered and then left.

- Input wrapper: removed `markAsTouched` from `_onChange`, `_onKeyPress`, and `_onFocus` (also removed the redundant `fcTouched` call in `_onKeyPress`); added it to `_onBlur`.
- Textarea wrapper: removed `markAsTouched` from `_onChange` and `_onKeyPress`; the existing `_onBlur` call remains.

This is an Angular wrapper only change. `markAsTouched` is part of the Angular `ControlValueAccessor` integration, so the Svelte source and React wrapper are unaffected.

## Steps needed to test

1. Run the Angular playground and open the `3964` bug page.
2. Tab into the Last name field and tab straight out without typing. The required error appears, because the control was entered and then left.
3. Tab into the field and type without leaving. No error appears while focused.
4. Repeat for the Bio textarea.
5. Confirm `lastName.touched` / `bio.touched` stay `false` on focus, change, and keypress, and flip to `true` on blur.

Fixes #3964

### DIFF
--- a/apps/prs/angular/src/routes/bugs/3964/bug3964.component.html
+++ b/apps/prs/angular/src/routes/bugs/3964/bug3964.component.html
@@ -1,0 +1,71 @@
+<h1>3964 - Input marked as touched when user tabs in (Angular)</h1>
+<p>
+  Tab into a field and tab straight back out without typing: the required error should
+  appear, because the control was entered and then left. Tabbing in alone, or typing
+  without leaving, should NOT show the error. Before the fix, the error appeared as soon
+  as the field received focus.
+</p>
+
+<h2>Reactive form (FormControl)</h2>
+<form [formGroup]="form">
+  <goab-form-item label="Last name" [error]="lastNameError">
+    <goab-input
+      formControlName="lastName"
+      name="lastName"
+      width="100%"
+      [error]="!!lastNameError"
+    />
+  </goab-form-item>
+
+  <goab-form-item label="Bio" [error]="bioError">
+    <goab-textarea formControlName="bio" name="bio" width="100%" [error]="!!bioError" />
+  </goab-form-item>
+</form>
+
+<dl>
+  <dt>lastName.touched</dt>
+  <dd>{{ form.get("lastName")?.touched }}</dd>
+  <dt>bio.touched</dt>
+  <dd>{{ form.get("bio")?.touched }}</dd>
+</dl>
+
+<h2>Template-driven form (ngModel)</h2>
+<p>Same behavior should hold for the ngModel binding.</p>
+<form>
+  <goab-form-item
+    label="Last name"
+    [error]="
+      tdLastName.touched && tdLastName.invalid ? 'Enter a last name.' : undefined
+    "
+  >
+    <goab-input
+      name="ngModelLastName"
+      width="100%"
+      required
+      [(ngModel)]="ngModelLastName"
+      #tdLastName="ngModel"
+      [error]="tdLastName.touched && tdLastName.invalid"
+    />
+  </goab-form-item>
+
+  <goab-form-item
+    label="Bio"
+    [error]="tdBio.touched && tdBio.invalid ? 'Enter a bio.' : undefined"
+  >
+    <goab-textarea
+      name="ngModelBio"
+      width="100%"
+      required
+      [(ngModel)]="ngModelBio"
+      #tdBio="ngModel"
+      [error]="tdBio.touched && tdBio.invalid"
+    />
+  </goab-form-item>
+</form>
+
+<dl>
+  <dt>ngModelLastName.touched</dt>
+  <dd>{{ tdLastName.touched }}</dd>
+  <dt>ngModelBio.touched</dt>
+  <dd>{{ tdBio.touched }}</dd>
+</dl>
--- a/libs/angular-components/src/lib/components/input/input.spec.ts
+++ b/libs/angular-components/src/lib/components/input/input.spec.ts
@@ -1,6 +1,6 @@
 import { ComponentFixture, TestBed, fakeAsync, tick } from "@angular/core/testing";
 import { GoabInput } from "./input";
-import { Component, CUSTOM_ELEMENTS_SCHEMA, TemplateRef } from "@angular/core";
+import { Component, CUSTOM_ELEMENTS_SCHEMA, TemplateRef, ViewChild } from "@angular/core";
 import {
   GoabIconType,
   GoabInputAutoCapitalize,
@@ -10,6 +10,13 @@ import {
 } from "@abgov/ui-components-common";
 import { By } from "@angular/platform-browser";
 import { fireEvent } from "@testing-library/dom";
+import {
+  FormControl,
+  FormGroup,
+  FormsModule,
+  NgModel,
+  ReactiveFormsModule,
+} from "@angular/forms";
 
 @Component({
   standalone: true,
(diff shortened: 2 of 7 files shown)


libs/angular-components/src/lib/components/input/input.ts
```diff
@@ -199,28 +199,24 @@ export class GoabInput extends GoabControlValueAccessor implements OnInit {
...
 
     this.fcChange?.(detail.value);
   }
 
   _onKeyPress(e: Event) {
-    this.markAsTouched();
     const detail = { ...(e as CustomEvent<GoabInputOnKeyPressDetail>).detail, event: e };
     this.onKeyPress.emit(detail);
-
-    this.fcTouched?.();
```
Spark450, line 214: @vanessatran-ddi not sure if this should have been removed?

libs/angular-components/src/lib/components/input/input.spec.ts
```diff
@@ -326,6 +327,65 @@ describe("GoABInput", () => {
...
+}
+
+describe("GoabInput reactive form touched state", () => {
+  let fixture: ComponentFixture<TestReactiveFormComponent>;
+  let component: TestReactiveFormComponent;
+  let input: HTMLElement;
+
+  beforeEach(fakeAsync(() => {
+    TestBed.configureTestingModule({
+      imports: [TestReactiveFormComponent, GoabInput],
```
vanessatran-ddi: We can remove `GoabInput` from this `imports` because we already import inside `TestReactiveFormComponent` line 352

libs/angular-components/src/lib/components/textarea/textarea.spec.ts
```diff
@@ -180,3 +181,57 @@ describe("GoABTextArea", () => {
...
+}
+
+describe("GoabTextArea reactive form touched state", () => {
+  let fixture: ComponentFixture<TestReactiveFormComponent>;
+  let component: TestReactiveFormComponent;
+  let el: HTMLElement;
+
+  beforeEach(fakeAsync(() => {
+    TestBed.configureTestingModule({
+      imports: [TestReactiveFormComponent, GoabTextArea],
```
vanessatran-ddi: Same here, we can remove `GoabTextArea`, it was being imported line 187

apps/prs/react/src/routes/bugs/bug3964.tsx
```diff
@@ -0,0 +1,27 @@
+import { GoabFormItem, GoabInput, GoabTextArea, GoabText } from "@abgov/react-components";
+
+export function Bug3964Route() {
```
vanessatran-ddi: We can remove added file under apps/prs/react since this change is for Angular only

apps/prs/angular/src/routes/bugs/3964/bug3964.component.html
```diff
@@ -0,0 +1,29 @@
+<h1>3964 - Input marked as touched when user tabs in (Angular)</h1>
+<p>
+  Tab into a field and tab straight back out without typing: the required error should
+  appear, because the control was entered and then left. Tabbing in alone, or typing
+  without leaving, should NOT show the error. Before the fix, the error appeared as soon
+  as the field received focus.
+</p>
+
+<form [formGroup]="form">
```
vanessatran-ddi, line 10: Can we add another example to test the ngModel template for both `input` and `textarea`? Just to make sure it does not affect ngModel way: 
image

libs/angular-components/src/lib/components/input/input.ts
vanessatran-ddi: Yes this should be removed.

libs/angular-components/src/lib/components/input/input.spec.ts
Spark450: Done, removed `GoabInput` from the TestBed imports in that block since the host component already imports it.

libs/angular-components/src/lib/components/textarea/textarea.spec.ts
Spark450: Done, removed `GoabTextArea` from the TestBed imports here too.

apps/prs/react/src/routes/bugs/bug3964.tsx
Spark450: Removed the React playground page and its route file since the fix is Angular only.

apps/prs/angular/src/routes/bugs/3964/bug3964.component.html
Spark450: Added template-driven (ngModel) examples for both input and textarea to the playground page, and also added unit tests asserting the touched behavior via ngModel for both components. Confirmed it does not affect the ngModel path.

libs/angular-components/src/lib/components/input/input.ts
Spark450: Confirmed, leaving this removed. The redundant fcTouched call was the same touch trigger as markAsTouched, which now runs on blur only.

libs/angular-components/src/lib/components/input/input.spec.ts
```diff
@@ -10,6 +10,14 @@ import {
...
 import { By } from "@angular/platform-browser";
 import { fireEvent } from "@testing-library/dom";
+import {
+  FormControl,
+  FormGroup,
+  FormsModule,
+  NgModel,
+  ReactiveFormsModule,
+} from "@angular/forms";
+import { ViewChild } from "@angular/core";
```
ArakTaiRoth: This should be imported on line 3, where we already import other stuff from `@angular/core`

libs/angular-components/src/lib/components/textarea/textarea.spec.ts
```diff
@@ -4,6 +4,14 @@ import { Component, CUSTOM_ELEMENTS_SCHEMA } from "@angular/core";
...
 import { fireEvent } from "@testing-library/dom";
 import { By } from "@angular/platform-browser";
+import {
+  FormControl,
+  FormGroup,
+  FormsModule,
+  NgModel,
+  ReactiveFormsModule,
+} from "@angular/forms";
+import { ViewChild } from "@angular/core";
```
ArakTaiRoth: Same thing here

vanessatran-ddi (approved):

libs/angular-components/src/lib/components/input/input.spec.ts
Spark450: Done, consolidated `ViewChild` into the existing `@angular/core` import.

libs/angular-components/src/lib/components/textarea/textarea.spec.ts
Spark450: Done, same fix here.

ArakTaiRoth (approved):

PR #4081, opened 2026-06-19
fix(#4027): hide number spinner on DatePicker day and year inputs

### LINKED ISSUE #4027: DatePicker: Day and Year should be typed text fields, not number inputs
## Current behavior

The Day (DD) and Year (YYYY) fields in the input-type DatePicker use `type="number"` (`DatePicker.svelte:372` and `:388`), so the browser draws the native up/down arrows on them.

Image

## Expected behavior

Day and Year should be plain typed fields: digits only, a numeric keypad on mobile, and no up/down arrows. The way to keep the digit behavior without the arrows is a text field with a numeric input mode (`inputmode="numeric"`) instead of `type="number"`.

## Why

Those up/down arrows are easy to miss, fiddly to use, and meaningless for a date. Nobody nudges a birth year up one click at a time, they type it. `type="number"` also brings real problems:

- The value changes if you accidentally scroll over the field.
- Some browsers let you type letters like `e`, `+`, and `-` into it.
- The arrows look and behave differently in every browser.

## Acceptance Criteria

1. Change day and year inputs on DatePicker to be string instead of number
2. Verify that the return type doesn't change

Spark450:
Fixes #4027

# Before

The Day (DD) and Year (YYYY) fields in the input-type DatePicker used `type="number"`, so the browser drew native up/down spinner arrows. Those arrows are easy to miss and meaningless for a date, and the value changes if you accidentally scroll over the field.

# After

Day and Year stay as `type="number"` (numeric keypad on mobile, native `min`/`max` validation), but the up/down spinner arrows are now hidden via CSS. The fix is scoped to a `no-spinner` class applied only to these two DatePicker fields, so other `type="number"` inputs elsewhere (for example `GoabInputNumber`) are unaffected and keep their spinner buttons, since incrementing by 1 can make more sense in those contexts. The emitted value is unchanged.

Notes:
- An earlier version of this PR moved Day/Year to `type="text"` with `inputmode="numeric"` and added a new `inputmode` prop to `Input`/`GoabInput`/`goab-input`. After discussion with the team we reverted that approach in favor of the CSS-only fix above, since it keeps native number validation and avoids a new prop with broader surface area than this fix needs.
- Added a comparison example to the bugs/4027 playground page (a plain `GoabInputNumber`) so a reviewer can confirm the spinner buttons are still present elsewhere.

## Steps needed to test

1. Run the React playground: `npm run serve:prs:react` and open `bugs/4027`.
2. Confirm the Day and Year fields have no up/down arrows.
3. Confirm the comparison number input below still shows its up/down arrows.
4. On a mobile device (or emulator), confirm focusing Day/Year shows the numeric keypad.
5. Enter a full date and confirm the selected value updates correctly.
6. Repeat in the Angular playground: `npm run serve:prs:angular`, open `bugs/4027`.

### DIFF
--- a/libs/web-components/src/components/date-picker/DatePicker.svelte
+++ b/libs/web-components/src/components/date-picker/DatePicker.svelte
@@ -379,9 +379,10 @@
         <goa-input
           name="day"
           type="number"
+          class="no-spinner"
           testid="input-day"
           on:_change={onInputChange}
-          width="4ch"
+          width="2ch"
           value={_date.day || ""}
           min="1"
           max="31"
@@ -395,9 +396,10 @@
         <goa-input
           name="year"
           type="number"
+          class="no-spinner"
           testid="input-year"
           on:_change={onInputChange}
-          width="6ch"
+          width="4ch"
           value={_date.year || ""}
           min="1800"
           max="2200"
--- a/apps/prs/angular/src/routes/bugs/4027/bug4027.component.html
+++ b/apps/prs/angular/src/routes/bugs/4027/bug4027.component.html
@@ -1,0 +1,22 @@
+<goab-text tag="h1" mt="m" mb="s">Bug 4027 - DatePicker: day and year hide number spinner</goab-text>
+<goab-text mb="l">
+  The Day (DD) and Year (YYYY) fields in the input-type DatePicker stay as type="number"
+  (numeric keypad on mobile, native validation), but the up/down spinner arrows are now
+  hidden via CSS. Other type="number" inputs elsewhere keep their spinner buttons. The
+  emitted value is unchanged.
+</goab-text>
+
+<goab-date-picker name="dob" type="input" [value]="value" (onChange)="onChange($event)" />
+
+<goab-text mt="l">Selected value: {{ value || "(none)" }}</goab-text>
+
+<goab-text tag="h2" mt="2xl" mb="m">Comparison: plain type="number" input</goab-text>
+<goab-text mb="l">
+  This is an unrelated GoabInputNumber. It should still show its up/down spinner arrows,
+  confirming the fix above is scoped to DatePicker's Day and Year fields only.
+</goab-text>
+<goab-input-number
+  name="comparison-number"
+  [value]="comparisonValue"
+  (onChange)="onComparisonChange($event)"
+/>
--- a/apps/prs/angular/src/routes/bugs/4027/bug4027.component.ts
+++ b/apps/prs/angular/src/routes/bugs/4027/bug4027.component.ts
@@ -1,0 +1,25 @@
+import { Component } from "@angular/core";
+import { GoabDatePicker, GoabInputNumber, GoabText } from "@abgov/angular-components";
+import {
+  GoabDatePickerOnChangeDetail,
+  GoabInputOnChangeDetail,
+} from "@abgov/ui-components-common";
+
+@Component({
+  standalone: true,
+  selector: "abgov-bug4027",
+  templateUrl: "./bug4027.component.html",
+  imports: [GoabDatePicker, GoabInputNumber, GoabText],
+})
+export class Bug4027Component {
+  value = "2025-06-15";
+  comparisonValue = 42;
+
+  onChange(detail: GoabDatePickerOnChangeDetail) {
+    this.value = detail.valueStr;
+  }
+
+  onComparisonChange(detail: GoabInputOnChangeDetail) {
+    this.comparisonValue = parseFloat(detail.value);
+  }
+}
--- a/apps/prs/angular/src/routes/bugs/4027/bug4027.route.json
+++ b/apps/prs/angular/src/routes/bugs/4027/bug4027.route.json
@@ -1,0 +1,6 @@
+{
+  "type": "bug",
+  "id": "4027",
+  "path": "bugs/4027",
+  "title": "DatePicker day/year number spinner"
+}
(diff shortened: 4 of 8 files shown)


Spark450:
A few notes for reviewers:

1. **Older iOS Safari and the numeric keypad.** Pre-iOS-12.2 Safari ignored `inputmode` and only honored `pattern="[0-9]*"` for the numeric keypad. We don't support iOS versions that old, so I did not add a `pattern` attribute. Flagging in case anyone disagrees on the support baseline.

2. **`inputmode` not added to `Input.html-data.json`.** That file is already missing some newer props (`textalign`, `size`), so I left it consistent rather than partially updating it. Happy to add an entry if we'd prefer to start closing that gap.

3. **DatePicker "emits a string date value" test tier.** This one verifies AC #2 (return type unchanged) via a synthetic `CustomEvent` dispatch in a unit test rather than a browser test, because the internal `goa-input` doesn't render a real input in jsdom. Calling it out since event behavior usually lives in browser tests. Open to moving it if preferred.

libs/web-components/src/components/date-picker/DatePicker.svelte
```diff
@@ -394,10 +395,11 @@
       <goa-form-item helptext="Year (YYYY)" {version}>
         <goa-input
           name="year"
-          type="number"
+          type="text"
```
ArakTaiRoth: There is a problem with this. By setting it to `text` and then setting `inputmode` to numeric, you do successfully remove the up/down arrows on desktop and allow mobile devices to use the proper inputs. However, this now allows things like letters to be typed into the text box, which would be invalid values. People will 100% report this as an issue.

To properly do this we would also need to introduce a `pattern` property on the `GoabInput` component that can be used to supply a regex on what's allowed in the input.

docs/generated/component-apis/input.json
```diff
@@ -666,6 +680,13 @@
           "default": "",
           "description": "Unique identifier for the input element. Used for label associations and accessibility."
         },
+        {
+          "name": "inputmode",
+          "type": "\"\"",
```
ArakTaiRoth: This is an incorrect type. Look at line 799 in this same file for an accurate way to set the `type` property

libs/web-components/src/components/date-picker/DatePicker.svelte
```diff
@@ -394,10 +395,11 @@
...
+          inputmode="numeric"
           testid="input-year"
           on:_change={onInputChange}
-          width="6ch"
+          width="4ch"
           value={_date.year || ""}
           min="1800"
```
ArakTaiRoth, line 404: If we change the `type` to text, we can't use the properties `min` or `max`. Which also means we lose the native HTML5 validation provided by these properties.

ArakTaiRoth:
@Spark450 Do we maybe need to discuss this a little more and what we actually need for the problem being solved? This seems like overkill for the actual problem that needs to be solved as written. The actual problem as written can be solved with a little bit of CSS, but that does have potentially other issues, hence the need to talk about this.

Spark450:
## Update: switched to a CSS-only fix

After discussing with the team, we're reverting the text-input/`inputmode` approach and going with a CSS-only fix instead.

**What changed:**
- Day (DD) and Year (YYYY) are back to `type="number"`, so they keep the native numeric mobile keypad and native `min`/`max` validation.
- The up/down spinner arrows are now hidden via CSS, scoped to a `no-spinner` class applied only to these two DatePicker fields. Other `type="number"` inputs elsewhere (for example `GoabInputNumber`) are unaffected and keep their spinner buttons, since incrementing by 1 can make more sense in those contexts.
- Removed the `inputmode` prop that the earlier version of this PR added to `Input`/`GoabInput`/`goab-input` (Svelte, React, Angular, and `libs/common`). It's no longer needed, so it's fully reverted, not deprecated.
- Updated the `date-picker.spec.ts` and `Input.spec.ts` unit tests, and the React/Angular playground pages (`bugs/4027`), to match.
- Regenerated `docs/generated/component-apis/input.json` since the `inputmode` prop is gone.

This also resolves the earlier review feedback: typing letters into Day/Year is no longer possible since the field is still `type="number"`, and `min`/`max` keep their native browser validation since we never moved off `type="number"`.

Updated the PR title and description above to match this approach.

**Edit:** Also added a comparison example (a plain `GoabInputNumber`) to the `bugs/4027` playground page so you can confirm the spinner buttons are still present on unrelated number inputs. While adding it, noticed the native spinner sits flush against the value text on `type="number"` inputs in general, unrelated to this fix. That's pre-existing and out of scope here (it'd need a spacing token change in the separate design-tokens repo), so it's tracked separately in #4089.

libs/web-components/src/components/date-picker/DatePicker.svelte
Spark450: As agreed type remains "number"

libs/web-components/src/components/input/Input.spec.ts
```diff
@@ -165,6 +165,7 @@ describe("GoAInput Component", () => {
       const root = el.container.querySelector("[autocomplete=off]");
       expect(root).toBeTruthy();
     });
+
```
ArakTaiRoth: This modified line can be removed so this file doesn't get an edit

Spark450: Removed, thanks.

ArakTaiRoth (approved):

willcodeforcoffee (approved):
Man I wish the `min` / `max` attributes worked in Chrome! It would make changes like this _so_ much cleaner.

PR #4083, opened 2026-06-19
fix(#4082): wait on real tooltip state in work side menu browser tests

### LINKED ISSUE #4082: Work side menu: tooltip browser tests are flaky due to fake-timer/rAF mismatch
## Description

The work side menu tooltip browser tests (`libs/react-components/specs/work-side-menu.browser.spec.tsx`) used `vi.useFakeTimers()` with a fixed `vi.advanceTimersByTimeAsync(400)` before asserting `tooltipEl?.classList.contains("show")`. The `.show` class is applied on a rAF/transition tick, which fake timers don't advance deterministically, so the assertion intermittently fails right after the fixed advance. A robust fix waits on the actual condition (e.g. via `vi.waitFor` against real timers) instead of a fixed timer duration.

The same tests also had a `toBeTruthy()` truthy-assertion-on-a-locator anti-pattern (`expect(tooltipEl).toBeTruthy()`), which our testing standards flag as meaningless since `getBy*`/query results are objects and the assertion doesn't test actual state.

## Steps to Reproduce

1. Run `npx vitest run libs/react-components/specs/work-side-menu.browser.spec.tsx` repeatedly (or under load/CI).
2. Observe intermittent failures on `expect(tooltipEl?.classList.contains("show")).toBe(true)` immediately after `advanceTimersByTimeAsync(400)`.

## Acceptance Criteria

- [ ] Tooltip show/hide browser tests wait on the actual `.show` class state (e.g. `vi.waitFor`) rather than a fixed fake-timer advance.
- [ ] No `toBeTruthy()` assertions on locators/query results in these tests.
- [ ] Tests pass reliably across repeated/parallel runs.

Spark450:
# Before (the change)

The work side menu tooltip browser tests (`libs/react-components/specs/work-side-menu.browser.spec.tsx`) used `vi.useFakeTimers()` with a fixed `vi.advanceTimersByTimeAsync(400)` before asserting `tooltipEl?.classList.contains("show")`.

The `.show` class is applied by a real `setTimeout` that runs only after the browser delivers a real `mouseenter` from the Playwright `hover()` interaction. Because the real event is not governed by fake timers, advancing fake time by a fixed amount raced the real event: when the mouseenter had not yet been delivered, the 300ms timer was never registered, so advancing fake time fired nothing and `.show` was never applied. The assertion then failed intermittently right after the fixed advance.

The same tests also used `expect(tooltipEl).toBeTruthy()`, a truthy assertion on a query result that our testing standards flag as meaningless.

# After (the change)

The three tooltip tests now drop fake timers and wait on the actual `.show` class state via `vi.waitFor` against real timers. The redundant `toBeTruthy()` assertions are removed, since asserting on `.show` already requires the element to exist.

This is a test only change. No component, prop, event, or wrapper behavior changed, so React and Angular wrappers are untouched.

## Steps needed to test

1. Check out the branch and run `npx vitest --run --project='*-headless' work-side-menu.browser -t "tooltip"`.
2. Confirm the three tooltip tests pass.
3. Run it repeatedly (or under CI load) and confirm there are no intermittent failures on `expect(tooltipEl?.classList.contains("show")).toBe(true)`.

Verified locally: 6/6 tooltip assertions pass across 3 repeat runs, and the full `work-side-menu.browser` spec passes 26/26.

Fixes #4082

### DIFF
--- a/libs/react-components/specs/work-side-menu.browser.spec.tsx
+++ b/libs/react-components/specs/work-side-menu.browser.spec.tsx
@@ -137,11 +137,11 @@ describe("WorkSideMenu", () => {
       const menuItem = result.getByTestId("hover-item");
 
       await menuItem.hover();
+      const tooltipEl = menu
+        .element()
+        .querySelector(".tooltip") as HTMLElement | null;
 
       await vi.waitFor(() => {
-        const tooltipEl = menu
-          .element()
-          .querySelector(".tooltip") as HTMLElement | null;
         expect(tooltipEl?.classList.contains("show")).toBe(true);
         expect(tooltipEl?.textContent?.trim()).toBe("Search");
         expect(tooltipEl?.style.left).not.toBe("");
@@ -151,9 +151,6 @@ describe("WorkSideMenu", () => {
       await menuItem.unhover();
 
       await vi.waitFor(() => {
-        const tooltipEl = menu
-          .element()
-          .querySelector(".tooltip") as HTMLElement | null;
         expect(tooltipEl?.classList.contains("show")).toBe(false);
       });
     });
@@ -186,11 +183,11 @@ describe("WorkSideMenu", () => {
       const group = result.getByTestId("hover-group");
 
       await group.hover();
+      const tooltipEl = menu
+        .element()
+        .querySelector(".tooltip") as HTMLElement | null;
 
       await vi.waitFor(() => {
-        const tooltipEl = menu
-          .element()
-          .querySelector(".tooltip") as HTMLElement | null;
         expect(tooltipEl?.classList.contains("show")).toBe(true);
         expect(tooltipEl?.textContent?.trim()).toBe("Applications");
         expect(tooltipEl?.style.left).not.toBe("");
@@ -200,9 +197,6 @@ describe("WorkSideMenu", () => {
       await group.unhover();
 
       await vi.waitFor(() => {
-        const tooltipEl = menu
-          .element()
-          .querySelector(".tooltip") as HTMLElement | null;
         expect(tooltipEl?.classList.contains("show")).toBe(false);
       });
     });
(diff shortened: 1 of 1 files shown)


Spark450:
@vanessatran-ddi I very quickly got in over my head here, I can leave this hanging around if you think it could be something we can use but if not, happy to close.

libs/react-components/specs/work-side-menu.browser.spec.tsx
```diff
@@ -136,35 +136,26 @@ describe("WorkSideMenu", () => {
...
-          expect(tooltipEl).toBeTruthy();
-          expect(tooltipEl?.classList.contains("show")).toBe(false);
-        });
-      } finally {
-        vi.useRealTimers();
-      }
+      await menuItem.hover();
+
+      await vi.waitFor(() => {
+        const tooltipEl = menu
```
vanessatran-ddi: This approach is correct, we should use waitFor, and remove `useFakeTimers`. Now for `waitFor` I think we can move `tooltipEl` here out of the `waitFor`. For example `const tooltipEl = ...; await vi.wairFor(() => { expect tooltipEl?....`

Spark450: Good call, done in bde1781. Moved `tooltipEl` out to a single declaration after the hover and reused it across the show and hide `waitFor` assertions, in all three tooltip tests.

Spark450: Quick summary of where this landed after your review:

- **bde1781**: addressed your feedback, `tooltipEl` is now declared once after the hover and reused across the show/hide `waitFor` blocks in all three tooltip tests.
- **c8d1362**: CI then failed again in Firefox under load, `.show` wasn't appearing within `waitFor`'s 1000ms default. Root cause: the tooltip's show/hide is gated by a 300ms debounce in `WorkSideMenu.svelte`, and on a contended CI runner that timer callback can get delayed past the 1000ms budget. Added an explicit 3000ms timeout (`TOOLTIP_WAIT`) to the six tooltip `waitFor` calls, with a comment explaining why, so it absorbs the debounce plus event latency rather than being a blind timeout bump.

Both passed locally and the latest CI run is green. Ready for another look when you have a chance.

ArakTaiRoth (approved):

ArakTaiRoth:
@Spark450 This is actually fine I think. I would leave this as is.

vanessatran-ddi (approved):
Looks good!

PR #4087, opened 2026-06-23
fix(#3815): expose date picker size prop on wrappers and restore docs sizes example

### LINKED ISSUE #3815: Date picker: add size prop
Right now `GoabDatePicker` doesn't expose the property `size`, 

Image

so PR https://github.com/GovAlta/ui-components/pull/3794 hides the example Sizes:

Image

Once we expose this property, we can add this example back to the docs.

Spark450:
# Before (the change)

`GoabDatePicker` did not expose a `size` prop, so consumers could not render the compact variant and the docs "Sizes" example was hidden. Separately, when `type="input"` was used in a narrow container, the Month/Day/Year fields overflowed instead of staying on one line.

# After (the change)

`GoabDatePicker` exposes a `size` prop ("default" | "compact") on the React and Angular wrappers, applying to both the calendar and input variants. The docs "Sizes" example is visible again. The `type="input"` layout is also responsive: the three fields stay on one line and the Month field shrinks to fit instead of overflowing.

## What to review

**`size` prop (wrappers + docs)**
- `GoabDatePickerSize` type added to `@abgov/ui-components-common`.
- `size` prop added to the React and Angular `GoabDatePicker` wrappers (the Svelte web component already supported it).
- Docs "Sizes" example re-enabled, generated date-picker API regenerated.

**Responsive input layout** (`libs/web-components/src/components/date-picker/DatePicker.svelte`, `dropdown/Dropdown.svelte`, `popover/Popover.svelte`)
- The date row (`goa-block`) is capped to `max-width: 100%` so it can shrink with its container, and the row gap tracks `size`.
- `goa-dropdown` and `goa-popover` gain an internal `shrinktarget` flag: when set, the dropdown's trigger fills its row up to its natural width and can shrink narrower than that, while the open menu keeps its full, readable width. This is opt-in and only used by the Month field here, so other dropdowns are unaffected.
- When the Month field shrinks enough that the empty placeholder ("—Select month—") would render as truncated noise, its text is hidden (`placeholdervalue` + a truncation check) rather than shown clipped. A real month selection (e.g. "September") is unaffected by this and ellipsizes normally when narrow.

**Even field spacing at compact size**
- At compact size, `goa-form-item` was sizing the Day/Year fields to fit their full helptext labels ("Day (DD)", "Year (YYYY)"), which are wider than the narrow Day/Year inputs. That left invisible slack after those fields, making the Day-Year gap look bigger than the Month-Day gap.
- Fixed by shortening the helptext to "Day"/"Year" at compact size only (default size keeps "Day (DD)"/"Year (YYYY)"). This removes the slack entirely since the shortened labels no longer exceed their input's width, so all gaps are now pixel-equal at compact size.

## Known limitation

- In Firefox, when the Month field has shrunk enough to truncate a real selection (e.g. "September" in a narrow container), the text is hard-clipped without a "…" ellipsis, unlike Chromium-based browsers. We looked at adding the standard `overflow: hidden` / `white-space: nowrap` pairing alongside `text-overflow: ellipsis` on the dropdown's input, but it didn't resolve the Firefox rendering. Decided to accept this as a known cross-browser gap rather than continue digging on this pass; the field still shrinks correctly and never overflows its container, it just won't show the ellipsis in Firefox specifically.

## Verification

- Unit tests added/passing for `size` (React, Angular wrappers) and for the responsive layout attributes (`shrinktarget`, `placeholdervalue`, `max-width`, size-driven gap, compact-only helptext) in the web component spec.
- Verified in the browser: calendar and input variants at both sizes, and the narrow-container example, one line, no overflow, month shrinks and ellipsizes (Chromium), open menu stays fully readable, real selections display correctly, compact-size field gaps are pixel-equal.
- One minor thing worth a second look during review: at the default (non-compact) size, the empty Month placeholder can render hidden even in a full-width container, not just a narrow one, likely a small margin in the dropdown's natural-width calculation at the larger font size. Doesn't affect compact size or any real selection.

## Steps needed to test

1. Run `npm run serve:prs:react` and open `bugs/3815`.
2. Confirm the calendar variant renders a default and a compact size (compact is shorter).
3. Confirm the input variant renders a default and a compact size, fields on one line, and gaps between fields look even at compact size.
4. Resize the narrow-container example (or shrink the browser) and confirm the fields stay on one line, the Month field shrinks/ellipsizes, and the menu opens at full readable width.
5. Run `npm run serve:prs:angular` and open `bugs/3815`, repeat the checks.

Fixes #3815


Spark450:
@twjeffery this is the PR that I got stuck on last week, ideas and suggestions are welcome 🙏

twjeffery:
@Spark450 I added the responsive input layout on top of your size prop work.

When `type="input"` is in a narrow container, the three fields now stay on one line and shrink to fit instead of overflowing.

https://github.com/user-attachments/assets/32e85589-6359-49a6-bc38-bce0b2f53e6e

The tricky part was the month dropdown, which couldn't get narrower than its own menu before. Now it can, but only for the date picker, so every other dropdown and popover works exactly as it did. When the month is empty it shows just the chevron, and the menu still opens at full width.

Spark450:
> @Spark450 I added the responsive input layout on top of your size prop work.
> 
> When `type="input"` is in a narrow container, the three fields now stay on one line and shrink to fit instead of overflowing.
> 
>  Screen.Recording.2026-06-30.at.7.29.33.PM.mov 
> 
> The tricky part was the month dropdown, which couldn't get narrower than its own menu before. Now it can, but only for the date picker, so every other dropdown and popover works exactly as it did. When the month is empty it shows just the chevron, and the menu still opens at full width.

@twjeffery This is great! I think not showing a truncated —Select month— in this scenario seems fine/good to me. I'll update the PR comment accordingly.

willcodeforcoffee (approved):

ArakTaiRoth:
@Spark450 @twjeffery I honestly really don't like this PR. If we ignore the "ideal scenario" put forth in the story (type "input" stays on one line on mobile). We are literally adding two more properties to Dropdown and one more property to Popover. Both components that we struggle to maintain properly to this day and get increasingly complicated everytime we so much as look at them. And we add these properties just to support this one very specific scenario.

I would propose a much simpler fix that is an additional `<div>` if the `input` type is used, and some additional CSS. All added to DatePicker, the component we're trying to fix, with no changes being made to Dropdown and to Popover. The fix would look like this, which I think is perfectly acceptable:
image

Spark450:
> @Spark450 @twjeffery I honestly really don't like this PR. If we ignore the "ideal scenario" put forth in the story (type "input" stays on one line on mobile). We are literally adding two more properties to Dropdown and one more property to Popover. Both components that we struggle to maintain properly to this day and get increasingly complicated everytime we so much as look at them. And we add these properties just to support this one very specific scenario.
> 
> I would propose a much simpler fix that is an additional `<div>` if the `input` type is used, and some additional CSS. All added to DatePicker, the component we're trying to fix, with no changes being made to Dropdown and to Popover. The fix would look like this, which I think is perfectly acceptable: image

@ArakTaiRoth When I picked up this issue @twjeffery and I discussed this as a potential solution and came to the conclusion that given the importance of helper text as labels for this configuration of the date picker having the inputs stack creates some ambiguity in the labeling and it was most desirable to have all three inputs stay on the same row. @bdfranck I'd be curious to get your thoughts on this as well.

bdfranck:
I made a test branch that cherry picks @Spark450's commits and includes a [CSS-only fix](https://github.com/GovAlta/ui-components/pull/4134/changes/c2a5dd9fa919c133ff53dab78b08e1a5ca0b140d) to the narrow width issue. @ArakTaiRoth Do you think this change would be less disruptive? If so, I can add my commit to this branch.

image


Here's the demo page to see it in action:
https://govalta.github.io/ui-components/pr-preview-react/pr-4134/bugs/3815

ArakTaiRoth:
> I made a test branch that cherry picks @Spark450's commits and includes a [CSS-only fix](https://github.com/GovAlta/ui-components/pull/4134/changes/c2a5dd9fa919c133ff53dab78b08e1a5ca0b140d) to the narrow width issue. @ArakTaiRoth Do you think this change would be less disruptive? If so, I can add my commit to this branch.
> 
> image
> Here's the demo page to see it in action: https://govalta.github.io/ui-components/pr-preview-react/pr-4134/bugs/3815

This is better, but still dangerous in my opinion, and maybe doesn't look quite as good as my suggestion (it cuts off the text in the Dropdown). It modifies the CSS in Popover, which could  have glaring repercussions on DatePicker, Dropdown, MenuButton, or even just vanilla Popover that we may not catch, and is modifying a component that shouldn't need to be modified for this story. I'm still wondering why we have this requirement that everything be on one line.

Spark450:
As discussed, we have opted for the more simple solution of letting the date picker fields wrap when there is not sufficient space. Due to the complexity of this PR in its current state it will be closed.

PR #4094, opened 2026-06-25
feat(#3917): add public form reference demo (Step 1)

### LINKED ISSUE #3917: Spike: identify public form-specific components needed
## Description

Before public form work can proceed, we need to investigate what (if any) new components are required to support the public form use case. This spike should produce a clear list of components to build or adapt, so that downstream work can be scoped and prioritized.

This issue must be resolved before the following can proceed:
- #3419
- #3420

## Acceptance Criteria

- [ ] Research completed and findings documented
- [ ] Go through existing Public Form components and identify any that can be used (perhaps with slight changes)
- [ ] List of net-new components (if any) identified with rationale
- [ ] List of existing components requiring changes (if any) identified
- [ ] Findings shared with the team so #3419 and #3420 can be unblocked

twjeffery:
# Before (the change)

#3917 asked which components the public form needs. The plan out of that spike makes Step 1 a coded reference demo, then Step 2 builds the real components against it. There's no such reference today: the current demo rides on an unmerged fork of the form runtime that only lives on a separate branch, and the page-scale starters (start, review, result) are generic for other scenarios.

# After (the change)

The Step 1 reference demo: a self-contained public form gallery built entirely from standard GoA components, with no form-runtime fork. It lives in the React example app under `apps/prs/react/src/routes/public-form/`, and you get to it from Public form in the playground side menu. It opens as its own full page rather than inside the playground layout, since these are whole public form pages with their own header and footer. Browser back returns to the playground. The task-list page is the home, and everything is reachable from it.

The gallery, by section:

- **Question types:** single question, multiple questions on a page, question with details, grouped fields, reveal, content before the question, file upload.
- **Repeating questions:** inline list, add in a modal, add in a drawer, multi-step entry, new page per item with branches. (Table is the one repeater pattern from the spike I didn't build, since it's the overkill-for-simple-cases case.)
- **Before you apply:** a real multi-page eligibility task with next-logic and branching, plus a change-from-review flow that rewalks any newly-opened questions.
- **Review pages:** editable, read-only, per-question-type, and a repeating-group summary, built from a form summary, section, and item hierarchy.
- **Result pages:** eligible, not eligible, submitted.

It's all additive and isolated to the example app, so it doesn't touch any shipped component.

This is the reference for the component list in the spike. Most of those pieces are hand-built here because they don't exist as components yet: the error summary, the task list, the parent and child form summaries, the inline repeater layout, and the public-form page layout. Building by hand also turned up a few smaller DX gaps (form item error not reddening its control, date picker losing partial and invalid states, file upload having no error prop), captured separately for Step 2.

One accessibility note for the Form page component: the single-question pattern uses the field label as the page heading, but Form Item can't render its label as a heading, so those pages have no programmatic h1. That's filed as #4096, along with the other findings for Step 2.

One thing left for later: the task-list completion statuses are decorative right now, they don't update when you come back to a task. That needs app-level state.

- [ ] I have created necessary unit tests (this is a reference demo, verified in-browser rather than unit-tested)
- [ ] I have tested the functionality in both React and Angular (React-only example app; the Angular reference comes in Step 2)

## Steps needed to test

The preview link is the quickest way in. Open it and choose Public form in the side menu.

To run it locally instead, in a fresh checkout of this branch:

1. `npm ci --legacy-peer-deps`
2. `npm run build`
3. `npm run serve:prs:react`

From the task-list home: submit a form with empty required fields (the error summary appears and moves focus to the field), walk the eligibility task and change an answer from the review page, and add a few items through the modal, drawer, and new-page patterns.

### DIFF
--- a/apps/prs/react/src/app/app.tsx
+++ b/apps/prs/react/src/app/app.tsx
@@ -183,6 +183,13 @@ export function App() {
             label="Everything"
             url={`${baseUrl}everything`}
           />
+          {/* Opens full page, outside this shell, since the examples are whole
+              public form pages with their own layout. Browser back returns here. */}
+          <GoabWorkSideMenuItem
+            icon="document-text"
+            label="Public form"
+            url={`${baseUrl}public-form`}
+          />
         </>
       }
     />
--- a/apps/prs/react/src/main.tsx
+++ b/apps/prs/react/src/main.tsx
@@ -8,6 +8,7 @@ import App from "./app/app";
 import { EverythingRoute } from "./routes/everything";
 import { EverythingBRoute } from "./routes/everything-b";
 import { prRouteDefinitions } from "./app/route-manifest";
+import { PublicFormApp } from "./routes/public-form/public-form-app";
 const root = ReactDOM.createRoot(document.getElementById("root") as HTMLElement);
 
 root.render(
@@ -16,6 +17,7 @@ root.render(
       <GoabThemeProvider>
         <Routes>
           <Route path="/everything/b" element={<EverythingBRoute />} />
+          <Route path="/public-form/*" element={<PublicFormApp />} />
           <Route path="/" element={<App />}>
             <Route path="everything" element={<EverythingRoute />} />
             {prRouteDefinitions.map((route) => {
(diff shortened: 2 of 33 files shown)


twjeffery:
Note on the force-push: this PR is back to its original scope, the reference demo only. The component work that was briefly on this branch moved to tom/public-form-components so it can be finished and reviewed properly on its own. Nothing was lost in the move.

ArakTaiRoth (approved):

PR #4095, opened 2026-06-25
fix(#3986): GoabButton doesn't take 100% width inside GoabButtonGroup on mobile

### LINKED ISSUE #3986: GoabButton doesn't take 100% width inside GoabButtonGroup on mobile
# Info

Our `GoabButton` normally takes 100% width of the parent container on mobile. However, if you place the button inside a `GoabButtonGroup`, it no longer takes up 100% width, instead fitting to content.

## Acceptance Criteria:

1. All GoabButton components inside GoabButtonGroup should take up 100% width on mobile

Spark450:
# Before (the change)

Buttons inside a GoabButtonGroup do not take 100% width on mobile. The group's shadow host has no explicit display rule, so it defaults to inline. When the button group sits inside a flex row parent (a common layout, such as a page actions row), it shrinks to its content width instead of filling the row, so each button's mobile width:100% resolves against that shrunk width rather than the real available space.

The Angular wrapper has a second, related issue: GoabButtonGroup (Angular) wraps the underlying goa-button-group element in its own host element with no display override. When that wrapper is itself a flex item in a row, the wrapper shrinks to content width, which prevents the web component level fix from taking effect for Angular consumers even after the Svelte fix.

# After (the change)

The button group's host is explicitly set to display:block and width:100% on mobile, and the inner flex container switches to flex-direction:column with align-items:stretch. Buttons inside the group now reliably stack and span 100% width on mobile regardless of the parent layout context.

The Angular wrapper now sets display:contents on its own host, matching the existing pattern used by the Spacer and Badge wrappers, so the wrapper element disappears from layout and the underlying web component participates directly in the parent's flex layout.

## Steps needed to test

1. Open the React playground at bugs/3986 or Angular playground at bugs/3986.
2. Resize the viewport below 624px wide.
3. Confirm the buttons in each test case stack and take 100% width, including Test 2, which places the button group inside a flex row container.

Fixes #3986

### DIFF
--- a/apps/prs/angular/src/routes/bugs/3986/bug3986.component.html
+++ b/apps/prs/angular/src/routes/bugs/3986/bug3986.component.html
@@ -1,0 +1,39 @@
+<goab-text tag="h1" mt="m" mb="m">
+  Bug 3986 - GoabButton doesn't take 100% width inside GoabButtonGroup on mobile
+</goab-text>
+<goab-text mb="l">
+  GoabButton normally takes 100% width of the parent container on mobile. However, if
+  you place the button inside a GoabButtonGroup, it no longer takes up 100% width,
+  instead fitting to content. Resize the viewport below 624px wide to test.
+</goab-text>
+
+<goab-text tag="h2" mb="s">Test 1: ButtonGroup in normal flow</goab-text>
+<goab-button-group alignment="end" gap="relaxed" mb="l">
+  <goab-button type="secondary">Cancel</goab-button>
+  <goab-button type="primary">Confirm</goab-button>
+</goab-button-group>
+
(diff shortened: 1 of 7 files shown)


ArakTaiRoth (approved):

bdfranck (approved):

| Browsers | Test 1 | Test 2 | Test 3 |
| --- | --- | --- | --- |
| Chrome | ✅ | ✅ | ✅ |
| Firefox | ✅ | ✅ | ✅ |
| Safari | ✅ | ✅ | ✅ |
| Mobile Chrome | ✅ | ✅ | ✅ |
| Mobile Safari | ✅ | ✅ | ✅ |

Looks good to me! 👍

PR #4097, opened 2026-06-25
fix(#3893): add stretch prop to GoabBlock to fix width with alignment

### LINKED ISSUE #3893: `alignment` property in GoabBlock breaks width
# Info

Reported by Jon Sauter

If you setup a `GoabBlock` component using the `alignment` property (doesn't seem to matter the value you use):
```typescript
<GoabBlock direction="column" alignment="start" minWidth="100%" width="full">
  <GoabText size='body-l' tag='div'>Hello World</GoabText>
  <GoabFormItem
    label='Test'
    maxWidth="100%"
  >
    <GoabTextArea name='test' value={test} />
  </GoabFormItem>
</GoabBlock>
```
Any `width` properties you set don't seem to matter, it just defaults to an extremely small width.

## Acceptance Criteria:

1. `width` properties should function as expected no matter the `alignment` property.

Spark450:
# Before (the change)

`GoabBlock` with `alignment` set to anything other than `normal` (e.g. `start`/`center`/`end`) causes children to collapse to their content width instead of respecting `width`/`maxWidth`. This isn't specific to `GoabFormItem` — it affects any child, because `alignment` maps directly to `align-items`, and `normal` is the only value that behaves like `stretch`.

# After (the change)

Added a new opt-in `stretch` boolean prop to `GoabBlock` (default `false`, fully backward compatible). When `true`, it forces `align-self: stretch` on slotted children via `:global(::slotted(*))`, independent of the container's `align-items`. This lets `alignment="start"/"center"/"end"` (positioning) and full-width children (sizing) work together, since `align-self` on a child overrides `align-items` on the container.

Full root-cause writeup and the discarded alternatives (an `alignment="stretch"` enum value doesn't work because `align-items` is a single property and can't be both `flex-start` and `stretch`; a per-component `width` prop on `GoabFormItem` only would leave the bug for every other component type) are in the issue comments: https://github.com/GovAlta/ui-components/issues/3893#issuecomment-4803660147 and https://github.com/GovAlta/ui-components/issues/3893#issuecomment-4804055397

Changes:
- `Block.svelte`: new `stretch` prop, `.stretch` class, `::slotted(*)` CSS rule
- React wrapper (`block.tsx`): `stretch?: boolean`, `"true"`/`undefined` pattern
- Angular wrapper (`block.ts`): `stretch` input with `booleanAttribute` transform
- Unit tests added in all three frameworks
- `doc.md` updated with a usage example
- Playground page (`bugs/3893`) updated with a Test 3 case demonstrating the fix
- Docs sandbox (`docs/src/data/configurations/block.ts`): new "Stretch" example alongside the existing Alignment example, showing the prop fixing the same collapse it documents

Verified in the React playground: with the fix, the form item renders at full width (matching the default-alignment baseline), versus collapsing to ~34px without it.

Fixes #3893

(Angular verified via unit tests; visual verification was done in the React playground.)

## Steps needed to test

1. Pull this branch and run `npx nx build web-components` (the playground app resolves `@abgov/web-components` from `dist/`, so a rebuild is needed to pick up the Svelte change)
2. Run `npm run serve:prs:react`
3. Navigate to `Bugs > 3893 - Block alignment breaks width`
4. Compare Test 1 (`alignment="start"`, no fix) vs Test 2 (default alignment) vs Test 3 (`alignment="start"` + `stretch="true"`) — Test 1's text area collapses to a tiny width, Test 2 and Test 3 both fill the available width


libs/web-components/src/components/block/doc.md
```diff
@@ -11,3 +11,13 @@ Use it like this:
   <div>Item 3</div>
 </goa-block>
 ```
+
```
ArakTaiRoth: This file doesn't serve a purpose. No need to modify it.

apps/prs/angular/src/routes/bugs/3893/bug3893.route.json
```diff
@@ -0,0 +1,6 @@
+{
+  "type": "bug",
+  "id": "3893",
+  "path": "bugs/bug3893",
```
ArakTaiRoth: Cleanup this route. It should be `bugs/3893` to be consistent with all other routes in Angular and in React

willcodeforcoffee (approved):

libs/web-components/src/components/block/doc.md
Spark450: Reverted, removed the doc.md changes.

apps/prs/angular/src/routes/bugs/3893/bug3893.route.json
Spark450: Fixed, route is now bugs/3893 and the folder is renamed to match (was bugs/bug3893).

ArakTaiRoth (approved):

Issue #2941, opened 2025-08-01
Block Component Enhancement: Support Wrapping Layout


twjeffery:
## Summary
Extend the existing [Block](https://design.alberta.ca/components/v6.0.0+-react/block#tab-0) utility component to support wrapping behavior when its contents exceed the available horizontal space. This enhancement will allow the Block component to behave more like a flexible container that wraps items onto the next line as needed, without enforcing strict widths or filling the space like the [Grid](https://design.alberta.ca/components/v6.0.0+-react/grid#tab-0) component.

- The Block component currently acts as a utility layout component for spacing and alignment but does not support wrapping.
- The Grid component supports wrapping but enforces a minChildWidth, making it less flexible for this use case.
- A wrapping Block would provide a more natural flow for horizontally aligned controls, like filters and search inputs, as they collapse on smaller screen sizes.

**Requested by a product team who needs a container that flows responsively and wraps filters/search controls on smaller screens. (See screenshot below.)**

<img width="1435" height="234" alt="Image" src="https://github.com/user-attachments/assets/50ca2e8a-d2d9-4eb0-a918-1e50fc5b9060" />

## Acceptance Criteria
- Block component accepts a wrap prop
- Children wrap onto new lines as needed without breaking layout
- Does not impact existing Block behavior
- Document wrap usage in component page with example

### Implementation ideas
```
<GoaBlock wrap gap="m" align="center">
  <GoaSearch />
  <GoaFilterChips />
</GoaBlock>
```

PR #4104, opened 2026-06-29
fix(#4004): forward heading slot to internal push drawer

### LINKED ISSUE #4004: PushDrawer heading doesn't work with slotted content
# Info

In our documentation for `GoabPushDrawer`, and in our code, support for slotted content in the `heading` property is there. However this doesn't work, only `string` content works with the `heading` property.

This is likely because in our `PushDrawer.svelte` we either create a `GoabDrawer` or a `GoabPushDrawerInternal`, and just pass `{heading}` as a property into each. This only passes `heading` as a string, not as a slot. We need to pass `heading` the same way we pass `actions` if it is a slot.

## Acceptance Criteria:

1. Need to be able to use slotted content with the `heading` property with `GoabPushDrawer`

Spark450:
Fixes #4004
Related: #2941

# Before (the change)

`GoabPushDrawer` documents support for slotted `heading` content, but only string headings rendered. `PushDrawer.svelte` forwarded `heading` only as a prop and never forwarded the `heading` slot to its internal drawer, so custom heading content was silently dropped.

# After (the change)

`PushDrawer.svelte` now forwards the `heading` slot into both `goa-drawer` (mobile) and `goa-push-drawer-internal` (desktop), the same way `actions` is already forwarded. Both internal components already rendered `<slot name="heading" />`, they just never received it. No wrapper changes were needed: the React and Angular wrappers already pass `<div slot="heading">` through.

**Reviewer note (header layout):** While testing the slotted heading, the V2 header in `PushDrawerInternal.svelte` needed alignment work, so this PR also tidies the `.drawer-default-header` row to match the regular `Drawer` header:

- Top-align the row (`align-items: flex-start`) with the same close-button nudge the `Drawer` uses, so the close button stays pinned top-right and aligns to the heading's first line.
- Let the heading content shrink (`min-width: 0`) so long, wrappable heading content flows onto the next line instead of pushing the close button past the header's right padding.
- Keep the close button at its intrinsic size (`flex-shrink: 0`) and add `column-gap: var(--goa-space-s)` for a minimum gap before the close button.

`Drawer.svelte` already has its own header layout, so it was left untouched.

**Known limitation (tracked by #2941):** `GoabBlock` has no wrapping support (its `direction: row` is a non-wrapping flex row). A block-based heading containing a badge will overflow in a narrow drawer because the block itself cannot wrap. This fits fine at the default drawer width and wraps correctly for plain wrappable content (see Test 3). Once #2941 lands and `GoabBlock` gains a `wrap` prop, a block heading with a badge can wrap as well.

## Steps needed to test

1. Run the React playground: `npm run serve:prs:react`
2. Open `http://localhost:4201/bugs/4004`
3. Under Test 2, click "Open slotted heading" and confirm the heading renders custom content (text plus a "New" badge).
4. Under Test 3, click "Open long heading" and confirm the long heading wraps onto multiple lines while the close button stays pinned top-right.
5. Narrow the window below ~1023px and reopen: it switches to the overlay drawer and still shows the slotted heading.
6. Repeat in Angular: `npm run serve:prs:angular`, open `http://localhost:4200/bugs/4004`.

### DIFF
--- a/libs/web-components/src/components/push-drawer/PushDrawerInternal.svelte
+++ b/libs/web-components/src/components/push-drawer/PushDrawerInternal.svelte
@@ -332,6 +332,12 @@
     font: var(--goa-push-drawer-heading-typography);
   }
 
+  /* V2: minimum separation so wide slotted heading content never butts
+     up against the close button */
+  .v2 .drawer-default-header {
+    column-gap: var(--goa-space-s);
+  }
+
   .drawer-actions {
     width: 100%;
     padding: var(--goa-push-drawer-actions-padding-top)
--- a/libs/web-components/src/components/push-drawer/PushDrawerInternal.svelte
+++ b/libs/web-components/src/components/push-drawer/PushDrawerInternal.svelte
@@ -145,26 +145,28 @@
     <div id="goa-drawer-heading" slot="header" class="drawer-header">
       <div class="drawer-default-header">
         {#if heading || $$slots.heading}
-          {#if heading}
-            {#if version === "2"}
-              <goa-text
-                size={_scrollPos === "middle" || _scrollPos === "bottom"
-                  ? "heading-xs"
-                  : "heading-s"}
-                as="h3"
-                mt="2xs"
-                mb="none"
-              >
-                {heading}
-              </goa-text>
+          <div class="drawer-heading-content">
+            {#if heading}
+              {#if version === "2"}
+                <goa-text
+                  size={_scrollPos === "middle" || _scrollPos === "bottom"
+                    ? "heading-xs"
+                    : "heading-s"}
+                  as="h3"
+                  mt="2xs"
+                  mb="none"
+                >
+                  {heading}
+                </goa-text>
+              {:else}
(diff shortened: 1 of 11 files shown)


libs/web-components/src/components/push-drawer/PushDrawerInternal.svelte
```diff
@@ -332,6 +334,29 @@
     font: var(--goa-push-drawer-heading-typography);
   }
 
+  /* V2: minimum separation so wide slotted heading content never butts
```
ArakTaiRoth: You can safely remove all the comments added in this file.

willcodeforcoffee (approved):
I agree with @ArakTaiRoth that we don't really need those extra comments. Otherwise LGTM!

libs/web-components/src/components/push-drawer/PushDrawerInternal.svelte
Spark450: Removed in d9f1aea.

ArakTaiRoth (approved):

PR #4105, opened 2026-06-29
fix(#3571): persist framework selection across docs pages

### LINKED ISSUE #3571: Framework state not saved / partially saved
# Info

When browsing components. The framework you select to view examples and properties in, is not saved or is partially saved between component and example changes.

## Replication Steps

1. Go to a component page
2. Change the framework to Angular (React seems to be the default) - notice that the code in the example changes
3. Change the component selected
4. Notice that the page is back to showing React example code

## Acceptance Criteria:

1. Framework chosen should be saved and used on all pages

Spark450:
# Before (the change)

Framework selection (React / Angular / Web Components) was only partially saved when browsing component pages. Picking Angular updated the current page, but navigating to another component reset the code examples back to React. The properties table kept the saved framework, but the code example switcher did not, so the two fell out of sync.

# After (the change)

The selected framework persists across all component and example pages for both the code examples and the properties table.

Root cause: the example switcher uses `goa-tabs`, which defaults to `navigation="hash"`. That deep-links the active tab into the URL and writes it back via `history.replaceState` on mount. On every fresh page load the switcher mounted as React (the SSR default), stamped `react` into the hash, and that hash navigation overrode the restored preference. The properties table reads `localStorage` directly, so it stayed correct, which is why the behavior looked partially saved.

Fix: set `navigation="none"` on the framework switcher tabs in `docs/src/components/CodeSnippet.tsx`. The saved preference (`localStorage` + cross island event sync) is now the single source of truth, and the URL is no longer polluted with `#react` / `#angular`. Section anchors such as `#properties` still work.

## Steps needed to test

1. Run the docs site (`astro dev --root docs`) and open any component page, for example `/components/modal/`.
2. In the code examples, switch the framework to Angular. Confirm the example and the properties table both show Angular.
3. Navigate to a different component, for example `/components/button/`.
4. Confirm the code examples and the properties table both still show Angular (before this change the examples reverted to React).
5. Switch back to React and repeat to confirm it persists both ways.

Notes:
- Docs only change. No web component, React, or Angular wrapper changes, so no `apps/prs` playground page applies.
- Unit tests box left unchecked: `CodeSnippet` has no existing test in the docs package and the change is a single declarative prop on a third party tabs component, verified manually in the running site.

Fixes #3571

### DIFF
--- a/docs/src/components/CodeSnippet.tsx
+++ b/docs/src/components/CodeSnippet.tsx
@@ -232,25 +232,21 @@ export function CodeSnippet({
     // querySelectorAll on the host always returns an empty list.
     const tabs = goaTabs.shadowRoot?.querySelectorAll<HTMLElement>('[role="tab"]');
     const targetTab = tabs?.[targetIndex];
-    if (!targetTab) return;
+    if (!targetTab || targetTab.getAttribute("aria-selected") === "true") return;
 
-    tabs.forEach((tab, i) => {
-      tab.setAttribute("aria-selected", i === targetIndex ? "true" : "false");
-      tab.setAttribute("tabindex", i === targetIndex ? "0" : "-1");
-    });
-
-    const tabContents = goaTabs.querySelectorAll("goa-tab");
-    tabContents.forEach((content, i) => {
-      // goa-tab listens for "tabs:set-open" on its own shadow-root section,
-      // not on the host element, so the event must be dispatched there.
-      const innerTarget = content.shadowRoot?.querySelector("section") ?? content;
-      innerTarget.dispatchEvent(
-        new CustomEvent("tabs:set-open", {
-          composed: true,
-          detail: { open: i === targetIndex },
-        }),
-      );
-    });
+    // Tab selection, panel visibility, and the segmented pill's position are
+    // all driven by Tabs.svelte's own internal click handler. Replicating
+    // those DOM mutations from outside missed pieces (the indicator stayed
+    // put, panels didn't toggle) because they're tracked by internal state
+    // this component can't reach. A real click reuses that logic wholesale.
+    // It also moves focus to the clicked tab, which would steal focus from
+    // wherever the user is reading for every other switcher on the page, so
+    // restore whatever was focused beforehand.
+    const previouslyFocused = document.activeElement as HTMLElement | null;
+    targetTab.click();
+    if (previouslyFocused && previouslyFocused !== targetTab && document.contains(previouslyFocused)) {
+      previouslyFocused.focus({ preventScroll: true });
+    }
   };
 
   useEffect(() => {
--- a/docs/src/components/CodeSnippet.tsx
+++ b/docs/src/components/CodeSnippet.tsx
@@ -228,8 +228,10 @@ export function CodeSnippet({
     if (!goaTabs) return;
 
     const targetIndex = availableFrameworks.indexOf(framework);
-    const tabs = goaTabs.querySelectorAll('[role="tab"]');
-    const targetTab = tabs[targetIndex] as HTMLElement | undefined;
+    // The tab buttons render inside goa-tabs' own shadow root, so a plain
+    // querySelectorAll on the host always returns an empty list.
+    const tabs = goaTabs.shadowRoot?.querySelectorAll<HTMLElement>('[role="tab"]');
+    const targetTab = tabs?.[targetIndex];
     if (!targetTab) return;
 
     tabs.forEach((tab, i) => {
@@ -239,7 +241,10 @@ export function CodeSnippet({
 
     const tabContents = goaTabs.querySelectorAll("goa-tab");
     tabContents.forEach((content, i) => {
-      content.dispatchEvent(
+      // goa-tab listens for "tabs:set-open" on its own shadow-root section,
+      // not on the host element, so the event must be dispatched there.
+      const innerTarget = content.shadowRoot?.querySelector("section") ?? content;
+      innerTarget.dispatchEvent(
         new CustomEvent("tabs:set-open", {
           composed: true,
           detail: { open: i === targetIndex },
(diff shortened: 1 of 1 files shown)


ArakTaiRoth (dismissed):

vanessatran-ddi (commented):
This AC is not met 

Image

So when I choose the Angular for this tab: 

Image

So if I go to another page (Badge) or go down to Examples, the default choice is React.

Spark450:
Good catch, thanks for testing that. Found and fixed the underlying bug.

`syncGoaTabsSelection` (the function that propagates a framework change to every other code switcher already mounted on the page) was querying `[role="tab"]` directly on the `goa-tabs` host element. Those tab buttons actually render inside `goa-tabs`' own shadow root, so that query always returned an empty list and the function silently did nothing for every switcher except the one you clicked. It also dispatched `tabs:set-open` on the `goa-tab` host, but `goa-tab` listens for that event on its own shadow-root section, not the host, so even reaching that code wouldn't have toggled the panel content.

Fixed both lookups to go through `shadowRoot`. Verified locally: selecting Angular on the Button page now updates the top demo switcher, the Properties tab, and every example in the Examples tab (35 examples) in one click, and it still persists across page navigation.

Pushed as a new commit, 154bb5c..cad43ed.

Spark450:
Fixed the indicator issue too.

The previous fix synced the active tab by manually replicating Tabs.svelte's DOM mutations (aria-selected, dispatching tabs:set-open) from outside the component. That covered the bold text and the code content, but missed the segmented variant's pill background: its position is recalculated internally by Tabs.svelte on click, driven by a private `_currentTab` variable this external code never touches, so the pill stayed wherever it last was.

Rather than keep replicating internal behavior piece by piece, the switcher now just calls `.click()` on the target tab button, reusing Tabs.svelte's own click handler wholesale, so tab selection, panel content, and the pill all stay consistent in one step. Since a real click also moves focus to the clicked tab, the previously-focused element is restored afterward so propagating a change to other switchers on the page doesn't steal focus from wherever you're reading.

Pushed as cad43ed..1b20bf1.

ArakTaiRoth:
@Spark450 I didn't care for the fix that you had, because it was using the shadow dom and not our actual Tab component properties to adjust. I submitted an entirely different change for this issue that should work and cleans up the tab handling code quite significantly.

@vanessatran-ddi Can you re-review my changes?

Spark450:
> @Spark450 I didn't care for the fix that you had, because it was using the shadow dom and not our actual Tab component properties to adjust. I submitted an entirely different change for this issue that should work and cleans up the tab handling code quite significantly.
> 
> @vanessatran-ddi Can you re-review my changes?

@ArakTaiRoth The functionality is maintained and if your version cleans up the code, all the better,  thanks 🙌

docs/src/components/CodeSnippet.tsx
```diff
@@ -195,6 +195,17 @@ function SingleCodeBlock({
 
 const VALID_FRAMEWORKS: Framework[] = ["react", "angular", "webComponents"];
 
+function getSupportedFramework(
```
vanessatran-ddi, line 196: I find this `getSupportedFramework` function hard to follow, mostly because `fallback` 

Image

1. Initial call pass the same value twice `initialFramework, availableFrameworks, initialFramework`, passing 1 value as both `preferred` and `fallback` reads like a mistake until you dig into the body. 
2. `initialFramework` is effectively dead, it is never passed anything but the default `react`. 

Image

3. `stored || current` never returns undefined or `stored`, the function `getFrameworkPreferences()` already returns as `react` by default. 


Image

4. `getSupportedFramework` in the subscribe callback passes the `current` into `fallback` position hides the actual intent. 

What I am thinking is we can make `getSupportedFramework` as simple as below: 

Image

Image

Image

This drops the `initialFramework` prop and the `fallback` param, and the subscribe intent becomes a one-liner.

ArakTaiRoth: @vanessatran-ddi I updated `CodeSnippet.tsx` based on your findings. I had developed it like I had earlier to be more defensive, largely in case there was a situation where a chosen framework wasn't an option, but to be fair that probably wasn't necessary like you said, so all of that has been removed.

docs/src/components/CodeSnippet.tsx
```diff
@@ -195,102 +193,71 @@ function SingleCodeBlock({
...
+  );
 
-    const tabsElement = tabsRef.current.querySelector("goa-tabs");
-    if (tabsElement) {
-      tabsElement.addEventListener("_change", handleTabChange);
-      return () => tabsElement.removeEventListener("_change", handleTabChange);
-    }
-  }, [availableFrameworks]);
+  const selectedTabIndex = Math.max(availableFrameworks.indexOf(selectedFramework), 0);
+  const tabsKey = `${availableFrameworks.join("-")}-${selectedFramework}`;
```
vanessatran-ddi: This `key` remounts `GoabTabs` whenever `selectedFrameworks` changes, so the saved preference visibility flashes on load, with Angular saved, every snippet paints the React tab first, then swaps a frame later. 

https://github.com/user-attachments/assets/42f80fb3-8204-439f-bf39-6c2e49a7886c

Spark450: Fixed the flash while keeping the key-remount approach.

Root cause: the saved preference was applied in a passive effect (`useEffect`), which runs after the browser paints. So the tabs mounted on the default React tab, painted, and then the effect changed `selectedFramework`, which the `key` turned into a remount to the saved framework a frame later, the flash you saw.

I moved that correction into a layout effect (`useLayoutEffect`) so the remount lands before the first paint, meaning the saved framework is the first thing painted. It's wrapped in a small `useIsomorphicLayoutEffect` (falls back to `useEffect` during SSR) to avoid React's "useLayoutEffect does nothing on the server" warning.

Verified locally with Angular saved: the top configuration preview now paints Angular directly with no React frame, and the Properties and Examples tabs still resolve to Angular (no regression). Cross-component sync and cross-page persistence still work.

Pushed as a33766f.

vanessatran-ddi (approved):

PR #4106, opened 2026-06-29
chore: Removed all code for VisualStudio Code Angular support

### LINKED ISSUE #3286: Remove all `.html-data.json` files
# Info

We created a bunch of `.html-data.json` files in all of our components to help VS Code give Angular developers better understanding of how our components worked, code completion, etc..

We don't need these any longer because we've dropped support for LTS, and our new version has Angular wrappers

## Acceptance Criteria:

1. Remove all `.html-data.json` files in all of our components
2. Remove building of these manifest files, `package.json` --> line 15

ArakTaiRoth:
Removed all `*.html-data.json` and `doc.md` files that were previously used to provide some kind of coding support for Angular developers using VisualStudio Code. Also removed all generation that used those files. They aren't necessary anymore with the introduction of our Angular wrappers.

Also updated the `extract-api.ts` script as it wasn't playing nicely for me, and was changing `/`'s in the generated files because I'm the only one running in Windows.

FOR TESTING:
1. Verify that `extract-api.ts` still works for you and produces the correct results
2. Verify that all removed files are actually no longer needed. The only reference to these files that I can still find is in `SEARCH.md` and `README.md` where it is insinuated that `doc.md` files are used to generate the search index, but the command it says it runs to generate that index doesn't actually exist @twjeffery ?


willcodeforcoffee (approved):
Its nice you got the extract-apis thing working!
The only enhancement I can think of is to maybe have it save with LF line endings instead of CRLF. When you run the script all the JSON files look touched in `git status`.
I think you're the only person affected by this ATM though., so it is probably not worth the extra time investment.

ArakTaiRoth:
> Its nice you got the extract-apis thing working! The only enhancement I can think of is to maybe have it save with LF line endings instead of CRLF. When you run the script all the JSON files look touched in `git status`. I think you're the only person affected by this ATM though., so it is probably not worth the extra time investment.

@willcodeforcoffee I implemented a small change to `.gitattributes` that now does what you were suggesting. I am free, thanks.

twjeffery (commented):
@ArakTaiRoth On your search index question: that `index:docs` command doesn't exist, and the index builds from the `.mdx` files now, not `doc.md`. Nothing in the build uses what you're removing, so it's safe.

A few leftovers worth catching:

1. The `doc.md` files under `src/layouts` got missed (`one-column-layout`, `two-column-layout`, `three-column-layout`).
2. `CLAUDE.md`'s folder diagram still lists `html-data.json` and `doc.md`.
3. `SEARCH.md` and `scripts/README.md` describe that old setup, so they can go too.

ArakTaiRoth:
> @ArakTaiRoth On your search index question: that `index:docs` command doesn't exist, and the index builds from the `.mdx` files now, not `doc.md`. Nothing in the build uses what you're removing, so it's safe.
> 
> A few leftovers worth catching:
> 
> 1. The `doc.md` files under `src/layouts` got missed (`one-column-layout`, `two-column-layout`, `three-column-layout`).
> 2. `CLAUDE.md`'s folder diagram still lists `html-data.json` and `doc.md`.
> 3. `SEARCH.md` and `scripts/README.md` describe that old setup, so they can go too.

@twjeffery I believe I've made all the changes you mentioned above. Specifically can you check `SEARCH.md` and `README.md`, I'm not sure if what I updated those sections to is actually entirely accurate.

PR #4112, opened 2026-06-30
fix(#4110): pin workspace bottom notification panels to viewport gap and profile button

### LINKED ISSUE #4110: Workspace side menu: bottom-pinned panels use hardcoded 140px offset (notification panel overflows, account menu has gap)
## Description
When integrating our GoabWorkspaceLayout to React playground template https://github.com/GovAlta/goa-workspace-playground/pull/8 @willcodeforcoffee  helped me test and also reported an issue with our current workspace layout and position of popover (especially after we move on with popover anchor position)

### 1. Notification panel overflows the viewport bottom (no gap)

Image

### 2. Account menu (Settings / Log out) has a gap above the profile button

Image

The profile button's actual top is ~114px from the viewport bottom, so the `140px` value leaves a ~26px gap between the account menu and the profile button. The fixed value also does not adapt to the profile's open vs. collapsed height.

## Steps to Reproduce

1. Run an app that uses `GoabWorkSideMenu` inside `GoabWorkspaceLayout` (e.g. the workspace playground).
2. Click the **Notifications** item near the bottom of the side menu → notification panel overflows the viewport bottom with no gap.
3. Click the **profile** button (Edna Mode) at the bottom → the Settings / Log out menu opens with a visible gap above the profile button.

## Acceptance Criteria
- [ ] Notification panel keeps a ~20px gap from the viewport bottom regardless of where the menu item sits
- [ ] Account menu sits flush against the top of the profile button with no gap, in both expanded and collapsed states

vanessatran-ddi:
# Before (the change)
image

image

# After (the change)

image

image

### DIFF
--- a/libs/web-components/src/components/popover/Popover.svelte
+++ b/libs/web-components/src/components/popover/Popover.svelte
@@ -148,6 +148,24 @@
     const xOffset = hoffset ? parseFloat(hoffset) : 0;
     const yOffset = voffset ? parseFloat(voffset) : 3;
 
+    // Right-positioned popovers/notification panels: mirror the CSS anchor rules
+    // in `.popover-content.position-right` for browsers without anchor positioning.
+    if (position === "right") {
+      // inset-inline-start: anchor(right) + translate-x (--offset-left, default 8px).
+      const rightXOffset = hoffset ? parseFloat(hoffset) : 8;
+      _popoverEl.style.left = `${targetRect.right + rightXOffset}px`;
+      // inset-block-end: max(20px, anchor(bottom) - --offset-bottom (default 0px)).
+      const rightYOffset = voffset ? parseFloat(voffset) : 0;
+      const bottomGap = Math.max(
+        20,
+        window.innerHeight - targetRect.bottom - rightYOffset,
+      );
+      _popoverEl.style.top = `${window.innerHeight - bottomGap - popoverRect.height}px`;
+      _popoverEl.style.transform = "";
+      _popoverEl.style.translate = "0px";
+      return;
+    }
+
     // Recalculate auto position based on current viewport space
     if (position === "auto") {
       const spaceAbove = targetRect.top;
@@ -534,10 +552,15 @@
 
   .popover-content.position-right {
     inset-block-start: unset;
-    inset-block-end: max(8px, anchor(bottom));
+    /* Keep right-positioned popovers at least 20px from the viewport bottom,
+       while preserving the configured bottom offset when there is room. */
+    inset-block-end: max(
+      20px,
+      calc(anchor(bottom) - var(--offset-bottom, 0px))
+    );
     inset-inline-start: anchor(right);
     --popover-translate-x: var(--offset-left, 8px);
-    --popover-translate-y: var(--offset-bottom, 0px);
+    --popover-translate-y: 0px;
     position-try-fallbacks: none;
   }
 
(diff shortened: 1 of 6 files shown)


libs/web-components/src/components/work-side-menu/WorkSideMenu.svelte
```diff
@@ -668,12 +668,25 @@
     padding: var(--goa-space-2xs);
     position: fixed;
     left: var(--goa-space-s);
+    /* Fallback for browsers without CSS anchor positioning. Approximates the
+       profile button's top; the @supports line 679 aligns it exactly. */
     bottom: 140px;
```
ArakTaiRoth: Curious why for the fallback we would leave the `140px` in place that the story directly calls out, "The profile button's actual top is ~114px from the viewport bottom, so the 140px value leaves a ~26px gap between the account menu and the profile button."

Why would we not change the fallback to be `114px` as the story calls for?

libs/web-components/src/components/popover/Popover.svelte
```diff
@@ -534,10 +534,16 @@
 
   .popover-content.position-right {
     inset-block-start: unset;
-    inset-block-end: max(8px, anchor(bottom));
+    /* When menu item/ target of popover is nears the bottom of the screen, we will make sure 20px is the min gap between bottom of notification panel/popover to the screen
```
ArakTaiRoth: Change this comment to:
```
/* Keep right-positioned popovers at least 20px from the viewport bottom,
   while preserving the configured bottom offset when there is room. */
```
That is a lot smaller and keeps the same idea to it.

libs/web-components/src/components/popover/Popover.svelte
```diff
@@ -534,10 +534,16 @@
...
+when the viewport is high, we will make sure the right positioned popover or notification panel lower than the popover target or menu item 140px
+*/
+    inset-block-end: max(
```
ArakTaiRoth, line 557: This fix won't work for browsers that can't implement anchor positioning. I'd recommend adding some manual positioning inside the function `updatePopoverPosition`, specifically for when `position === "right"`.

libs/web-components/src/components/work-side-menu/WorkSideMenu.svelte
vanessatran-ddi: good catch! I updated to 114px

libs/web-components/src/components/popover/Popover.svelte
vanessatran-ddi: I updated the comment

libs/web-components/src/components/popover/Popover.svelte
vanessatran-ddi: You are right. I added the logic to handle position right when browser does not support anchor css

willcodeforcoffee:
Using Safari 18 the horizontal gap between the sidebar and notification panel is noticeably too large:
image

versus Safari 26:
image

Safari 18 being Safari.  🙃

vanessatran-ddi:
> Using Safari 18 the horizontal gap between the sidebar and notification panel is noticeably too large:
> Safari 18 being Safari. 🙃

@willcodeforcoffee I fixed it and also tested with safari 18, it works now 
image

My local (after)
image

and before:
image

willcodeforcoffee (approved):
Tested on Safari 18, 26, modern Chrome and Firefox.
LGTM! Thanks @vanessatran-ddi

docs/package-lock.json
ArakTaiRoth, line 1: Just concerned about this file? There doesn't appear to be an associated `package.json` change for this?

package-lock.json
ArakTaiRoth, line 1: Just concerned about this file? There doesn't appear to be an associated `package.json` change for this?

ArakTaiRoth (approved):

PR #4115, opened 2026-07-02
Test: Minimal fix for modal scrolling

bdfranck:
# Before (the change)

# After (the change)

### DIFF
--- a/apps/prs/angular/src/routes/bugs/2710/bug2710.component.ts
+++ b/apps/prs/angular/src/routes/bugs/2710/bug2710.component.ts
@@ -1,0 +1,58 @@
+import { Component } from "@angular/core";
+import {
+  GoabBlock,
+  GoabButton,
+  GoabButtonGroup,
+  GoabCircularProgress,
+  GoabDetails,
+  GoabDivider,
+  GoabDrawer,
+  GoabLink,
+  GoabModal,
+  GoabText,
+} from "@abgov/angular-components";
+
+@Component({
+  standalone: true,
+  selector: "abgov-bug2710",
+  templateUrl: "./bug2710.component.html",
+  imports: [
+    GoabBlock,
+    GoabButton,
+    GoabButtonGroup,
+    GoabCircularProgress,
+    GoabDetails,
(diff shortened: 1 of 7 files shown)


apps/prs/angular/src/routes/bugs/2710/bug2710.component.ts
```diff
@@ -0,0 +1,58 @@
...
+    this.drawerOpen = true;
+  }
+
+  closeDrawer() {
+    this.drawerOpen = false;
+  }
+
+  showProgress() {
+    this.progressVisible = true;
+    setTimeout(() => (this.progressVisible = false), 3000);
```
Spark450, line 56: The fix looks great to me @bdfranck 🙌 @willcodeforcoffee 

 I noticed a small issue with the Progress Indicator Angular example. The test case doesn't actually dismiss after 3 seconds when I test it — `progressVisible` does flip to `false` (confirmed via `ng.getComponent`), but nothing tells Angular to re-check bindings, so `[visible]` never propagates down to `<goab-circular-progress>` and the loader stays on screen.

Apparently, this is a test-page bug, not an issue with `GoabCircularProgress` or the underlying fix — manually forcing change detection (`ng.applyChanges`) immediately fixes the DOM. It also matches an existing convention in this codebase: every Angular wrapper's `ngOnInit` explicitly calls `this.cdr.detectChanges()` after its own `setTimeout(0)` (see `circular-progress.ts`'s `ngOnInit`) for the same reason.

Suggested fix:

```ts
import { ChangeDetectorRef, Component, inject } from "@angular/core";
// ...
export class Bug2710Component {
  private cdr = inject(ChangeDetectorRef);
  // ...
  showProgress() {
    this.progressVisible = true;
    setTimeout(() => {
      this.progressVisible = false;
      this.cdr.detectChanges();
    }, 3000);
  }
}
```

bdfranck: Thanks! I cherry picked these commits onto your branch and fix the issue there.

PR #4116, opened 2026-07-02
fix(#3738): correct Examples tab URL hash and add example deep links

### LINKED ISSUE #3738: Docs: Examples tab not working as expected
# Info

On a component page, if you click any tab except examples, you get an expected URL hash (ie. click "Usage Guidelines", get #usage-guidelins in the hash in the URL). If you click examples though, you get the hash #tab-1 instead.

Also, if you select any example on the Examples tab, the subsequent URL isn't updated to that example. So there's no way to give someone a URL pointing directly at a specific example.

## Acceptance Criteria:

1. Examples tab updates the # in the URL correctly
2. Clicking an example adds to the URL so that a link direct to that example can be given

Spark450:
# Before (the change)

On a component page, clicking the Examples tab produced a generic `#tab-1` hash instead of a proper slug, and selecting an example didn't update the URL at all, so there was no way to share a link directly to a specific example.

# After (the change)

- Examples tab now uses `slug="examples"` so the URL hash correctly reads `#examples`.
- Clicking an entry in the tab's table of contents (or the example's "copy link" button) now updates the URL to include both the tab hash and the example anchor, e.g. `#examples#example-confirm-a-destructive-action`.
- Loading that URL directly reopens the Examples tab and scrolls to the correct example, including a retry/resettle fix for pages where other examples are still hydrating and would otherwise shift the layout out from under the scroll.

Closes #3738

## Steps needed to test

1. Run `npx nx run web-components:build` then `npm run docs:dev`
2. Open any component page, e.g. `/components/button`
3. Click the Examples tab — the URL hash should become `#examples`
4. Click an example in the right-side table of contents — the URL should update to `#examples#example-<slug>`
5. Reload that exact URL — it should reopen the Examples tab and scroll to that example

### DIFF
--- a/docs/src/components/ExamplePreview.tsx
+++ b/docs/src/components/ExamplePreview.tsx
@@ -13,6 +13,7 @@ import { useState, useEffect, useRef } from "react";
 import { GoabTooltip } from "@abgov/react-components";
 import { CodeSnippet } from "./CodeSnippet";
 import type { ExampleCode } from "../lib/example-code";
+import { getActiveTabHash } from "../lib/tab-hash";
 import DOMPurify from "dompurify";
 
 // Allow GoA web component custom elements through DOMPurify.
@@ -114,7 +115,9 @@ export function ExamplePreview({
   // Handle copy link
   const handleCopyLink = async () => {
     try {
-      const url = `${window.location.origin}${window.location.pathname}#${anchorId}`;
+      const tabHash = getActiveTabHash(previewRef.current);
+      const hash = tabHash ? `${tabHash}#${anchorId}` : anchorId;
+      const url = `${window.location.origin}${window.location.pathname}#${hash}`;
       await navigator.clipboard.writeText(url);
       setCopied(true);
       setTimeout(() => setCopied(false), 2000);
--- a/docs/src/components/TableOfContents.tsx
+++ b/docs/src/components/TableOfContents.tsx
@@ -8,6 +8,7 @@
 
 import css from "./toc.module.css";
 import { useEffect, useState, useRef } from "react";
+import { getActiveTabHash } from "../lib/tab-hash";
 
 export type TOCItem = {
   title: string;
@@ -23,6 +24,7 @@ export function TableOfContents({ cssQuery }: TOCProps) {
   const [items, setItems] = useState<TOCItem[]>([]);
   const [activeId, setActiveId] = useState<string | null>(null);
   const containerRef = useRef<HTMLElement>(null);
+  const initialHashHandled = useRef(false);
 
   useEffect(() => {
     // Check if element is visible (not hidden by display:none or in hidden tab)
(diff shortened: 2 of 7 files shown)


willcodeforcoffee (approved):
I'm growing a little concerned with all the inline functions and size of the `useEffect` method in `TableOfContents.tsx`... Taken as a whole its getting a little challenging to read!
Maybe in a future PR if we add anything else we can refactor that.

ArakTaiRoth (commented):
There are 2 issues present in testing this. Though I'm not entirely sure where they lie.

1. If a user switches between the Properties, Examples, Usage Guidelines, Accessibility tabs with the keyboard, the hash doesn't get updated it still gets added to.
2. If a user goes into examples, selects an example, it updates the hash correctly, but then if a user selects a different language for that example, the hash completely changes to that language hash, and the example hash is completely removed.

Also, I'm in agreement with @willcodeforcoffee suggestion above, the `useEffect` function inside these pages is getting very unwieldy and too large to maintain correctly. I'd lean towards something that trims that down a lot more to something more manageable, rather than adding a lot more to it like we are here.

Spark450:
While fixing the two hash bugs @ArakTaiRoth found above, I noticed a separate, pre-existing issue: loading any component page with zero interaction already leaves extra segments in the URL hash (e.g. `#react#properties`), because each code example's nested language-switcher tabs write their own hash into the URL on mount. Confirmed this reproduces identically on `dev` with `Tabs.svelte` unmodified, so it's not caused by this PR and I'm leaving it out of scope here. Filed separately as #4126.

Spark450:
@ArakTaiRoth thanks for finding these, both are fixed now.

Root cause for both was the same: `Tabs.svelte`'s hash-merge logic only recognized a tab's own previous hash if it started with `"tab-"`, and the enabled-tab click handler never called `preventDefault()`, so the browser's native anchor navigation could still overwrite the URL right after the correct hash was computed.

1. Keyboard nav accumulating stale hash segments: fixed by tracking the actual set of hashes each tabs instance can produce, instead of pattern-matching on `"tab-"`.
2. Language switch wiping the example hash: fixed by preventing the native navigation so the merged hash sticks.

Also split up the `TableOfContents.tsx` `useEffect` per your and @willcodeforcoffee  readability feedback, and added regression tests for both bugs (verified they fail on the old code and pass on the fix).

willcodeforcoffee (approved):
I took some time reading and understanding the code. I was thinking originally of keeping one `useEffect` method and extract the methods inside (like `refreshHeadings` or `handleTabChange`) out...

Google isn't a big fan:
> Using multiple `useEffect` hooks in a single React component is technically allowed, but it frequently leads to serious architectural flaws, performance bottlenecks, and hard-to-track bugs

The AI result from Google included this insight from Reddit:

>  Multiple `useEffect` calls are not a problem if each one is focused i.e. one handles an `IntersectionObserver` subscription, another makes an HTTP call. Personally I'd favor multiple smaller `useEffect` calls.

Now that I'm used to reading multiple `useEffect` calls this actually feels like a clean enough solution. The callbacks are cleaned up nicely, and each effect and method is documented.

We _are_ counting on future us/future devs to respect the focus and not append extra logic in one of these more "focused" `useEffects`. That seems reasonable.

If you really want to go fancy you _could_ extract some into their own hooks, but it is debatable how much more that would actually benefit the code anymore.

Image

ArakTaiRoth (approved):

PR #4120, opened 2026-07-03
fix(#4013): remove wrong React type names from Angular Footer TemplateRef docs

### LINKED ISSUE #4013: Type of TemplateRef under Footer docs is wrong
The extrac-api ts generated wrongly for the type of `meta` and `nav`, the `GoabAppFooterMetaSection` and `GoabAppFooterNavSection` are for React, not for Angular.
Beside the docs, it also affects the generated output for MCP and md files.  

Image

MD file generated also wrong:

Image

## Acceptance Criteria:
- extrac-api should remove `GoabAppFooterMetaSection`and`GoabAppFooterNavSection` from template reference

Spark450:
## Summary
- Angular's Footer docs listed `meta`/`nav` slot types as `GoabAppFooterMetaSection` and `GoabAppFooterNavSection`, but those are React component names, not Angular's.
- Removed the `angular` entry from `SLOT_TYPE_OVERRIDES.footer` in `docs/src/scripts/extract-api.ts` so it falls back to the correct `TemplateRef` type (the React override is unaffected).
- Regenerated `docs/generated/component-apis/footer.json` via `npm run extract-api`.

## Before / After
Before: Angular TemplateRef section showed `meta: GoabAppFooterMetaSection`, `nav: GoabAppFooterNavSection`.
After: Angular TemplateRef section shows `meta: TemplateRef`, `nav: TemplateRef`. React section is unchanged.

## Test plan
- [x] Ran `npm run extract-api` and confirmed only `footer.json` changed, and only the `angular` slot types were affected.
- [x] Verified in the docs site (`npm run dev` in `docs/`) that the Footer page's Angular tab now shows `TemplateRef` for both `meta` and `nav`, and the React tab still shows `GoabAppFooterMetaSection`/`GoabAppFooterNavSection`.

Fixes #4013

### DIFF
--- a/docs/generated/component-apis/footer.json
+++ b/docs/generated/component-apis/footer.json
@@ -70,13 +70,11 @@
       "slots": [
         {
           "name": "meta",
-          "type": "GoabAppFooterMetaSection",
           "description": "",
           "required": false
         },
         {
           "name": "nav",
-          "type": "GoabAppFooterNavSection",
           "description": "",
           "required": false
         }
--- a/docs/generated/component-apis/footer.json
+++ b/docs/generated/component-apis/footer.json
@@ -70,11 +70,13 @@
       "slots": [
         {
           "name": "meta",
+          "type": "TemplateRef",
           "description": "",
           "required": false
         },
         {
           "name": "nav",
+          "type": "TemplateRef",
           "description": "",
           "required": false
         }
--- a/docs/generated/component-apis/footer.json
+++ b/docs/generated/component-apis/footer.json
@@ -27,20 +27,7 @@
         }
       ],
       "events": [],
-      "slots": [
-        {
-          "name": "meta",
-          "type": "GoabAppFooterMetaSection",
-          "description": "",
-          "required": false
-        },
-        {
-          "name": "nav",
-          "type": "GoabAppFooterNavSection",
-          "description": "",
-          "required": false
-        }
-      ]
+      "slots": []
     },
     "angular": {
       "props": [
@@ -67,20 +54,7 @@
         }
       ],
       "events": [],
-      "slots": [
-        {
-          "name": "meta",
-          "type": "TemplateRef",
-          "description": "",
-          "required": false
-        },
-        {
-          "name": "nav",
-          "type": "TemplateRef",
-          "description": "",
-          "required": false
-        }
-      ]
+      "slots": []
     },
     "webComponents": {
       "props": [
--- a/docs/src/scripts/extract-api.ts
+++ b/docs/src/scripts/extract-api.ts
@@ -199,18 +199,19 @@ const SLOT_TYPE_OVERRIDES: Record<
   Partial<Record<"react" | "angular" | "webComponents", Record<string, string>>>
 > = {
   footer: {
-    react: {
-      nav: "GoabAppFooterNavSection",
-      meta: "GoabAppFooterMetaSection",
-    },
-    angular: {
-      nav: "TemplateRef",
-      meta: "TemplateRef",
-    },
     webComponents: {
       nav: "goa-app-footer-nav-section",
       meta: "goa-app-footer-meta-section",
     },
+  },
+};
+// Slots consumed as dedicated sub-components (e.g. GoabAppFooterMetaSection) rather than
+// typed props/inputs in React and Angular. Documenting them as ReactNode/TemplateRef props
+// would be wrong since consumers never bind a value to them, they nest the sub-component instead.
+const SLOT_FRAMEWORK_EXCLUDES: Record<string, Partial<Record<"react" | "angular" | "webComponents", string[]>>> = {
+  footer: {
+    react: ["nav", "meta"],
+    angular: ["nav", "meta"],
   },
 };
 const ALLOW_INTERNAL_PROP_BY_COMPONENT: Record<string, Set<string>> = {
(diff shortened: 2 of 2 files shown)


docs/generated/component-apis/footer.json
```diff
@@ -70,13 +70,11 @@
       "slots": [
         {
           "name": "meta",
-          "type": "GoabAppFooterMetaSection",
```
ArakTaiRoth, line 73: This removes the type completely, it's not changing it to `TemplateRef` which is the correct output I believe. Same with the below `nav` section.

Spark450: Pushed 242409d47 to restore the Angular override in `extract-api.ts`, this time with the correct value (`TemplateRef` instead of the React section names). `footer.json` now explicitly shows `"type": "TemplateRef"` for both `meta` and `nav`, confirmed in the docs site as well.

docs/src/scripts/extract-api.ts
```diff
@@ -204,8 +204,8 @@ const SLOT_TYPE_OVERRIDES: Record<
       meta: "GoabAppFooterMetaSection",
     },
     angular: {
```
vanessatran-ddi, line 210: For angular, we don't have meta and nav template 

Image
Only 3 properties so this fix is not correct for Angular 

Image

vanessatran-ddi (changes requested):
Same with React, react doesn't have ReactNode Meta and Nav. the fix here is to not display the Template table for both angular and react. 

Image

docs/src/scripts/extract-api.ts
Spark450: @vanessatran-ddi Added a `SLOT_FRAMEWORK_EXCLUDES` list in `extract-api.ts` that drops `meta`/`nav` from the React and Angular slot output for footer, and removed the now-unused react/angular entries from `SLOT_TYPE_OVERRIDES`. Web Components keeps its Slots table since `goa-app-footer` genuinely exposes named slots there. Verified in the docs preview, React and Angular now go straight from Props to Footer Meta/Nav Section Props with no ReactNode/TemplateRef table in between, Web Components is unchanged.

docs/src/scripts/extract-api.ts
```diff
@@ -199,20 +199,21 @@ const SLOT_TYPE_OVERRIDES: Record<
...
-      nav: "GoabAppFooterNavSection",
-      meta: "GoabAppFooterMetaSection",
-    },
     webComponents: {
       nav: "goa-app-footer-nav-section",
       meta: "goa-app-footer-meta-section",
     },
   },
 };
+// Slots consumed as dedicated sub-components (e.g. GoabAppFooterMetaSection) rather than
```
vanessatran-ddi: Should we fold `SLOT_FRAMEWORK_EXCLUDES` into `SLOT_TYPE_OVERRIDES`? Excluding a slot from a framework is just another kind of slot override, (for example render this slot as nothing on the doc), so having two separate configs for the same `nav/meta` slots means the full picture is split across 2 places and can drift when we edit one but not the other one. 
Suggestion is like this: 
```
footer: {
react: {nav: null, meta: null},
angular: {nav: null, meta: null},
webComponents: {
nav: "goa-app-footer-nav-section",
meta: "goa-app-footer-meta-section",,
},
}
```
We now have 2 states (override the type, hide the slot), so hide needs a value like `null` and the filter has to be updated accordingly.

Spark450: Good idea, done. Pushed 53ecfd162: folded `SLOT_FRAMEWORK_EXCLUDES` into `SLOT_TYPE_OVERRIDES` using `null` as the hide-slot sentinel, so footer's `meta`/`nav` behavior per framework now lives in one map instead of two:

```
footer: {
  react: { nav: null, meta: null },
  angular: { nav: null, meta: null },
  webComponents: {
    nav: "goa-app-footer-nav-section",
    meta: "goa-app-footer-meta-section",
  },
},
```

Added a `getSlotOverride()` helper so the lookup can tell "no override" (fall back to the generic ReactNode/TemplateRef detection) apart from "override is explicitly null" (hide it), since `||`/`??` alone can't distinguish an absent key from a null value. Regenerated all 71 components' JSON, footer.json came out identical to the previous fix and nothing else changed. Re-verified all three doc tabs in the preview.

docs/src/scripts/extract-api.ts
```diff
@@ -194,18 +194,22 @@ const WEB_COMPONENT_EVENT_TYPE_OVERRIDES: Record<string, Record<string, string>>
...
     react: {
-      nav: "GoabAppFooterNavSection",
-      meta: "GoabAppFooterMetaSection",
+      nav: null,
+      meta: null,
     },
     angular: {
-      nav: "GoabAppFooterNavSection",
-      meta: "GoabAppFooterMetaSection",
+      nav: null,
```
vanessatran-ddi: `null` here works but  when I read it is  a bit difficult to understand unless we check the comment below at line 2605. I will suggest we do like this: 

```
const HIDE_SLOT = null;
```

Then at line 207 for example:
```
{
 nav: HIDE_SLOT, 
meta: HIDE_SLOT
}...
```

Then under the `getSlotOverride` function below, we can keep like now
Under line 2622 we can do like this: 
```
getSlotOverride(framework, name) !== HIDE_SLOT && 
```

It will be easier to read.

Spark450: Done, pushed ecfdbc804: added `const HIDE_SLOT = null` and swapped it in at the override entries (`nav: HIDE_SLOT`) and the filter check (`getSlotOverride(framework, name) !== HIDE_SLOT`).

Also trimmed both nearby comments since part of what they explained is now redundant with the name: the type-definition comment no longer needs to spell out "null hides the slot" (kept the why footer needs this at all), and the `getSlotOverride` comment no longer needs to spell out what null means (kept the undefined vs HIDE_SLOT distinction, since that's the part that explains why the lookup can't just be a `??`).

Regenerated all 71 components, no JSON changed, this is a pure readability refactor.

vanessatran-ddi (approved):

ArakTaiRoth (approved):

PR #4124, opened 2026-07-06
chore(#4007): upgrade astro to v6.4.8

### LINKED ISSUE #4007: Upgrade Astro
# Info

There was a Dependabot PR (#3938) created to update Astro, it was a major version upgrade. But it failed to build docs, likely due to an actual issue because of the major version. This is related to this raised [security issue](https://github.com/GovAlta/ui-components/security/dependabot/377).

## Acceptance Criteria:

1. `astro` upgraded to at least 6.3.8
2. `astrojs/mdx` upgraded to at least 5.0.6

Spark450:
## Summary
- Upgrades astro (5.18.1 to 6.4.8) and @astrojs/mdx (4.3.14 to 5.0.6) per the acceptance criteria, plus @astrojs/react (4.4.2 to 5.0.7) as a required companion bump for Astro 6.
- Migrates docs/src/content/config.ts to docs/src/content.config.ts using the loader based content layer API, since Astro 6 removed legacy content collections. Updates every callsite that read the old auto generated entry.slug to use entry.id, and swaps legacy entry.render() calls for the imported render(entry) function.
- Works around a real bug in @astrojs/react 5.0.7 where its SSR renderer entrypoint gets externalized by Vite instead of bundled, leaving an internal astro:react:opts virtual import unresolved and crashing the build. Fixed with a vite.resolve.alias pointing the bare specifier to its resolved absolute path.

## Test plan
- [x] npm run build (all nx targets, including docs:build) passes
- [x] docs' own test suite (34 tests) passes
- [x] Manually verified in browser preview: homepage, a component detail page (Properties/Examples/Usage guidelines/Accessibility tabs), and a nested /examples/workspace/dashboard page all render correctly

Fixes #4007


Spark450:
## Follow-up fix: work side menu (and 11 other components) not rendering on the docs site

After the Astro 6 upgrade, `<goa-work-side-menu>` and several other components rendered as empty inline elements on the docs site. Pushed a fix in 4370235.

### Root cause
Astro 6 does not apply `vite.resolve.alias` to imports inside hoisted `<script>` blocks. The `import "@abgov/web-components"` in `BaseLayout.astro` / `PreviewLayout.astro` therefore skipped the alias to the locally built `dist/libs/web-components/` and fell back to normal node resolution, silently bundling the published npm copy (v1.39.3) from `docs/node_modules`. That copy predates 12 components, so their custom elements were never registered:

- work side menu family (menu, item, group, notification panel/item)
- workspace layout
- scroll panel
- data grid
- push drawer (+ internal)
- app header navigation
- linear progress

Confirmed three ways: the built docs chunk was byte-identical to the published package's define list (81 defines vs 93 expected); removing the npm copy flipped resolution to the local build; and a debug resolver traced the client pipeline resolving the bare id into `docs/node_modules`, while island imports (`@abgov/react-components`) got alias-rewritten correctly.

### Fix
A small `enforce: "pre"` resolver plugin in `docs/astro.config.mjs` that pins `@abgov/web-components`, its `index.css`, and `@abgov/style` to the local `dist` build in every pipeline. The earlier alias-only approach was reverted since hoisted scripts never consult aliases.

### Verification
- Built bundle now has all 93 `customElements.define` calls (was 81).
- Browser confirms all 10 previously-missing elements register; Work Side Menu and Linear Progress pages render correctly.
- 216 pages build, 34/34 docs tests pass.

### Caveat for reviewers
This bug pattern is quietly dangerous: because the fallback is a valid but older copy of the same package, any newly added component will silently disappear from the docs if this resolution ever regresses, with no build error. A registry-level check (`customElements.get` across all `goa-*` tags on a page) catches it immediately, whereas a normal visual smoke test can pass by coincidence on pages whose components already exist in the stale published bundle. Worth keeping in mind for future dependency upgrades on this repo. If Astro later fixes hoisted-script alias handling, this plugin stays harmless: it just performs the same mapping earlier.

docs/astro.config.mjs
```diff
@@ -65,6 +68,28 @@ export default defineConfig({
   },
 
   vite: {
+    plugins: [
```
ArakTaiRoth: I could be wrong, but it's worth trying I think, because it simplifies things quite a bit. Can we remove this whole change, and instead in `BaseLayout.astro` and `PreviewLayout.astro`, where we `import '@abgov/web-component':`, instead have it import a relative file path to our local `dist/libs/web-components/index.js`? I'm pretty sure that would force it to load our local implementation, rather than a stale package file.

Spark450: Good call, this works and is simpler. Verified it end to end and switched to it in 5c8f8c3.

Root cause confirmed: Astro does not apply Vite `resolve.alias` to imports inside hoisted `<script>` blocks, so the bare `@abgov/web-components` there resolved to the published npm copy in `docs/node_modules` (missing 12 newer components) with no build error. Pointing the two hoisted imports (`BaseLayout`, `PreviewLayout`) at a relative path into `dist/libs/web-components/index.js` bypasses aliasing and always loads the local build. The built docs bundle now registers all 93 components (was 81), and the work side menu / preview pages render.

Removed the whole plugin. Two things beyond your suggestion:
- Found and deleted two dead files that carried the same bare import and were never referenced (`Layout.astro`, `load-web-components.ts`), so the trap does not live on in unused code.
- Added a small guard test (`docs/src/scripts/web-components-import.test.ts`) that fails if any docs source file reintroduces the bare specifier, since the relative-path convention is now the only thing keeping the local build wired up.

The frontmatter `@abgov/web-components/index.css` imports keep using the existing alias (Astro honours aliases outside hoisted scripts), so those are untouched.

docs/src/scripts/web-components-import.test.ts
ArakTaiRoth, line 1: This test, and it's corresponding file don't actually serve a purpose. These aren't even run at the moment in our PR workflow or our Build pipeline, so they can easily be bypassed without saying anything.

In addition, they're scanning ordinary Typescript files even though this is only problematic in Astro's hoisted scripts, and it checks spelling basically instead of component registration.

If we really want tests for this (I don't), do something small tied to either `BaseLayout.astro` or `PreviewLayout.astro`, and ensure actual component registration. Also, you would need to modify the base (not docs) `package.json` to ensure `test:pr` actually runs the docs tests, which it doesn't currently.

Spark450: The fix commit is now just your suggestion: the two relative-path imports, plus deleting `Layout.astro` and `load-web-components.ts`, which were unreferenced and carried the same bare import.

The built docs bundle registers all 93 components (was 81), and the work side menu and preview pages render.

docs/src/content.config.ts
```diff
@@ -1,12 +1,13 @@
 import { defineCollection, z } from "astro:content";
```
willcodeforcoffee, line 1: import `z`  is deprecated, here. You should change this to

```ts
import { defineCollection } from "astro:content";
import { z } from 'astro/zod';
```

ArakTaiRoth (approved):

willcodeforcoffee (approved):
There is one deprecation left over but it can wait until the Astro 7 upgrade.

PR #4127, opened 2026-07-08
feat(#1351): Dropdown item slot for rich HTML content

### LINKED ISSUE #1351: [Feature Request]: Dropdown item slot
### Is your feature request related to a problem? Please describe

To be able to display dropdown item slots with multiple pieces of metadata within a dropdown menu

### Describe the solution you'd like

A flexible slot "area" where different types of formatted text, icons, and other assets can be placed in a way that makes sense for a variety of given contexts.

Requirements:
-Any/all text included in the slot can be filterable for the combo box version
-Multiple line support

### Provide evidence this is a needed component

image
image

### Describe alternatives you've considered

_No response_

### Do you have anything already created for this that we can use?

yes

### Are you currently using this proposal inside your own service

no

### Are you able to assist to bring the feature to reality?

no

### Additional context

_No response_

Spark450:
## Summary
- Adds a slot to goa-dropdown-item so items can contain arbitrary HTML (multi line text, icons, badges) that renders inside the dropdown menu. Implemented with dynamic named slots: each rich item is assigned to an option-{value} slot rendered inside its menu li, so consumer markup (and framework bindings) stay in the light DOM.
- Works with the filterable property (combobox). All text in the slotted content is filterable by default, built from individual text nodes so words in adjacent elements stay separated. The filter property replaces the slot text when provided.
- The label property is shown in the dropdown input when a slotted item is selected. Label only items keep the existing text rendering path.
- React wrapper (GoabDropdownItem) now accepts children; Angular wrapper projects content via ng-content.
- Slot assignment happens on the parent side after the mount relay is received. An element assigned to a slot name that does not exist yet has no slot in its event path, so its relayed messages would never reach the dropdown.
- Slot assignment walks up from the goa-dropdown-item to the ancestor that is a direct child of goa-dropdown before setting the slot attribute. Framework wrappers like Angular's goab-dropdown-item nest the custom element one level deeper, and slot assignment only honors direct children of the shadow host. Verified in the Angular playground.

## Filter matching change
Raised in review. `isFilterMatch` used to take the first non-empty of `filter`, `label`, `value`, so supplying `filter` stopped the label from being searched at all. It now matches against the filter text **and** the visible text (the label, or the value when no label is set).

| Item | Before | After |
| --- | --- | --- |
| `value` only | matches value | matches value (unchanged) |
| `value` + `label` | matches label | matches label (unchanged) |
| `value` + `label` + `filter` | filter only | filter and label |

This is additive rather than breaking. `value` stays searchable because `label` is optional and the value is the visible text when no label is set (items like `<GoabDropdownItem value="Airdrie" />`, including the existing filterable example on the docs page). The one behavior change: an item using `filter` to deliberately narrow matching will now also match its label.

## Documentation
- Two examples on the Dropdown base component page (acceptance criteria item 3): "Rich item content" and "Filterable rich item content", each covering React, Angular (reactive forms and ngModel) and web components.
- Guidance: native ignores slotted content and falls back to the label.
- Guidance: a filterable dropdown searches the label and the content text, but not text inside a nested component's shadow DOM (a Badge for example); `filter` adds those terms.
- Guidance: don't put interactive elements in item content. Each item is a `role="option"` inside a `role="listbox"`, the menu preventDefaults `mousedown`, and the li click handler selects the item and returns focus to the input, so a nested button or link is unreachable by keyboard and its own click never fires. The issue originally asked for "icon + tooltip to indicate status", which is not achievable accessibly since `goa-tooltip` renders with `tabindex="0"`; that line of the acceptance criteria has been updated to "icon or badge", and the examples use a Badge for status.

## Known limitations
- Text rendered inside a nested component's own shadow DOM (for example GoabBadge content) is not part of the slot text and is not filterable. The filter property is the escape hatch for adding extra search terms.
- Native mode (native="true") ignores slotted content and falls back to the label, since a native select cannot contain HTML.
- Slot content is read once when the item mounts, matching existing item behavior.
- The `children` prop does not appear in the generated API table. `extract-api` only collects named slots, and a pure slot carrier prop is filed under a `default` slot that never renders, so `DropdownItem`'s bare `<slot />` falls into that gap. This is repo wide (Button, Badge, Callout and Text all report `slots: []`), so it is not addressed here. Tracked in #4136.

## Pre-existing issue surfaced by the docs example
The menu auto-width comes from `getLongestChildWidth`, which measures `option.label || (option.hasSlotContent ? option.filter : "") || option.value` in `ch`. A slotted item effectively always has a label (it is what shows in the input on selection), so `option.label ||` short-circuits and the slot text branch never runs. The menu is sized to the label and wider content clips: measured li clientWidth 185px against scrollWidth 201px, with `ul` set to `overflow-x: auto`, so it clips rather than wrapping. That also conflicts with the existing "don't truncate labels" guidance.

Not fixed here. There is no good text length proxy for multi line content: the label is too narrow, the joined slot text is too wide (longest line 22 characters, joined text 36), and `filter` is worse. A real fix means measuring rendered width after slot distribution, which wants its own issue and browser tests. Setting `width` on the dropdown works around it today (verified). Tracked in #4135.

## Test plan
- [x] 63 web component unit tests pass, including 6 new filter matching cases covering value only items, the filter property, the label alongside the filter property, and slot text replaced by an explicit filter
- [x] React wrapper tests pass, including a children passthrough test
- [x] 277 Angular tests pass, including an ng-content projection test
- [x] 42 React browser tests pass across Chromium and Firefox, including 2 that verify real shadow DOM slot distribution: rich content renders inside the menu li and selection works, plus a regression test for items nested inside a wrapper element (the Angular case)
- [x] The two new filter matching cases that assert the new behavior fail against the previous implementation, and the cases covering existing behavior pass against both, confirming the change is additive
- [x] Both new documentation examples verified rendering in the docs site: rich content renders in the open menu with name, role, location and status badge
- [x] Verified in both playgrounds (React features/1351 and Angular features/1351): rich items render in the open menu; selection fires the change event and shows the label in the input; typing slot only text (for example "calg", "super") filters correctly; keyboard navigation and aria-activedescendant work on rich items; native mode (Test 4) shows plain labels in the select and leaks no rich content into the page

## Steps needed to test
1. npm run build, then npm run serve:prs:react (or npm run serve:prs:angular)
2. Open Features > 1351 Dropdown item slot in the side menu
3. Test 1: open the base dropdown, confirm multi line items with badges render, select one and confirm the label appears in the input
4. Test 2: type a name, role, or location in the combobox and confirm filtering matches slot text
5. Test 3: confirm plain items work alongside rich items, and that the filter property replaces the slot text while the label stays searchable ("preferred" matches, "Sarah" matches, "Edmonton" does not)
6. Test 4: confirm the native select shows plain labels for rich items, selection works, and no rich content leaks into the page
7. Docs: npm run dev in docs/, open Components > Dropdown and check the "Rich item content" and "Filterable rich item content" configurations

Fixes #1351

Once this passes code review, notify @twjeffery so the Figma assets can be prepared for release (acceptance criteria item 5).

### DIFF
--- a/libs/angular-components/src/lib/components/dropdown-item/dropdown-item.ts
+++ b/libs/angular-components/src/lib/components/dropdown-item/dropdown-item.ts
@@ -20,6 +20,7 @@ import { GoabDropdownItemMountType } from "@abgov/ui-components-common";
       [attr.name]="name"
       [attr.mount]="mountType"
     >
+      <ng-content />
     </goa-dropdown-item>
   }`,
   schemas: [CUSTOM_ELEMENTS_SCHEMA],
@@ -30,7 +31,7 @@ export class GoabDropdownItem implements OnInit {
 
   /** The value submitted when this item is selected. */
   @Input() value?: string;
-  /** Text used to filter and match this item in typeahead search. */
+  /** Additional text used to match this item in typeahead search, alongside the label. Defaults to the projected content's text. */
   @Input() filter?: string;
   /** Display label for the dropdown item. */
   @Input() label?: string;
--- a/libs/angular-components/src/lib/components/dropdown-item/dropdown-item.ts
+++ b/libs/angular-components/src/lib/components/dropdown-item/dropdown-item.ts
@@ -31,7 +31,7 @@ export class GoabDropdownItem implements OnInit {
 
   /** The value submitted when this item is selected. */
   @Input() value?: string;
-  /** Additional text used to match this item in typeahead search, alongside the label. Defaults to the projected content's text. */
+  /** Rich item content. On selection, `label` is shown and `filter` defaults to the content's text. */
   @Input() filter?: string;
   /** Display label for the dropdown item. */
   @Input() label?: string;
--- a/libs/react-components/src/lib/dropdown/dropdown-item.tsx
+++ b/libs/react-components/src/lib/dropdown/dropdown-item.tsx
@@ -1,4 +1,4 @@
-import { useEffect } from "react";
+import { ReactNode, useEffect } from "react";
 import { GoabDropdownItemMountType } from "@abgov/ui-components-common";
 
 interface WCProps {
@@ -25,12 +25,14 @@ export interface GoabDropdownItemProps {
   value: string;
   /** Display label for the dropdown item. */
   label?: string;
-  /** Text used to filter and match this item in typeahead search. */
+  /** Additional text used to match this item in typeahead search, alongside the label. Defaults to the children's text. */
   filter?: string;
   /** Sets a data-testid attribute for automated testing. */
   testId?: string;
   /** Controls how the item is registered with the parent dropdown. */
   mountType?: GoabDropdownItemMountType;
+  /** Rich content rendered for this item in the dropdown menu. When set, `label` is shown in the dropdown input on selection and `filter` defaults to the content's text. */
+  children?: ReactNode;
   /** @deprecated */
   name?: string;
 }
@@ -51,6 +53,7 @@ export function GoabDropdownItem({
   filter,
   name,
   mountType = "append",
+  children,
 }: GoabDropdownItemProps) {
   return (
     <goa-dropdown-item
@@ -59,6 +62,8 @@ export function GoabDropdownItem({
       filter={filter}
       name={name}
       mount={mountType}
-    />
+    >
+      {children}
+    </goa-dropdown-item>
   );
 }
--- a/libs/react-components/src/lib/dropdown/dropdown-item.tsx
+++ b/libs/react-components/src/lib/dropdown/dropdown-item.tsx
@@ -25,7 +25,7 @@ export interface GoabDropdownItemProps {
   value: string;
   /** Display label for the dropdown item. */
   label?: string;
-  /** Additional text used to match this item in typeahead search, alongside the label. Defaults to the children's text. */
+  /** Rich item content. On selection, `label` is shown and `filter` defaults to the content's text. */
   filter?: string;
   /** Sets a data-testid attribute for automated testing. */
   testId?: string;
--- a/libs/web-components/src/components/dropdown/Dropdown.svelte
+++ b/libs/web-components/src/components/dropdown/Dropdown.svelte
@@ -306,6 +306,10 @@
     // send message back to child that contains a reference to this component
     relay(detail.el, "dropdown:bind", { el: _rootEl });
 
+    if (detail.hasSlotContent) {
+      assignOptionSlot(detail);
+    }
+
     // ensure bind only runs once for all children
     if (_bindTimeoutId) {
       clearTimeout(_bindTimeoutId);
@@ -320,6 +324,27 @@
         }
       }
     }, 1);
+  }
+
+  /**
+   * Assign the slot after mount. For wrapped items, apply the slot attribute
+   * to the wrapper element directly inside the dropdown.
+   * @param detail
+   */
+  function assignOptionSlot(detail: DropdownItemMountedRelayDetail) {
+    const dropdownRoot = _rootEl.getRootNode();
+    const itemRoot = detail.el.getRootNode();
+    if (!(dropdownRoot instanceof ShadowRoot) || !(itemRoot instanceof ShadowRoot)) {
+      return;
+    }
+
+    const dropdownHost = dropdownRoot.host;
+    let slotTarget: Element | null = itemRoot.host as Element;
+    while (slotTarget && slotTarget.parentElement !== dropdownHost) {
+      slotTarget = slotTarget.parentElement;
+    }
+
+    slotTarget?.setAttribute("slot", `option-${detail.value}`);
   }
 
   /**
@@ -353,7 +378,12 @@
     // set width to longest item
     const optionsWidth = Math.max(
       ...options.map((option: Option) => {
-        const label = `${option.label}` || `${option.value}` || "";
+        // slotted items fall back to their filter text, which carries the slot's text content
+        const label =
+          option.label ||
+          (option.hasSlotContent ? option.filter : "") ||
+          option.value ||
+          "";
         return label.length;
       }),
     );
@@ -471,19 +501,27 @@
     }
   }
 
+  // Match against the filter text and the text the user sees, which is the
+  // label, or the value when no label is set. Both are searched so that adding
+  // a filter supplies extra search terms rather than replacing the visible text.
   function isFilterMatch(option: Option, filter: string, partialMatch = true) {
     // empty string matches all
     if (filter.length === 0) return true;
 
-    let value = option.filter || option.label || option.value;
-    value = value.toLowerCase();
+    const targets = [option.filter, option.label || option.value].filter(
+      (target) => target !== "",
+    );
     filter = filter.toLowerCase().trim();
 
-    if (!partialMatch) {
-      return value === filter;
-    }
+    return targets.some((target) => {
+      const value = target.toLowerCase();
 
-    return value.startsWith(filter) || value.includes(" " + filter);
+      if (!partialMatch) {
+        return value === filter;
+      }
+
+      return value.startsWith(filter) || value.includes(" " + filter);
+    });
   }
 
   // update the value show to the user in the <input> element
@@ -940,7 +978,11 @@
               _inputEl?.focus();
             }}
           >
-            {option.label || option.value}
+            {#if option.hasSlotContent}
+              <svelte:element this={"slot"} name={`option-${option.value}`} />
+            {:else}
+              {option.label || option.value}
+            {/if}
           </li>
         {:else}
           {#if _filterable}
--- a/libs/web-components/src/components/dropdown/DropdownItem.svelte
+++ b/libs/web-components/src/components/dropdown/DropdownItem.svelte
@@ -11,6 +11,7 @@
     filter: string;
     value: string;
     label: string;
+    hasSlotContent: boolean;
     mountType: DropdownItemMountType;
   }
   export type DropdownItemDestroyRelayDetail = {
@@ -23,6 +24,7 @@
     label: string;
     value: string;
     filter: string;
+    hasSlotContent: boolean;
     mountType: DropdownItemMountType;
   };
 </script>
@@ -33,7 +35,7 @@
 
   // Props
 
-  /** Text used to filter and match this item in typeahead search. */
+  /** Additional text used to match this item in typeahead search, alongside the label. Defaults to the slotted content's text. */
   export let filter: string = "";
   /** Display label for the dropdown item. */
   export let label: string = "";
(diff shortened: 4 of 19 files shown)


libs/web-components/src/components/dropdown/Dropdown.svelte
ArakTaiRoth, line 1: I wasn't aware this was how our actual filter worked. That if `filter` is supplied, `label` and `value` isn't used to match at all. I'm pretty sure we should change how our filtering actually works. I think if `filter` is supplied, it should take precedence, but the `label` should also be allowed to be searched, and `value` shouldn't be used at all.

I know that's not the point of this story, but this is a good place to do this change, as with Item slots in Dropdown, filtering using label text is much much harder.

libs/angular-components/src/lib/components/dropdown-item/dropdown-item.ts
```diff
@@ -30,9 +31,9 @@ export class GoabDropdownItem implements OnInit {
 
   /** The value submitted when this item is selected. */
   @Input() value?: string;
-  /** Text used to filter and match this item in typeahead search. */
+  /** Text used to filter and match this item in typeahead search. Defaults to the text of the projected content when provided. */
```
ArakTaiRoth: This change to the comment isn't needed, the property didn't change. And this is wrong in addition.

libs/angular-components/src/lib/components/dropdown-item/dropdown-item.ts
```diff
@@ -30,9 +31,9 @@ export class GoabDropdownItem implements OnInit {
...
   @Input() filter?: string;
-  /** Display label for the dropdown item. */
+  /** Display label for the dropdown item. Shown in the dropdown input when an item with projected content is selected. */
```
ArakTaiRoth: This change to the comment isn't needed, the property didn't change.

libs/react-components/src/lib/dropdown/dropdown-item.tsx
```diff
@@ -23,14 +23,16 @@ declare module "react" {
 export interface GoabDropdownItemProps {
   /** @required The value submitted when this item is selected. */
   value: string;
-  /** Display label for the dropdown item. */
+  /** Display label for the dropdown item. Shown in the dropdown input when an item with children is selected. */
```
ArakTaiRoth: This change to the comment isn't needed, the property didn't change.

libs/react-components/src/lib/dropdown/dropdown-item.tsx
```diff
@@ -23,14 +23,16 @@ declare module "react" {
...
   label?: string;
-  /** Text used to filter and match this item in typeahead search. */
+  /** Text used to filter and match this item in typeahead search. Defaults to the text of the children when provided. */
```
ArakTaiRoth: This change to the comment isn't needed, the property didn't change. And this is wrong anyways.

libs/web-components/src/components/dropdown/Dropdown.svelte
```diff
@@ -322,6 +326,36 @@
     }, 1);
   }
 
+  /**
```
ArakTaiRoth, line 329: This whole comment can be reduced to something like:
```
/**
   * Assign the slot after mount. For wrapped items, apply the slot attribute
   * to the wrapper element directly inside the dropdown.
   * @param detail
   */
```

libs/web-components/src/components/dropdown/DropdownItem.svelte
```diff
@@ -33,9 +35,9 @@
 
   // Props
 
-  /** Text used to filter and match this item in typeahead search. */
+  /** Text used to filter and match this item in typeahead search. Defaults to the text of the slotted content when provided. */
```
ArakTaiRoth: Remove this comment change

libs/web-components/src/components/dropdown/DropdownItem.svelte
```diff
@@ -33,9 +35,9 @@
...
   export let filter: string = "";
-  /** Display label for the dropdown item. */
+  /** Display label for the dropdown item. Shown in the dropdown input when a slotted item is selected. */
```
ArakTaiRoth: Remove this comment change

libs/web-components/src/components/dropdown/DropdownItem.svelte
```diff
@@ -48,14 +50,59 @@
...
+        value,
+        label,
+        hasSlotContent,
+        mountType: mount,
+      },
       { bubbles: true, timeout: 10 },
       );
   });
 
+  // Slotted children live in the host's light DOM; their text makes the item
```
ArakTaiRoth: This could be changed to something easier to understand:
```
// Use the slotted content's text as the default filter value. Separate text
// nodes with spaces so words in adjacent elements do not run together.
```

ArakTaiRoth (commented):
There's no documentation added to any of this. No examples created, no notice that this doesn't work for the `native` Dropdown type. The story itself explicitly calls for at least 2 examples created in the base component page, let alone any further examples in the Examples tab.

The only change to documentation is unnecessary comment changes. Indeed, there's not even a change indicating that DropdownItem can even accept a slot now.

Spark450:
## Documentation

You were right that the playground pages are not documentation. Added two examples to the Dropdown base component page, which is what AC #3 asks for:

- **Rich item content** (base dropdown)
- **Filterable rich item content** (uses `filterable`)

Both cover React, Angular (reactive forms and ngModel) and web components, matching the existing configs.

Also added three guidance entries:

- `native` renders a native `select`, which cannot contain HTML, so items fall back to their label.
- A filterable dropdown searches the label and the content text, but not text inside a nested component's shadow DOM (a Badge, for example). `filter` is the escape hatch for those terms.
- Don't put interactive elements in item content. Each item is a `role="option"` in a `role="listbox"`, the menu preventDefaults `mousedown`, and the `li` click handler selects the item and returns focus to the input. A nested button or link is unreachable by keyboard and its own click never fires. This ruled out the "icon + tooltip to indicate status" wording in the original issue, since `goa-tooltip` renders with `tabindex="0"`. I've updated that line of the AC to "icon or badge", which is what was actually meant. The examples use a Badge for status.

On "not even a change indicating that DropdownItem can even accept a slot": agreed the API table does not show it, but that turned out to be repo-wide rather than specific to this PR. `extract-api` only collects **named** slots (`<slot name="...">`), and a prop typed as a pure slot carrier (`children?: ReactNode`) gets pulled out of `props` and filed under a `"default"` slot that then never renders. `DropdownItem.svelte` has a bare `<slot />`, so it lands in that gap. Button, Badge, Callout and Text all take `children` and all report `slots: []` today. Teaching the extractor about default slots would rewrite all 71 generated files, so I left it alone and let the examples carry it. Opened #4136 for that.

## Filtering

Made the change. `isFilterMatch` now matches against the filter text **and** the visible text, instead of the old `filter || label || value` chain that stopped at the first non-empty value.

One deviation from what you asked: I kept `value` searchable, because dropping it entirely would be a breaking change rather than a cleanup. `label` is optional, and when it is absent the **value is what the user sees** (`setDisplayedValue` is `label || value`). Items declared as `<GoabDropdownItem value="Airdrie" />` with no label are the common case, including the `filterable` example already on the docs page. Remove `value` from matching and every one of those becomes unfindable: you type the exact text in front of you and get "No matches found", with no error and no migration path.

So the rule is now "you can always search what you see, and `filter` adds extra terms":

| Item | Before | After |
| --- | --- | --- |
| `value` only | matches value | matches value (unchanged) |
| `value` + `label` | matches label | matches label (unchanged) |
| `value` + `label` + `filter` | filter **only** | filter **and** label |

That makes it purely additive. I confirmed this empirically: the new test cases covering the first two rows pass against both the old and the new implementation, and only the `filter` + `label` case fails on the old code. The one behavior change is that an item using `filter` to deliberately *narrow* matching will now also match its label.

`value` is still ignored once a `label` is set, which matches the old precedence.

## Comments and JSDoc

Applied all of these. Reverted the `label` JSDoc on all three wrappers since the property genuinely did not change. `filter`'s description does change, since its behavior changed above, so it now reads "Additional text used to match this item in typeahead search, alongside the label."

## One thing worth flagging (pre-existing, not fixed here)

The new docs example surfaces a sizing bug that predates this change. The menu auto-width comes from `getLongestChildWidth`, which measures `option.label || (option.hasSlotContent ? option.filter : "") || option.value` in `ch`. For a slotted item the label is effectively always set (it is what shows in the input on selection), so `option.label ||` short-circuits and the slot-text branch is dead. The menu gets sized to the label, and wider content clips: I measured `li` clientWidth 185px against scrollWidth 201px, and `ul` is `overflow-x: auto`, so it clips rather than wrapping. That also conflicts with the existing "don't truncate labels" guidance.

I left it alone deliberately, because there is no good text-length proxy for multi-line content: the label is too narrow, the full slot text is too wide (the longest *line* is 22 characters but the joined text is 36), and `filter` is worse. Doing it properly means measuring rendered width after slot distribution, which is a layout change that wants its own issue and its own browser tests. Setting `width` on the dropdown works around it today (verified). Opened #4135 for the real fix.

## Tests

- 63 web component unit tests, including 6 new filter matching cases
- 42 React browser tests across Chromium and Firefox
- 277 Angular tests

libs/web-components/src/components/dropdown/Dropdown.svelte
Spark450: Changed `isFilterMatch` to match against `filter` and the visible text (`label || value`), rather than the old first-non-empty-of-three chain. One adjustment: I kept `value` searchable rather than dropping it, since `label` is optional and `value` is what the user sees when it's unset (e.g. the existing `filterable` docs example uses value-only items). Dropping `value` entirely would make those unfindable. Full reasoning and a before/after table in the summary comment below.

libs/angular-components/src/lib/components/dropdown-item/dropdown-item.ts
Spark450: Fair on "wrong" -- fixed now that `filter`'s behavior actually changed (see the isFilterMatch thread). It now reads: "Additional text used to match this item in typeahead search, alongside the label. Defaults to the projected content's text."

libs/angular-components/src/lib/components/dropdown-item/dropdown-item.ts
Spark450: Reverted, `label`'s behavior didn't change.

libs/react-components/src/lib/dropdown/dropdown-item.tsx
Spark450: Reverted, `label`'s behavior didn't change.

libs/react-components/src/lib/dropdown/dropdown-item.tsx
Spark450: Fixed to match the new behavior, same as the Angular side: "Additional text used to match this item in typeahead search, alongside the label. Defaults to the children's text."

libs/web-components/src/components/dropdown/Dropdown.svelte
Spark450: Used your wording verbatim.

libs/web-components/src/components/dropdown/DropdownItem.svelte
Spark450: Removed.

libs/web-components/src/components/dropdown/DropdownItem.svelte
Spark450: Removed.

libs/web-components/src/components/dropdown/DropdownItem.svelte
Spark450: Used your wording verbatim.

libs/react-components/src/lib/dropdown/dropdown-item.tsx
ArakTaiRoth: @Spark450 I would shorten what you had to something like `/** Rich item content. On selection, `label` is shown and `filter` defaults to the content's text. */`. Having it be too wordy makes things more awkward in the documentation and can be more confusing.

libs/angular-components/src/lib/components/dropdown-item/dropdown-item.ts
ArakTaiRoth: @Spark450 I would shorten what you had to something like `/** Rich item content. On selection, `label` is shown and `filter` defaults to the content's text. */`. Having it be too wordy makes things more awkward in the documentation and can be more confusing.

Spark450: Done, updated to your wording. Also pulled in your width="250px" fix from 35e65dcf2f during the rebase and re-verified in the browser: li scrollWidth now equals clientWidth on the Rich item content example, no more clipping.

libs/react-components/src/lib/dropdown/dropdown-item.tsx
Spark450: Done, updated to your wording here too.

Spark450:
@ArakTaiRoth all your comments are addressed now, including the two follow-ups on the filter JSDoc. Ready for another look whenever you get a chance.

vanessatran-ddi (approved):

ArakTaiRoth (approved):

PR #4128, opened 2026-07-09
feat(#3858): add tertiary type to Icon Button

### LINKED ISSUE #3858: Icon Button new variant "tertiary
# Info

Feature requested by Matthew Tycholaz and confirmed by @twjeffery 

It's been requested to be able to have an icon button that features a border (or box-shadow). This border (or box-shadow) would use design-tokens so it was customizable.

This would help mark icon buttons as potentially more interactive for users.

## Acceptance Criteria:

1. New `variant` "outline"
2. Designs added in the [Figma component library](https://www.figma.com/design/Jpy1Ea5qglwnp1SgGnagY9/%E2%9D%96-Component-library?node-id=56969-757260&t=FGtXz61NeqApmwMo-4) as a new type

Image

twjeffery:
Closes #3858

# Before
image

- Icon buttons have no bordered option to mark them as interactive.
- Disabled icon buttons show the default arrow cursor.

# After
image

- New `type` prop on Icon Button: `"default" | "tertiary"` (defaults to `"default"`, fully backward-compatible). Mirrors Button's existing `type`/`variant` split and the Figma component set (Type=Tertiary), so the bordered look crosses with every colour variant and size.
- `type="tertiary"` renders a 1px border at rest — greyscale-200, or emergency-light with `variant="destructive"` — via new component tokens (`--goa-icon-button-tertiary-*`) with global-token fallbacks, so nothing depends on a tokens release. The border is an inset box-shadow, so the button footprint is identical in every state (no 1px shift on hover).
- Per the Figma states, the border only shows at rest: hover, focus, and disabled fall back to the existing variant styling.
- All disabled icon buttons (any type) now show `cursor: not-allowed`. This required removing `pointer-events: none` from the disabled state; hover styling is now suppressed per-variant with `:not(:disabled)`. **Deliberate behaviour change that comes with this:** disabled icon buttons now show their `title` tooltip on hover.
- Icon Button's transitions now use the global motion tokens (`--goa-motion-duration-short-3/-4`, `--goa-motion-curve-expressive`/`-transform`) with the previous literal values as fallbacks.
- Companion tokens PR: GovAlta/design-tokens#166.

### Notes for reviewers
- React wrapper defaults `type="default"`, so the attribute now always renders on `<goa-icon-button>` (same pattern as `variant="color"`). Angular leaves it undefined, same as it does `variant`.
- `.tertiary:focus-visible` re-asserts the focus ring because `.tertiary.destructive` would otherwise outrank `button:focus-visible`; the tie with `.tertiary.destructive` is carried by source order (verified in the compiled bundle).
- Drive-by (flagged deliberately): the React spec's variant loop was missing `"light"` — added, since this PR already extends that spec. One word, no behaviour change.
- Drive-by #2: `.vitest-attachments/` (Vitest browser-mode failure screenshots) wasn't gitignored and briefly got committed here before being removed — added the ignore line so it can't happen to anyone else.
- A new browser test covers the disabled state end-to-end: computed `not-allowed` cursor, and no `action`/`onClick` dispatch on click.
- Motion tokens are v2-only, so the transition keeps the old literals as fallbacks (v1 consumers keep today's exact behaviour). Nearest-token call: the 0.2s fades became `short-4` (180ms); the 0.1s press transform is `short-3` (100ms, exact) with `curve-expressive-transform` (exactly `ease-in-out`). Curve for fades follows ScrollPanel/Drawer precedent (`curve-expressive`).
- WorkSideMenu tooltip browser tests are flaky in headless Firefox (different cases fail run to run, pass in isolation; untouched by this diff).
- Please also review companion tokens PR: GovAlta/design-tokens#166

## Steps needed to test

1. `npm run serve:prs:react` → side menu → **3858 Icon Button Tertiary Type**
2. Walk the five test cases: type × variant matrix, light variant on dark background, sizes, interactive states (hover / keyboard focus — check the destructive focus ring — / disabled cursor), and default-type regression.
3. Compare against the Figma component set: https://www.figma.com/design/Jpy1Ea5qglwnp1SgGnagY9/%E2%9D%96-Component-library?node-id=56969-757260
4. Angular: the wrapper threads `type` to the same web component; attribute threading is covered by the Jest spec.

- [x] I have tested the functionality in both React and Angular


libs/web-components/src/components/icon-button/IconButton.html-data.json
ArakTaiRoth, line 1: Please remove this file, it's being removed in one of my PR's that hasn't gone through yet.

libs/web-components/src/components/icon-button/doc.md
ArakTaiRoth, line 1: Please remove this file, it's being removed in one of my PR's that hasn't gone through yet.

libs/web-components/src/components/icon-button/IconButton.html-data.json
twjeffery: Done. Merged dev in and this file went with it, both are out of the diff now.

libs/web-components/src/components/icon-button/doc.md
twjeffery: Done. Merged dev in and this file went with it, both are out of the diff now.

ArakTaiRoth (approved):

PR #4129, opened 2026-07-09
bug(#3762): Updated Slots for AppHeader

### LINKED ISSUE #3762: AppHeader examples using `slot` property

ArakTaiRoth:
# Before (the change)

Slots didn't work properly for AppHeader for either React or Angular, and the code to implement them wasn't inline with slot usage elsewhere.

# After (the change)

This is a pretty large change to how some slots (2, `utilities` and `navigation`) are handled. Essentially on the user facing end, they work the same as any of our other slots, you setup the property in the component and then output whatever you want to output.

On our end, they are handled far more differently. Basically in both React and Angular, essentially what's being done is that everything output by the user is being copied, and then pasted in the output but with the addition of adding a slot property to every element they use in their output. So, basically they can use code like this on their end:

```html
<ng-template #navigationTemplate>
  <a href="#">Dashboard</a>
  <goab-app-header-menu heading="Applications">...</goab-app-header-menu>
  <a href="#">Help</a>
</ng-template>

<goab-app-header [navigation]="navigationTemplate" />
```

But on the output side, you'll get code like this:

```html
<a slot="navigation">Dashboard</a>
<goab-app-header-menu slot="navigation">...</goab-app-header-menu>
<a slot="navigation">Help</a>
```

## Steps needed to test

There are PR files located in bugs/3762 for both React and Angular. They should touch most things, I think.

You also need to test and verify that all the updated examples and provided code works via copy and paste. And please verify the updated json files are accurate.


libs/angular-components/src/lib/components/app-header/app-header.ts
```diff
@@ -8,11 +8,15 @@ import {
   OnInit,
```
willcodeforcoffee, line 8: In my pr I renamed the Angular controls to be consistent with Svelte and React:

- libs/angular-components/src/lib/components/header/header.ts to
libs/angular-components/src/lib/components/app-header/app-header.ts

- libs/angular-components/src/lib/components/header-menu/header-menu.ts to
libs/angular-components/src/lib/components/app-header-menu/app-header-menu.ts

It doesn't _have_ to be done in this PR, but it makes sense since you're already changing these files.

willcodeforcoffee (changes requested):
Looks great! The only changes I would recommend are:
- to rename the Angular controls to `app-header` and `app-header-menu`
- run the `docs:build` task

libs/angular-components/src/lib/slotted-directive.ts
```diff
@@ -0,0 +1,51 @@
...
+  OnChanges,
+  SimpleChanges,
+  TemplateRef,
+  ViewContainerRef,
+  inject,
+} from "@angular/core";
+
+@Directive({
+  standalone: true,
+  selector: "[goabSlottedTemplate]",
```
vanessatran-ddi, line 14: `goabSlottedTemplate` reads more generic than its scope(only AppHeader `nav/utilties` use it, and `banner/phase` are also template), can we rename it to something like `goabAppHeaderSlotItems` with the file renamed to `app-header-slot-items-directive.ts` to match with other directive files, for example `value-directive.ts`

libs/angular-components/src/lib/components/header/header.ts
```diff
@@ -25,6 +29,28 @@ import {
       [attr.version]="version"
       (_menuClick)="_onMenuClick()"
     >
+      @if (this.banner) {
```
vanessatran-ddi: we do not need to use `this.banner` in angular template, we can `@if(banner)` directly (remove `this`)

libs/angular-components/src/lib/components/app-header-menu/app-header-menu.ts
```diff
@@ -26,19 +27,20 @@ import {
...
   @Input() leadingIcon?: GoabIconType;
   /** The menu heading text displayed as the dropdown trigger. */
   @Input() heading?: string;
   /** Sets a data-testid attribute for automated testing. */
   @Input() testId?: string;
-  /** Sets the slot name for the component. */
+  /** @deprecated Sets the slot name for the component. Deprecated: use AppHeader navigation TemplateRef instead. */
   @Input() slotName?: string;
 
   @HostBinding("attr.slot")
```
vanessatran-ddi, line 41: When I read this function, I was confused because why we need to read the binding attribute and return it, turned out we need this one in case it is put under `navigation` slot, for ex, under `goab-app-header` and mapped to the directive. can we add a comment here to explain it

libs/angular-components/src/lib/components/header/header.spec.ts
```diff
@@ -161,3 +221,114 @@ describe("GoabAppHeader without onMenuClick", () => {
...
+      imports: [GoabAppHeader, TestAppHeaderSlotsComponent],
+      schemas: [CUSTOM_ELEMENTS_SCHEMA],
+    }).compileComponents();
+
+    fixture = TestBed.createComponent(TestAppHeaderSlotsComponent);
+    fixture.detectChanges();
+    tick();
+    fixture.detectChanges();
+    tick();
+    fixture.detectChanges();
```
vanessatran-ddi: Can we add a comment to explain why we need to `tick()` 2 times

libs/angular-components/src/lib/slotted-directive.ts
ArakTaiRoth: @vanessatran-ddi The thing is, it is generic, it's not built to only work with AppHeader, it could be used anywhere in the future if we need it. So I'm actually more inclined to leave it named as is.

ArakTaiRoth: It could however be moved somewhere else, I'll look into that

libs/angular-components/src/lib/components/app-header-menu/app-header-menu.ts
ArakTaiRoth: @vanessatran-ddi I'm not sure what you're referring to with this comment? The line you commented on isn't a function and already has a clear comment?

libs/angular-components/src/lib/components/app-header/app-header.ts
ArakTaiRoth: @willcodeforcoffee I made that file name change

libs/angular-components/src/lib/slotted-directive.ts
ArakTaiRoth: @vanessatran-ddi It's been moved and renamed

libs/angular-components/src/lib/components/header/header.ts
ArakTaiRoth: @vanessatran-ddi Wild, `this.` is just so much more expected, but cool, change made

libs/angular-components/src/lib/components/header/header.spec.ts
ArakTaiRoth: @vanessatran-ddi Good catch, this was a hold over from when I was having a hard time getting the tests to pass because of the conditional rendering. I've changed it to how it should be now.

libs/angular-components/src/lib/components/app-header/app-header.ts
ArakTaiRoth: @willcodeforcoffee I also fixed the generated files (the issue was because of my personal machine changing the `/` to `\\`)

libs/angular-components/src/lib/slotted-directive.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  @Input() goabSlottedTemplate?: TemplateRef<unknown>;
+  @Input() slotName?: string;
+
+  ngOnChanges(changes: SimpleChanges): void {
+    if (changes["goabSlottedTemplate"]) {
+      this.viewContainer.clear();
+      this.view = undefined;
+
+      // This creates an embedded view of the entire slot
```
vanessatran-ddi: This comment just restates the code. I think we should drop it, or explain why instead

libs/angular-components/src/lib/slotted-directive.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    this.applySlotToRootNodes();
+  }
+
+  private applySlotToRootNodes(): void {
+    if (!this.view || !this.slotName) {
+      return;
+    }
+
+    // This goes through each element in the created view and applies the slot property to everything
```
vanessatran-ddi: Same here. A \"why\" would help, for example we set `slot` per root node so AppHeader overflow logic can count each item

libs/angular-components/src/lib/components/app-header-menu/app-header-menu.ts
vanessatran-ddi: I mean the `hostSlot` getter on lines 41-44. My ask is the `?? this.el.nativeElement.getAttribute(slot)` fallback is not obvious. It is there so that when this menu is placed inside the `navigation` template and GoabSlottedTemplate sets slot=`navigation` on the host, the HostBinding reads that value back instead of overwriting it with `null`. A comment explaining this would save the next reader the same confusion I had.

vanessatran-ddi (approved):

ArakTaiRoth:
@vanessatran-ddi I resolved all of your comment issues. If you could resolve those, that would be great.

@willcodeforcoffee Can you please re-review?

vanessatran-ddi (approved):
Looks good!

willcodeforcoffee (approved):

PR #4131, opened 2026-07-09
fix(#4130): retry hover in work side menu tooltip browser tests

### LINKED ISSUE #4130: Work Side Menu tooltip browser tests flaky in CI (react-headless firefox)
## Summary

\`libs/react-components/specs/work-side-menu.browser.spec.tsx\` is intermittently failing in CI under the \`react-headless (firefox)\` test project, on the tooltip show/hide assertions:

\`\`\`
AssertionError: expected false to be true // Object.is equality
❯ libs/react-components/specs/work-side-menu.browser.spec.tsx:236:54
  expect(tooltipEl?.classList.contains("show")).toBe(true);
\`\`\`

This happens on the "should show and hide tooltip for toggle button", "for item", and "for group" cases.

## Evidence this is a shared CI issue, not per-PR

Observed failing on the same assertion, in the same time window (2026-07-09 ~19:30-19:40 UTC), across three unrelated PRs whose diffs don't touch Work Side Menu or its tooltip:

- PR #4122 (mark/1666-focus-blur-events)
- PR #4129 (dustin/3762)
- PR #4121

PR builds that ran earlier the same day passed cleanly, so this looks like it started affecting the whole CI fleet partway through the day rather than being caused by a specific PR's changes.

## Context

fix(#4082) (commit 591c5675b, "wait on real tooltip state in work side menu browser tests") already attempted to fix flakiness in this same test file and merged cleanly with a passing CI run on \`dev\`. Despite that fix being present, the failures above are still occurring for branches rebased on top of it. Worth checking whether the fix only covers some of the flake's causes, or whether something separate (Firefox version bump, runner timing, animation duration) has reintroduced it.

## Suggested next steps

- Reproduce locally with repeated runs (\`npx vitest run libs/react-components/specs/work-side-menu.browser.spec.tsx --project react-headless\` in a loop) to gauge flake rate.
- Check if the tooltip's CSS transition/animation timing lines up with the \`waitFor\` timeout used in the test.
- Consider whether recent Firefox or Playwright version bumps in devDependencies changed timing behavior.

Spark450:
# Before (the change)

The three Work Side Menu tooltip browser tests drove the tooltip with real pointer hovers. In CI under react-headless (firefox) they failed intermittently on the show assertion. Evidence from the failed runs shows headless Firefox can stop delivering synthesized pointer boundary events for around 10 seconds at a time: a first attempt that retried the full unhover/hover cycle 3 times with 3s waits still never saw a single mouseenter reach the toggle button, while the same code passed on chromium and on an earlier identical run.

# After (the change)

The tests dispatch mouseenter/mouseleave directly at the elements that own the component's listeners, the same approach tooltip.browser.spec.tsx already uses for the Tooltip component. Specifically:

- item: mouseenter on the item's root div (carries the data-testid), mouseleave on the menu's primary slot wrapper
- group: mouseenter on the group's inner details element, mouseleave on the primary slot wrapper
- toggle: mouseenter and mouseleave on the toggle button itself

Each attempt waits for the web component to upgrade before dispatching (the target element does not exist until then), and dispatches exactly once per attempt because every mouseenter re-arms the component's 300ms show debounce. The tooltip element is re-queried on every waitFor poll instead of being captured once.

No component code changed, this is test-only.

Fixes #4130

## Steps needed to test

1. Run the browser spec repeatedly with retries disabled to gauge stability: `npx vitest run libs/react-components/specs/work-side-menu.browser.spec.tsx --project=react-headless --retry=0`
2. Verified locally: 10/10 consecutive runs pass (chromium and firefox headless instances, 26 tests each run).

### DIFF
--- a/libs/react-components/specs/work-side-menu.browser.spec.tsx
+++ b/libs/react-components/specs/work-side-menu.browser.spec.tsx
@@ -10,16 +10,9 @@ import { page, type Locator } from "@vitest/browser/context";
 // assertions extra headroom to absorb the debounce plus event latency.
 const TOOLTIP_WAIT = { timeout: 3000 };
 
-// A single hover is not enough to guarantee the tooltip shows, and waiting
-// longer can't recover because waitFor never re-hovers. Two failure modes
-// were hit in CI (headless Firefox): the pointer move can land before the
-// component's hover listeners are attached, and a vitest retry re-renders
-// fresh DOM under a pointer that hasn't moved, so the browser never emits a
-// new mouseenter. Each attempt therefore unhovers first (unhover targets the
-// page body, forcing real pointer movement) before hovering again, and the
-// tooltip element is re-queried on every poll so a null captured before the
-// web component upgraded isn't held for the whole wait. The tooltip tests
-// get a generous per-test timeout to cover the retried hovers.
+// Re-drive real pointer movement instead of only waiting longer. Each attempt
+// moves off the target before hovering again, and the tooltip is re-queried on
+// every poll so retries can observe newly rendered or upgraded DOM.
 const TOOLTIP_TEST_TIMEOUT = 20000;
 const HOVER_ATTEMPTS = 3;
 
--- a/libs/react-components/specs/work-side-menu.browser.spec.tsx
+++ b/libs/react-components/specs/work-side-menu.browser.spec.tsx
@@ -12,13 +12,8 @@ const TOOLTIP_WAIT = { timeout: 3000 };
 const TOOLTIP_TEST_TIMEOUT = 20000;
 const HOVER_ATTEMPTS = 3;
 
-// Real pointer hovers are not reliable here: CI headless Firefox can stop
-// delivering synthesized boundary events for many seconds at a time (#4130),
-// long enough to outlast any retry budget. Drive the tooltip by dispatching
-// mouseenter/mouseleave at the elements that own the component's listeners
-// instead, the same approach as tooltip.browser.spec.tsx. The dispatch is
-// retried because it is fire-and-forget: an event sent before the web
-// component has upgraded and attached its listeners is silently lost.
+// Drive the component's hover handlers directly. This function verifies tooltip
+// behaviour after hover events, not browser pointer movement.
 async function waitForTooltipToShow(
   getHoverTarget: () => Element | null,
   getTooltip: () => HTMLElement | null,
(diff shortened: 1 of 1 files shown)


libs/react-components/specs/work-side-menu.browser.spec.tsx
```diff
@@ -3,13 +3,49 @@ import { useState } from "react";
...
 import { expect, describe, it, vi } from "vitest";
-import { page } from "@vitest/browser/context";
+import { page, type Locator } from "@vitest/browser/context";
 
 // The tooltip uses a 300ms show/hide debounce. On contended CI runners that
 // timer can be delayed past waitFor's 1000ms default, so give the tooltip
 // assertions extra headroom to absorb the debounce plus event latency.
 const TOOLTIP_WAIT = { timeout: 3000 };
 
+// A single hover is not enough to guarantee the tooltip shows, and waiting
```
ArakTaiRoth: Another comment comment. Can you shorten this to something like:
```
// Re-drive real pointer movement instead of only waiting longer. Each attempt
// moves off the target before hovering again, and the tooltip is re-queried on
// every poll so retries can observe newly rendered or upgraded DOM.
```

Simpler, more concise, and to the point

Spark450: Done in 54389f36d, thanks!

Spark450:
The first approach (retrying real pointer hovers) still failed in CI: the failing run shows the toggle test burned all 3 unhover/hover attempts (~10s) without a single mouseenter getting through, while the item and group tests in the same session only passed thanks to their retries. That points at headless Firefox intermittently not delivering synthesized pointer events at all, so no amount of pointer-level retrying is reliable.

Reworked to dispatch mouseenter/mouseleave at the listener-owning elements directly, which is the pattern tooltip.browser.spec.tsx already uses for the same reason. PR body updated to match.

libs/react-components/specs/work-side-menu.browser.spec.tsx
```diff
@@ -9,6 +9,49 @@ import { page } from "@vitest/browser/context";
 // timer can be delayed past waitFor's 1000ms default, so give the tooltip
 // assertions extra headroom to absorb the debounce plus event latency.
 const TOOLTIP_WAIT = { timeout: 3000 };
+const TOOLTIP_TEST_TIMEOUT = 20000;
+const HOVER_ATTEMPTS = 3;
+
+// Real pointer hovers are not reliable here: CI headless Firefox can stop
```
ArakTaiRoth: More comments about comments. Change this comment to:
```
// Drive the component's hover handlers directly. This function verifies tooltip
// behaviour after hover events, not browser pointer movement.
```

libs/react-components/specs/work-side-menu.browser.spec.tsx
```diff
@@ -9,6 +9,49 @@ import { page } from "@vitest/browser/context";
...
+// instead, the same approach as tooltip.browser.spec.tsx. The dispatch is
+// retried because it is fire-and-forget: an event sent before the web
+// component has upgraded and attached its listeners is silently lost.
+async function waitForTooltipToShow(
+  getHoverTarget: () => Element | null,
+  getTooltip: () => HTMLElement | null,
+  label: string,
+) {
+  for (let attempt = 1; ; attempt++) {
+    // Wait for the web component to upgrade before dispatching: the target
```
ArakTaiRoth: Remove this comment, it isn't necessary after the above comment explains what the function does.

Spark450: Done in 81b19e78a, thanks!

vanessatran-ddi (commented):
It fixed for my local after clearing/reset nx

Image

vanessatran-ddi (approved):

ArakTaiRoth (approved):

PR #4133, opened 2026-07-11
fix(#3860): App Header menu aligns right when squeezed against the screen edge

### LINKED ISSUE #3860: App Header Menu aligned to the left instead of the right
# Info

If you shrink our Header down to mobile size, it will create a Menu link using our AppHeaderMenu component. This component naturally aligns the associated Popover to the left of the button. This means the Menu can only take up the rest of the space between the left of the button and the right edge of the screen, which often makes the menu appear not as wide as it should.

Image

## More Info

I believe this could be fixed with providing an `alignment` property to Popover, could be internal only if we want it to be. The options could be "left" and "right". Depending on the alignment given (default could be left, so as to not change anything elsewhere), it would change the properties `inset-inline-start` and `inset-inline-end` to anchor correspondingly.

## Acceptance Criteria:

1. Provide a way for the Popover created by AppHeaderMenu (specifically the mobile version created) to be aligned to the right of the button (and the screen) instead of the left, allowing it to take up its full width
2. Alignment to the right only occurs when there isn't space to align left as expected

twjeffery:
## Summary

When the app header narrows and its navigation overflows into a "More" menu near the right edge, the menu was opening to the right of its button, into the sliver of space before the screen edge. That squeezed it narrower than it should be, so a long item like "Organization settings" wrapped to two lines.

The Popover already had logic to open a menu leftward when there isn't room on the right, but it was comparing against the menu's already-squeezed width, so it almost never kicked in. It now compares against the width the menu actually wants, so a menu near the right edge opens leftward at its full width.

## What changed

- Popover: the right-align check now uses the width the menu would take with room (its max width) instead of the width it got squeezed into. One function, no new props or events.
- No React or Angular wrapper changes. This lives entirely in the Popover web component's positioning, so nothing changes in how the wrappers are used.

## One note

It compares against the menu's max width rather than measuring its exact content, because measuring means resizing the live popover, which isn't safe while it's open. The effect is that a menu very close to the right edge can open leftward even when its content would have just fit, which reads fine that close to an edge.

## Test plan

- Added a browser test that opens a menu near the right edge and checks it aligns to the right of its button (opens leftward). Fails on the old behavior, passes on the new.
- Full browser, unit, and Angular suites pass with no regressions.
- Playground: `bugs/3860` has a header that overflows into a "More" menu. Narrow to around 700px so "More" sits near the right edge, then open it.

### Before

image

### After

image
image
image

### DIFF
--- a/apps/prs/react/src/routes/bugs/bug3860.tsx
+++ b/apps/prs/react/src/routes/bugs/bug3860.tsx
@@ -1,0 +1,126 @@
+import {
+  GoabAppHeader,
+  GoabAppHeaderMenu,
+  GoabBlock,
+  GoabDetails,
+  GoabDivider,
+  GoabLink,
+  GoabText,
+} from "@abgov/react-components";
+
+const ISSUE_NUMBER = "3860";
+const ISSUE_TITLE = "App Header Menu aligned to the left instead of the right";
+
+export function Bug3860Route() {
+  return (
+    <div>
+      <GoabText tag="h1" mt="m" mb="m">
+        Bug #{ISSUE_NUMBER}: {ISSUE_TITLE}
+      </GoabText>
+
+      <GoabBlock gap="m" direction="column">
+        <GoabLink trailingIcon="open">
+          <a
+            href={`https://github.com/GovAlta/ui-components/issues/${ISSUE_NUMBER}`}
+            target="_blank"
+            rel="noopener"
+          >
+            View on GitHub
+          </a>
+        </GoabLink>
+
+        <GoabDetails heading="Issue description">
+          <GoabText tag="p" mb="s">
+            When the App Header collapses to mobile size, overflowing navigation is
+            collected into a "More" menu (an AppHeaderMenu). Its popover anchors to the
+            LEFT of the button, so when the button sits near the right edge of the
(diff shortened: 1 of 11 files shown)


apps/prs/react/src/routes/bugs/bug3860.tsx
ArakTaiRoth, line 1: Just a couple suggestions here:
1. I notice there's no Angular test. In theory, this is just CSS primarily, so there shouldn't be a difference, but whenever I've been surprised by something it's when I thought something shouldn't happen.
2. You modified `Popover.svelte`, that component is also internally used by `GoabMenuButton`, `GoabDatePicker`, `GoabDropdown`, and can also just be used itself by the public. Could you add tests for all of those components in here as well to confirm they still work as expected?

twjeffery: Added browser tests for MenuButton, DatePicker, Dropdown, and a plain Popover. The Angular tests can't check where a menu opens on screen, so I added an Angular playground page for this bug instead, and both playground pages now have a section where you can push each component against the screen edge. Good to review again.

ArakTaiRoth (approved):

Issue #3609, opened 2026-03-19
Modal: Content clipping in scroll area and insufficient actions padding


twjeffery:
## Description

Two issues with the modal component in V2:

### 1. Content clipping in scroll area
Content is getting clipped at the top and bottom of the scroll area. When modal content exceeds the available height, internal padding or overflow is cutting off content. (it looks like there is additional padding around all of the content: top, bottom, left, right)

### 2. Actions too close to footer edge
The action buttons (e.g. Cancel / Save) are too close to the bottom edge of the pinned footer area. Needs more padding between the buttons and the footer edge.

## Where to see it

Visible on the [Add another item in a modal](https://goa-design-2.netlify.app/components/modal/#tab-1) example on the docs site.

## Notes

Should be reviewed thoroughly for V2 to make sure styling is all correct, and perhaps how/when this broke.

## Screenshots
<img width="1196" height="827" alt="Image" src="https://github.com/user-attachments/assets/49bbd646-e243-42d5-98b9-fc9afa790647" />

https://github.com/user-attachments/assets/e24a6747-1f36-4cad-b646-c5e20b874cc8

bdfranck:
I started working on this then realized that the Modal is about to be completely refactored with https://github.com/GovAlta/ui-components/pull/3933. I'll wait until the work is merged before seeing if this is still an issue.

Spark450:
Checked in on the modal after #3933 It looks like issue #1 is resolved but there is still a problem with the spacing for the action buttons

PR #4147, opened 2026-07-17
test(#3609): add modal actions padding playground pages and test

Spark450:
# Before

The modal footer actions had uneven padding, 8px at the top against 32px at the bottom on desktop, so the Cancel and Save buttons looked cramped against the divider above them.

# After

The fix is a design token change rather than a component change. `modal-actions-padding` becomes an even 24px on all sides on desktop, and 16px on mobile. That's in GovAlta/design-tokens#168.

That fix is already live. It's in design-tokens 2.12.0, and this repo has been on 2.12.0 since #4184 merged on July 31.

This PR is now just the playground pages so the spacing can be checked by eye:

- React page at `bugs/3609`
- Angular page at `bugs/3609`

## Why there is no browser test

This originally included a browser test, skipped until the design tokens version bump.

Picking it back up, the test can't really see the fix. Our browser tests only load the V1 tokens, and this fix is a V2 token, so in that environment the modal falls back to its hardcoded values and the test ends up checking numbers that never reach anyone.

Getting it working means loading the V2 tokens in the browser test setup. That does work, but it also turns up a few other tests that look like they're checking fallback values, so it's better as its own change than folded in here.

## Steps to test

1. Run the React playground: `npm run serve:prs:react`, then open `bugs/3609`.
2. Click "Add another item" to open the modal.
3. Check that the Cancel and Save buttons have even spacing around them.
4. Same in the Angular playground: `npm run serve:prs:angular`, then open `bugs/3609`.


twjeffery:
Closing this. The actual fix shipped in GovAlta/design-tokens#168 and #3609 is closed, so the playground pages here aren't worth carrying on their own.

The test that was on this branch is dropped rather than fixed. Our browser tests load V1 tokens so it could never have seen a V2 token fix, which is tracked in #4200.

PR #4148, opened 2026-07-17
chore: Added a TESTING_STANDARDS file for AI to use when reviewing

ArakTaiRoth:
This is just a file to help AI when reviewing and testing PR's. It just documents what I generally do and how I try to test things.

### DIFF
--- a/.claude/skills/review-pr/SKILL.md
+++ b/.claude/skills/review-pr/SKILL.md
@@ -1,3 +1,8 @@
+---
+name: review-pr
+description: Use when reviewing a pull request in GovAlta/ui-components. Covers review scope, acceptance criteria, validating Goab / goab- / goa- component usage against the libraries, generated-file and documentation-migration checks, spotting unnecessary or technically wrong code, and the review output format.
+---
+
 # Testing Standards for PR Reviews
 
 Use this guide when reviewing pull requests in this repository. The goal is to verify that the PR meets its acceptance criteria, does not introduce regressions, and follows the standards already established by nearby code.
--- a/.claude/skills/review-pr/TESTING_STANDARDS.md
+++ b/.claude/skills/review-pr/TESTING_STANDARDS.md
@@ -1,0 +1,164 @@
+# Testing Standards for PR Reviews
+
+Use this guide when reviewing pull requests in this repository. The goal is to verify that the PR meets its acceptance criteria, does not introduce regressions, and follows the standards already established by nearby code.
+
+## Review Scope
+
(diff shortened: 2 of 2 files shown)


Spark450:
@ArakTaiRoth cc. @twjeffery 

Proposal: move relevant content from TESTING_STANDARDS.mdx into .claude/rules/testing.md

Flagging: the content is a general PR-review checklist (component-API drift, generated-file checks, docs regressions, output format), not "testing standards" in the sense our existing .claude/rules/testing.md uses that term (test tiers, enforced patterns). It could be worth just augmenting the current testing.md rules instead of creating a new file.

Follow-up: portability across AI tools

.claude/rules/ and .claude/skills/ are Claude Code-specific — other tools (Cursor, Copilot, Windsurf, etc.) won't read them at all. If we want this guidance to hold regardless of which AI tool someone's using, we'd need a separate tool-agnostic entry point. There's an emerging convention for this (AGENTS.md at repo root, which several tools are starting to read natively), but it wouldn't replace .claude/rules — it'd would just point back at the .claude/ files for other ai tools to follow. 

That's a bigger, separate piece of work than this PR and probably deserves its own issue rather than folding into this one.

twjeffery (commented):
@ArakTaiRoth @Spark450 Agreed it shouldn't sit at the repo root. Rather than folding it into testing.md, I think we make it its own review skill (.claude/skills/review-pr/)

testing.md is about writing tests, and this is about how to review a PR, so they're doing different jobs. A rule also loads into every session, but you only need this when you're actually reviewing, which is what skills are for. It'd sit right alongside open-pr and write-issue. 

One thing in the content itself: the Verification section says to assume tests and builds already ran. For an AI reviewer, should we have it check the result on GitHub instead of assuming it passed?

And agreed with Mark that making this work across other tools like Cursor or Copilot (something like AGENTS.md) is worth its own issue, separate from this one.

ArakTaiRoth:
@Spark450 @twjeffery The file has been moved under skills and under a new folder `review-pr`

twjeffery:
Pushed a small commit on top of yours. Renamed the file to SKILL.md and added the name and description frontmatter, which is the part that makes Claude Code actually load it as a skill when someone's reviewing a PR. Without the frontmatter it just sits in the folder as a plain doc and never gets picked up. Squash it into yours, keep it, or drop it and do your own version, whatever works.

Spark450:
Added the related issue for an AGENTS.md file, link is above.

twjeffery (approved):

PR #4150, opened 2026-07-20
feat: add component-search skill for AI-friendly markdown endpoints

bdfranck:
Add new skill that enables AI agents to search and retrieve structured documentation from the design system's AI-friendly markdown endpoints:
- /llms.txt for component and example discovery
- /components/[slug].md for full API, usage, and accessibility guidance

#### The skill supports
- Component discovery by purpose or category
- Detailed API lookups (React, Angular, Web Components)
- Cross-component searches (e.g., find all components with a prop)
- Guidance searches (usage and accessibility)
- Example searches
- Component comparisons

#### Example links
- llms.txt: https://govalta.github.io/ui-components/pr-preview/pr-4150/llms.txt
- Example component MD page: https://govalta.github.io/ui-components/pr-preview/pr-4150/components/button.md 

#### How to test locally
1. Run docs locally (docs app) on a port (ex: 4203).
2. In the [component-search skill file](https://github.com/GovAlta/ui-components/pull/4150/changes#diff-03344aaa52feb5e194ec074700628c0dbde532620be5116e02cd6601bda5c3fa), temporarily set Base URL to http://localhost:4203 (instead of prod).
3. Ask the agent to use the skill for...
    - component discovery (/llms.txt)
    - component details (/components/button.md)
    - cross-component search flow

### Demos

I asked, "What components are available for choosing between multiple options on a form?"

image

---

I asked, "What components could I use for displaying a warning on a page?"

image

### DIFF
--- a/docs/src/content/get-started/ai-tools-and-resources/skills.mdx
+++ b/docs/src/content/get-started/ai-tools-and-resources/skills.mdx
@@ -34,6 +34,13 @@ import { withBase } from "@/lib/base-url";
 <CodeSnippet language="bash" showCopy client:visible code={`npx skills add GovAlta/ui-components --skill content-design`} />
 <p><goa-link mt="s" trailingicon="open"><a href="https://github.com/GovAlta/ui-components/tree/dev/skills/content-design" target="_blank" rel="noopener noreferrer">View on GitHub</a></goa-link></p>
 
+<goa-text size="heading-s" mt="xl">component-search</goa-text>
+<goa-text size="body-m">This skill looks up component APIs, props, events, and usage or accessibility guidance from the Design System's AI-friendly Markdown endpoints (<code>llms.txt</code> and <code>/components/{"{slug}"}.md</code>).</goa-text>
+<goa-text size="body-m">It loads when you ask about a specific component's API or guidance, or want to find components for a purpose — for example "what components handle warnings?" or "show me React props for Button".</goa-text>
+<goa-text size="body-m"><strong>How to add it:</strong></goa-text>
+<CodeSnippet language="bash" showCopy client:visible code={`npx skills add GovAlta/ui-components --skill component-search`} />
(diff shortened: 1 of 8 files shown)


twjeffery (approved):
@bdfranck this is a great addition, thanks for building it. I ran your test flow against the PR preview (component discovery, a component lookup, and the cross-component prop search) and it seems to work well. I also pushed one commit on top of yours with a few small doc corrections I picked up while going through it, all in the two skill files:

1. Guidance markers, emoji to brackets. The .md pages output the guidance as [Do], [Don't], [Tip], [Warning], [Note], and there's a test that keeps the output pure ASCII. The skill described them in the emoji style (✅ Do / ❌ Don't), so I switched every place it showed up: the structure block, the "preserve the prefixes" note, the accessibility example, and the README. An agent following the emoji examples could add emoji the real pages don't have, or miss the guidance when it scans for markers that aren't there.
2. Dropped the "Tag: (none for React)" line from the structure example. The generator only prints a Tag line for Web Components, so React and Angular never show one.
3. Component counts. Changed "70+" and "60+ components" to "all components". The live index has 51 components and 69 examples right now, and a fixed number will drift, so a plain "all" holds up.

All of it sits on top of your commit, nothing rebased or squashed. Adjust or drop anything you don't agree with.

Spark450:
@bdfranck FYI @twjeffery This will be a great way to get some DS MCP like functionality 🙌 

**Follow-up commits on this branch:**

- **`docs: add component-search to the Skills page`** — the public Skills page only listed `using-goa-design-system` and `content-design`; added a `component-search` entry (description, when it loads, install command, GitHub link) so it's discoverable.
- **`fix(component-search): correct Category claim and thread Base URL consistently`** — two accuracy fixes in `SKILL.md`:
  - Component Discovery told the agent to return "Category" from the `llms.txt` index, but that index has no category field (only the full component page does) — removed the false claim.
  - Fetch instructions, the "site is down" error message, and the "always include" link rule all hardcoded `https://design.alberta.ca` instead of the documented **Base URL** variable — now consistently use `{Base URL}` so a non-default base actually gets honored.

Verified locally against a `docs` dev server: confirmed both endpoints, ran the existing unit tests (19/19 pass), and walked the skill through several real queries end-to-end.

PR #4160, opened 2026-07-22
fix(#3887): Updated Angular badge to use iconType the same as React

### LINKED ISSUE #3887: Badge `iconType` not working Angular
# Info

Using the example code for badge with custom icon from our documentation:

```html
<goab-badge type="success" content="Approved" iconType="checkmark"></goab-badge>
```

Doesn't produce an icon on a badge. It only works if you also provide `[icon]="true"`.

This is wrong according to our code and how the React component works, `icon` is supposed to be deprecated, and you're only supposed to need to use `iconType`

## Acceptance Criteria:

1. Make it so that `iconType` works the same in Angular as it does in React, that is without the use `icon`

ArakTaiRoth:
# Before (the change)

When using `iconType` in Angular, you also had to supply `[icon]="true"`. Whereas for React, you just had to supply `iconType`

# After (the change)

Now when using the Badge component in Angular, you only need to supply `iconType` in order to get a custom icon showing. Just like in React.

## Steps needed to test

Use the badge docs PR to see the change. The Angular PR page already had `iconType` being used without `icon`. It should work now.

### DIFF
--- a/libs/angular-components/src/lib/components/badge/badge.spec.ts
+++ b/libs/angular-components/src/lib/components/badge/badge.spec.ts
@@ -1,7 +1,7 @@
 import { ComponentFixture, TestBed, fakeAsync, tick } from "@angular/core/testing";
 import { GoabBadge } from "./badge";
 import { Component, CUSTOM_ELEMENTS_SCHEMA } from "@angular/core";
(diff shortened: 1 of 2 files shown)


twjeffery (approved):
@ArakTaiRoth Looks good to me

Image

PR #4166, opened 2026-07-24
fix(#3826): Modified Pagination button placement for desktop and mobile

### LINKED ISSUE #3826: Pagination: Previous and next buttons should be left and right aligned at mobile breakpoint
## Description

At the mobile breakpoint, the previous and next buttons in the simple pagination control should be left and right aligned respectively. Currently both buttons are stacked and centered vertically with no responsive alignment behaviour.

## Steps to Reproduce

1. Render the pagination component
2. Resize the viewport to mobile breakpoint
3. Observe the previous and next buttons are centered rather than left and right aligned

## Acceptance Criteria

- [ ] At the mobile breakpoint, the previous button is left-aligned
- [ ] At the mobile breakpoint, the next button is right-aligned

ArakTaiRoth:
# Before (the change)

Pagination component placement before
Desktop (Both buttons left aligned):
image

Mobile (Both buttons left aligned):
image

# After (the change)

Pagination button placement has changed.
Desktop (placed on the far right of the parent container):
image

Mobile (Previous on left side, Next on right side):
image

## Steps needed to test

Use the Pagination docs PR and the table examples to see the new layout being used.

### DIFF
--- a/apps/prs/angular/src/routes/docs/pagination/pagination.component.html
+++ b/apps/prs/angular/src/routes/docs/pagination/pagination.component.html
@@ -5,7 +5,8 @@
   [pageNumber]="basicPage"
   [itemCount]="100"
   [perPageCount]="10"
-  (onChange)="handleBasicPageChange($event)">
+  (onChange)="handleBasicPageChange($event)"
+>
 </goab-pagination>
 
 <h3>Simple</h3>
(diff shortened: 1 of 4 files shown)


twjeffery (commented):
@ArakTaiRoth Testing this out, I'm seeing the pagination controls run past the right edge whenever the component sits in something narrower than the full page. 

Image

Could we have it collapse based on the width of the space it's in, rather than the screen?

And then when it does start to stack, the page #/# stacks above the previous/next buttons, but the show #/page stays to the left. When it runs out of room to sit on one line, the whole bar should stack into a single column, instead of leaving the "Show per page" part inline while the page count and buttons drop below it. So it reads cleanly top to bottom rather than half wrapping.

Image

twjeffery (approved):
After we discussed, this looks good. The overflow I flagged is in the example code, so that's a separate issue, not this PR.

PR #4179, opened 2026-07-27
feat(#1514): Updated content property for FilterChip to support HTML …

### LINKED ISSUE #1514: [Feature Request]: Styling of Key-Value Pairs in Chip Component
### Is your feature request related to a problem? Please describe

As part of our filtering functionality on the Find Opportunities & Notices page which displays all the available procurement opportunities available to a supplier we display a list of chips representing each type of filter invoked (the Key) and its specific configuration (the Value) in a Key-Value pair. 
For clarity we’d like the ability to style Key-Value pairs in these filter chips with the “Key” as regular text and the “Value(s)” as bolded text.

### Describe the solution you'd like

Please see Figma mockup.
https://www.figma.com/file/ITHqvarhTtYal257Z4JzCB/Share-with-DS-Team?type=design&node-id=1072-116679&mode=design&t=f1GBRRudhd4O9aj8-4

### Provide evidence this is a needed component

_No response_

### Describe alternatives you've considered

_No response_

### Do you have anything already created for this that we can use?

None

### Are you currently using this proposal inside your own service

None

### Are you able to assist to bring the feature to reality?

yes

### Additional context

_No response_

ArakTaiRoth:
# Before (the change)

The `content` property in our `GoabFilterChip` component only accepted "string" content

# After (the change)

The `content` property in our `GoabFilterChip` component now accepts both "string" content and "ReactNode" for React and TemplateRef for Angular.

## Steps needed to test

Use the Feature 1514 PR Playground pages for both Angular and React

### DIFF
--- a/apps/prs/angular/src/routes/features/feat1514/feat1514.component.html
+++ b/apps/prs/angular/src/routes/features/feat1514/feat1514.component.html
@@ -1,0 +1,33 @@
+<goab-text tag="h1" mt="m">Feature #1514: FilterChip rich content</goab-text>
+
+<goab-text tag="h3">1. Text content</goab-text>
+<goab-filter-chip content="Text only" />
+
+<goab-text tag="h3">2. Bold text</goab-text>
+<goab-filter-chip [content]="boldContent" />
+
+<goab-text tag="h3">3. Strikethrough text</goab-text>
+<goab-filter-chip [content]="strikethroughContent" />
+
+<goab-text tag="h3">4. Underline text</goab-text>
+<goab-filter-chip [content]="underlineContent" />
+
+<goab-text tag="h3">5. Text and badge</goab-text>
+<goab-filter-chip [content]="badgeContent" />
+
+<ng-template #boldContent>
+  <strong>Bold text</strong>
+</ng-template>
+
+<ng-template #strikethroughContent>
+  <s>Strikethrough text</s>
+</ng-template>
+
+<ng-template #underlineContent>
+  <u>Underline text</u>
+</ng-template>
+
+<ng-template #badgeContent>
+  Application status
+  <goab-badge type="success" content="Active"></goab-badge>
+</ng-template>
--- a/apps/prs/angular/src/routes/features/feat1514/feat1514.component.ts
+++ b/apps/prs/angular/src/routes/features/feat1514/feat1514.component.ts
@@ -1,0 +1,10 @@
+import { Component } from "@angular/core";
+import { GoabBadge, GoabFilterChip, GoabText } from "@abgov/angular-components";
+
+@Component({
+  standalone: true,
+  selector: "abgov-feat1514",
+  templateUrl: "./feat1514.component.html",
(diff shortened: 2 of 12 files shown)


twjeffery (commented):
@ArakTaiRoth One thing on the screen reader side. Tabbing to the remove button on the bold example, VoiceOver says "remove filter, button, group, , removable, group". On the text only example it says "text only, removable", so you know which filter you're about to remove. All four rich content examples lose that.

The label still comes from the content string. That used to be required so it always had something in it, but with the slot it's empty. ariaLabel isn't exposed on the React or Angular wrappers the way it is on Badge. Should the label follow the slotted content instead?

ArakTaiRoth:
> @ArakTaiRoth One thing on the screen reader side. Tabbing to the remove button on the bold example, VoiceOver says "remove filter, button, group, , removable, group". On the text only example it says "text only, removable", so you know which filter you're about to remove. All four rich content examples lose that.
> 
> The label still comes from the content string. That used to be required so it always had something in it, but with the slot it's empty. ariaLabel isn't exposed on the React or Angular wrappers the way it is on Badge. Should the label follow the slotted content instead?

@twjeffery I modified a lot. I first off added `ariaLabel` as a property for both Angular and React wrappers (not sure why it wasn't, it was a property on the web component already). I then removed the `aria-label` from the div that had `role="presentation"` (this literally says this div is presentation only, it doesn't matter for screen readers). I'm again not sure why the aria label was placed here, because it serves no purpose here. I then modified the aria label on the icon button (what is actually tabbed to), to read either the supplied aria label, the text content, or the slotted content (whichever is given) and add "Remove filter" the beginning of it.

I hope this makes it better for you.

twjeffery:
> > @ArakTaiRoth One thing on the screen reader side. Tabbing to the remove button on the bold example, VoiceOver says "remove filter, button, group, , removable, group". On the text only example it says "text only, removable", so you know which filter you're about to remove. All four rich content examples lose that.
> > The label still comes from the content string. That used to be required so it always had something in it, but with the slot it's empty. ariaLabel isn't exposed on the React or Angular wrappers the way it is on Badge. Should the label follow the slotted content instead?
> 
> @twjeffery I modified a lot. I first off added `ariaLabel` as a property for both Angular and React wrappers (not sure why it wasn't, it was a property on the web component already). I then removed the `aria-label` from the div that had `role="presentation"` (this literally says this div is presentation only, it doesn't matter for screen readers). I'm again not sure why the aria label was placed here, because it serves no purpose here. I then modified the aria label on the icon button (what is actually tabbed to), to read either the supplied aria label, the text content, or the slotted content (whichever is given) and add "Remove filter" the beginning of it.
> 
> I hope this makes it better for you.

That reads much better, I'm getting "remove filter: text only" and the right text on the others.

One thing I noticed that's outside the scope of this one. The badge example reads "remove filter: application status" and doesn't pick up the Active badge, since the badge renders its text inside itself. Nothing to change here, just noting it.

twjeffery (approved):

PR #4180, opened 2026-07-27
feat(#2374): Updated content property for Badge to support HTML content

### LINKED ISSUE #2374: Badge: Slotted content
# Info

Badge `content` property should be changed to use a default slot for composition, to make it similar to other components and to allow html content. E.g. `<goa-badge type="information" content="Info"></goa-badge>` would become: `<goa-badge type="information">Info</goa-badge>`.

## Acceptance Criteria

1. Allow for slotted content inside Badge
2. Update the documentation to show slot for content property

ArakTaiRoth:
# Before (the change)

`content` property for `GoabBadge` only supported "string" values

# After (the change)

`content` property for `GoabBadge` now supports both "string" values and HTML content

## Steps needed to test

Use the PR pages for feature 2374

### DIFF
--- a/apps/prs/angular/src/routes/features/feat2374/feat2374.component.html
+++ b/apps/prs/angular/src/routes/features/feat2374/feat2374.component.html
@@ -1,0 +1,33 @@
+<goab-text tag="h1" mt="m">Feature #2374: Badge rich content</goab-text>
+
+<goab-text tag="h3">1. Text content</goab-text>
+<goab-badge type="information" content="Text only"></goab-badge>
+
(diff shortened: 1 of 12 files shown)


twjeffery (approved):
@ArakTaiRoth Looks good overall. The rich content cases all show correctly in the playground and plain string content is unchanged.

The issue asks for a default slot, so <goa-badge type="information">Info</goa-badge> would work. This uses a named content slot instead, so that form shows nothing. The named slot looks right to me, since content still falls back to the string and a default slot would treat stray whitespace as content. Should we put that on #2374 so it's clear why it closed this way?

PR #4181, opened 2026-07-27
fix(#2378): Accordion and Details content read aloud

### LINKED ISSUE #2378: When expanded content is not read
# Info

Found via routine accessibility testing

When you expand either Accordion or the Details component, the component is correctly read out as expanded, but the content inside the component is not read out. It would only be read out if there was an interactive element inside and it was then tabbed to, but if no interactive element exists, nothing will ever be read out.

## Acceptance Criteria:

1. Content inside a Details and/or Accordion component should be read automatically when expanded

# More Info

This might be related to this story, #2365 , in that it is slotted content not being read by a screen reader

ArakTaiRoth:
# Before (the change)

Content inside `GoabDetails` and `GoabAccordion` wasn't read out by a screen reader unless there was interactive content inside and the user tabbed to it.

# After (the change)

Content inside `GoabDetails` and `GoabAccordion` should now be read out as soon as the component is expanded.

## Steps needed to test

Use the PR bug 2378

### DIFF
--- a/apps/prs/angular/src/routes/bugs/bug2378/bug2378.component.html
+++ b/apps/prs/angular/src/routes/bugs/bug2378/bug2378.component.html
@@ -1,0 +1,17 @@
+<main>
+  <goab-text tag="h1">Bug #2378: Expanded content is not read</goab-text>
+
+  <goab-text tag="h3">Details</goab-text>
+  <goab-details heading="Supporting details">
+    <goab-text tag="p">
+      This Details content should be announced when the component expands.
+    </goab-text>
+  </goab-details>
+
+  <goab-text tag="h3">Accordion</goab-text>
+  <goab-accordion heading="Supporting information">
+    <goab-text tag="p">
+      This Accordion content should be announced when the component expands.
+    </goab-text>
+  </goab-accordion>
+</main>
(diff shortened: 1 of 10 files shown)


twjeffery (commented):
@ArakTaiRoth I tried both with VoiceOver and I'm not getting the content read out when they expand. I checked while they were open and role="alert" is on both, so it's being applied, it just doesn't announce anything for me.

Are you seeing something different?

ArakTaiRoth:
> @ArakTaiRoth I tried both with VoiceOver and I'm not getting the content read out when they expand. I checked while they were open and role="alert" is on both, so it's being applied, it just doesn't announce anything for me.
> 
> Are you seeing something different?

@twjeffery Apparently VoiceOver handles this differently than JAWS or NVDA. I've made a small change that forces the Svelte component to wait a second before putting content inside the component. This "should" fix the issue, but as I can't test it, I'm unsure, let me know.

twjeffery (commented):
@ArakTaiRoth I'm still not getting anything read out on either one.

But when I was testing, if the role gets set a bit later after the expanding has finished, it does announce. So maybe it just needs a longer delay?

ArakTaiRoth:
> @ArakTaiRoth I'm still not getting anything read out on either one.
> 
> But when I was testing, if the role gets set a bit later after the expanding has finished, it does announce. So maybe it just needs a longer delay?

@twjeffery I've changed how it works. I was using `await tick()`. I've removed that and added an actual defined timeout of 100ms, hopefully this gives enough time and works. Let me know.

twjeffery (approved):
@ArakTaiRoth both reading out now

PR #4183, opened 2026-07-29
fix(#4140): vertically align radio item label in V2

### LINKED ISSUE #4140: [Bug]: Vertical alignment issue on radio buttons
### Your email

bruno.alves@gov.ab.ca

### Web components version

Unknown

### React or Angular components version

Unknown

### Expected Behaviour

Radio buttons with correct vertical alignment. See DS Component Library as a reference.
Image

### Actual Behaviour

When radio buttons are implemented in our questionnaire, the text is misaligned with the radio button.

Image

The same issue is happening on the DS website:
https://design.alberta.ca/components/radio-group/#react#properties

Image

### Replication Steps

Create a question with a radio button group.

### Stackblitz URL

_No response_

### Jam.dev URL

_No response_

### Additional context

_No response_

twjeffery:
The radio label sits too high, so it does not line up with the circle beside it. It is most noticeable at the compact size.

In V2 the label was still getting an upward nudge that only V1 needs. V1's label line height is taller than the circle, so it needs the nudge to look centred. V2's line height matches the circle height, so it does not. The V2 checkbox already works this way, and now the V2 radio matches it.

V1 renders exactly as it did before.

Closes #4140

image

## Labels that wrap

Items whose label wraps onto more than one line are 3px taller than they were. The label was being pulled up out of its own box, so those items were rendering 3px shorter than their content. A wrapping item now measures the same as a wrapping item in a checkbox list.

Single line items are unchanged, with or without descriptions, including the spacing between items in a radio group.

## Steps needed to test

Open bugs/4140 in the React and Angular test apps. Check that the label text looks vertically centred against the circle at both sizes, and that it sits the same way as the labels in the checkbox list at the bottom of the page.

The page also covers descriptions, a mix of items with and without descriptions, labels that wrap, and the disabled and error states.

### DIFF
--- a/apps/prs/react/src/routes/bugs/bug4140.tsx
+++ b/apps/prs/react/src/routes/bugs/bug4140.tsx
@@ -1,0 +1,282 @@
+import React, { useState } from "react";
+import {
+  GoabBlock,
+  GoabCheckbox,
+  GoabCheckboxList,
+  GoabDivider,
+  GoabOneColumnLayout,
+  GoabPageBlock,
+  GoabRadioGroup,
+  GoabRadioItem,
+  GoabText,
+} from "@abgov/react-components";
+import { Link } from "react-router-dom";
+
+const narrow: React.CSSProperties = { maxWidth: "20rem" };
+
+export function Bug4140Route() {
+  const [payment, setPayment] = useState("cheque");
+  const [compactPayment, setCompactPayment] = useState("cheque");
+  const [described, setDescribed] = useState("deposit");
+  const [compactDescribed, setCompactDescribed] = useState("deposit");
+  const [mixed, setMixed] = useState("cheque");
+  const [wrapping, setWrapping] = useState("deposit");
+  const [wrappingDescribed, setWrappingDescribed] = useState("deposit");
+  const [compactWrapping, setCompactWrapping] = useState("deposit");
+
+  return (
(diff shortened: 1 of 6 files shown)


apps/prs/react/src/routes/bugs/bug4140.tsx
```diff
@@ -0,0 +1,282 @@
...
+  const [mixed, setMixed] = useState("cheque");
+  const [wrapping, setWrapping] = useState("deposit");
+  const [wrappingDescribed, setWrappingDescribed] = useState("deposit");
+  const [compactWrapping, setCompactWrapping] = useState("deposit");
+
+  return (
+    <GoabPageBlock width="full">
+      <GoabOneColumnLayout>
+        <GoabBlock direction="column" gap="l">
+          <goa-link mt={"xl"} leadingicon={"arrow-back"}>
```
ArakTaiRoth: You're using `goa-link` here which I don't think even works? And notable I don't think we need a back link here anyways.

twjeffery: > You're using `goa-link` here which I don't think even works? And notable I don't think we need a back link here anyways.

@ArakTaiRoth Removed the back link. Should be good now.

ArakTaiRoth (approved):

PR #4184, opened 2026-07-29
chore: Adding a new example for GoabTabs

ArakTaiRoth:
This just adds a new example for the GoabTabs component. The example is making a specific tab active via a button click (or really any action on the page). This was a requested example on the Support page.

### DIFF
--- a/apps/prs/angular/src/routes/docs/tabs/tabs.component.html
+++ b/apps/prs/angular/src/routes/docs/tabs/tabs.component.html
@@ -27,6 +27,17 @@
   <goab-tab heading="Third">Third tab content.</goab-tab>
 </goab-tabs>
 
+<h3>Activate a tab with a button</h3>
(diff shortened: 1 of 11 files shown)


twjeffery (commented):
@ArakTaiRoth The example works great in the playground. 

https://github.com/user-attachments/assets/d60bc87f-91d1-4021-9f86-ea968fe43e6f

One thing to update: the example doesn't have an index.mdx so it won't show up on the docs site. The code files are there but nothing picks them up without it.

ArakTaiRoth:
> @ArakTaiRoth The example works great in the playground.
> 
>  Screen.Recording.2026-07-29.at.4.42.54.PM.mov 
> One thing to update: the example doesn't have an index.mdx so it won't show up on the docs site. The code files are there but nothing picks them up without it.

@twjeffery Added an `index.mdx` file, sorry about that, didn't realize it was needed

twjeffery (approved):

PR #4185, opened 2026-07-29
chore: Fix the multiple flaky tests we have

ArakTaiRoth:
This makes massive sweeping changes to our tests written for `datepicker.browser.spec.tsx`, `dropdown-multiselect.browser.spec.tsx`, and `tabs.browser.spec.tsx`. This is to hopefully alleviate all the flaky tests that have been failing constantly for the last while.


twjeffery (approved):
@ArakTaiRoth Looks good

PR #4197, opened 2026-08-05
feat(#1741): Added ariaHidden property to Icon

### LINKED ISSUE #1741: Feature Request: aria-hidden label on icon
# Info

Raised by Rob Schrage

There are instances where our icons are used, like buttons and inputs, where having an aria-label doesn't make sense as screen readers will read the button or the form label. In these instances, it could be nice to have an aria-hidden attribute on our icons to hide those icons from screen readers, and from automated accessibility scans, so they aren't highlighted as issues that need to be fixed.

## Acceptance Criteria:

1. goa-icon has an `ariahidden` property that can be used in specific situations like buttons and inputs
2. React also features an `ariaHidden` property for `GoAIcon`
3. In cases of buttons and inputs using our icons, if the property can automatically be used that would be great.

ArakTaiRoth:
# Before (the change)

All icons were found by automated accessibility tools and screen readers, even though in a lot of cases they didn't need to be, as they were just adding visual flare to something that was already read.

# After (the change)

This commit adds the `ariaHidden` property to the Icon component. By using this, you hide the icon from screen readers and automated accessibility tools. This way you can hide icons that don't need to be seen by screen readers.

## Steps needed to test

Use the PRs feat1741 for both Angular and React

### DIFF
--- a/apps/prs/angular/src/routes/features/feat1741/feat1741.component.html
+++ b/apps/prs/angular/src/routes/features/feat1741/feat1741.component.html
@@ -1,0 +1,23 @@
+<goab-text tag="h1" mt="m">Feature #1741: Icon aria-hidden label</goab-text>
+
+<goab-text tag="h2">Test Cases</goab-text>
+
+<goab-text tag="p">
+  Test both icons with a screen reader. The 1st one has ariaLabel set to "Applicaton saved
+  icon" and it shouldn't be read. The second one has ariaLabel set to "Important
+  information" and it should be read.
+</goab-text>
+
+<goab-text tag="h3">Test 1: Decorative icon is hidden</goab-text>
+
+<goab-block direction="row" alignment="center">
+  <goab-icon
+    type="checkmark-circle"
+    [ariaHidden]="true"
+    ariaLabel="Application saved icon"
+  />
+  <span>Application saved</span>
+</goab-block>
+
+<goab-text tag="h3">Test 2: Default icon remains accessible</goab-text>
+<goab-icon type="information-circle" ariaLabel="Important information" />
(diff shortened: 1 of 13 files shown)


twjeffery (commented):
@ArakTaiRoth Looks good, I tested both cases with a screen reader and the decorative one stays quiet.

**Three small things:**
1. An icon with nothing set on it still shows aria-hidden="false". React and Angular leave the attribute off when it's false, so should we do the same on the web component?
2. On the docs site the ariahidden description is showing the @ default "false" text, just above the "Defaults to false" line that's already there. Looks like only the Svelte one needs it taken out, React and Angular need theirs to keep showing a default.
3. The accessible-icons-need-label guidance already covers Icon but only says when to use ariaLabel. Should we add ariaHidden there so teams know what to reach for on decorative icons? There's role="presentation" doing something similar too, which might be worth separating out.

ArakTaiRoth:
> @ArakTaiRoth Looks good, I tested both cases with a screen reader and the decorative one stays quiet.
> 
> **Three small things:**
> 
> 1. An icon with nothing set on it still shows aria-hidden="false". React and Angular leave the attribute off when it's false, so should we do the same on the web component?
> 2. On the docs site the ariahidden description is showing the @ default "false" text, just above the "Defaults to false" line that's already there. Looks like only the Svelte one needs it taken out, React and Angular need theirs to keep showing a default.
> 3. The accessible-icons-need-label guidance already covers Icon but only says when to use ariaLabel. Should we add ariaHidden there so teams know what to reach for on decorative icons? There's role="presentation" doing something similar too, which might be worth separating out.

@twjeffery I've made all the changes you've requested

twjeffery (approved):
@ArakTaiRoth Looks good to me now

https://github.com/user-attachments/assets/e9d42e35-ed90-4648-9c24-ea63328e4d84

Issue #171, opened 2021-10-25
Container


Spark450:
# What is being submitted?
- Container component and pattern

[Link to design](https://www.figma.com/file/3pb2IK8s2QUqWieH79KdN7/DIO-DS---Components?node-id=1789%3A12623)

**Properties**

- Tombstone information variant
- “edit” variant
- read-only data view
- Actions (toolbar?)
- Buttons
- Sections and sub-sections
  - Divider and Child card examples
- Others? (collect use cases)
- Badges
- Description
- expand collapse?
- Nesting containers?
  - Only 1 level of nesting
- Drop shadow?
- Responsive

# Examples, links, and research
- Is this a pattern or a component?
- Need to account for containers that need hierarchies and no hierarchies
- grouping data 
- Can we expand the card component or a whole new component

[Mindsphere example](https://design.mindsphere.io/patterns/container.html)
[Shopify example](https://polaris.shopify.com/components/structure/card#navigation)
[Salesforce example](https://www.lightningdesignsystem.com/components/cards/)

**Container types**

- General Information
  - Tombstone (read only)
  - Information with actions (editable)
  - Generic container (included for design flexibility)
- Child information
  - Tombstone (read only)
  - Information with actions (editable)
  - Generic container (included for design flexibility)

# Anything else
- Add any other context or screenshots about the feature request here.
<img width="1440" alt="Application page layout - Assessor - Financial award" src="https://user-images.githubusercontent.com/51723535/138740412-f0718f64-f8d8-4733-9985-f6ff5dc624ea.png">
<img width="1440" alt="Application page layout - Assessor - Overview v2" src="https://user-images.githubusercontent.com/51723535/138740423-f10bddc4-5493-4df8-bdf6-e02ea1527917.png">

![image](https://user-images.githubusercontent.com/41582/144634338-90c81895-3ec5-4286-a960-0739192d31e2.png)

![JP Detail View – Multi Docket Numbers D#1 - REVISED](https://user-images.githubusercontent.com/51723535/138740431-0bb5c5ee-6fa4-473a-981a-2ebda8ebae4c.png)

![JP Detail View – Multi Docket Numbers D#3 – 1](https://user-images.githubusercontent.com/51723535/138740433-a9c010b6-fae9-4c11-b4f9-c7c753a58c1c.png)

![Provider platform - Program + Session date (added)](https://user-images.githubusercontent.com/51723535/138740436-e86d4d7a-5273-432b-8d1d-266b066a1c7a.png)

[...]

PR #4199, opened 2026-08-05
fix(#4198): use the placeholder colour for multiselect placeholder fallbacks

### LINKED ISSUE #4198: Dropdown: placeholder and option text colours are inverted
## Description

In the dropdown, the `—Select—` placeholder renders in the default text colour (#000000) while the options in the open menu render in the secondary colour (#6f6f6f). The weighting is backwards. The placeholder, which should be de-emphasized, is the darkest text in the component, and the options, which are the content, are the lightest.

Two separate corrections:

**Placeholder too dark.** Text Input and Text Area both use the lighter placeholder colour. Dropdown and Dropdown Multiselect have the same token, but it points at the default text colour, so an empty field reads as filled. This makes filled and empty fields hard to distinguish when scanning a long form.

**Options too light.** Dropdown options should use the default text colour. Raised by Bruno Alves and Lori in the support channel.

## Location

`data/component-design-tokens/` in GovAlta/design-tokens:

```
text-input-color-text-placeholder           = {input.color.text.placeholder}   ->  #9f9f9f  correct
text-area-color-text-placeholder            = {input.color.text.placeholder}   ->  #9f9f9f  correct
dropdown-color-text-placeholder             = {input.color.text.default}       ->  #000000
dropdown-multiselect-color-text-placeholder = {input.color.text.default}       ->  #000000
dropdown-item-color-text                    = {color.text.secondary}           ->  #6f6f6f
```

Consumed at `Dropdown.svelte:1187` (placeholder) and `Dropdown.svelte:1107` (options), and at `DropdownMultiselect.svelte:784` (closed state span) and `DropdownMultiselect.svelte:808` (filter input).

## Expected

```
dropdown-color-text-placeholder             = {input.color.text.placeholder}
dropdown-multiselect-color-text-placeholder = {input.color.text.placeholder}
dropdown-item-color-text                    = {color.text.default}
```

Dropdown Multiselect's two placeholder fallbacks should also move from `var(--goa-input-color-text-default)` to `var(--goa-input-color-text-placeholder)`.

## Notes

Hover and selected states need no change. `dropdown-item-color-text-hover` is already `{color.text.default}`, and both states carry their own background change (`greyscale.100` and `interactive.default`), so the affordance survives the resting colour matching the hover colour.

**The options change applies to Dropdown only.** Dropdown Multiselect's options are `goa-checkbox` elements, so their text is governed by `checkbox-color-label` (`{input.color.text.secondary}`, #353535), which is already near black. Matching it exactly would mean repainting every checkbox in the system for a #353535 to #000000 difference, which is not worth it.

Related cleanup, not part of this fix: `dropdown-multiselect-item-color-text` and `dropdown-multiselect-item-color-bg-hover` have zero references in the codebase. They look like the tokens that control the multiselect options, and they do not.

twjeffery:
Companion to GovAlta/design-tokens#171, which carries the actual colour change for issue #4198. Best reviewed together with that one.

## What is here

**The placeholder fallbacks.** Two CSS fallback values in `DropdownMultiselect.svelte` move from `var(--goa-input-color-text-default)` to `var(--goa-input-color-text-placeholder)`, on `.value-display.placeholder` and `.filter-input::placeholder`. These are the fallbacks for `--goa-dropdown-multiselect-color-text-placeholder`. That token is defined, so the fallback never fires and this changes no rendering on its own. They are wrong on their own terms and this makes them honest.

**Playground pages** at `bugs/4198` in both the React and Angular playgrounds. Empty fields side by side across dropdown, multiselect, text input and text area, an empty field next to a filled one, an open menu, the multiselect whose options are deliberately unchanged, and the disabled and error states as a no drift check. Use the theme toggle in the nav for dark mode.

## One commit still to come

`@abgov/design-tokens-v2` is pinned at 2.12.0. Once #171 merges and publishes, this PR takes a one line bump to the published version, which is what actually makes the change visible to consuming teams.

Flagging it up front so it is not a surprise after review. **Until that commit lands the playground pages render the unfixed state, which is expected.** Nothing else about this PR changes.

## Notes

For anyone applying the Svelte change by hand, `var(--goa-input-color-text-default)` appears five times in that file. The other three belong to `--goa-dropdown-multiselect-color-text`, the regular text colour, where falling back to the default is correct.

Dark mode needs no work. The tokens resolve through greyscale values that the dark theme already overrides, so `dark-theme.css` came back byte identical from the tokens rebuild and is not in #171.

### DIFF
--- a/apps/prs/angular/src/routes/bugs/4198/bug4198.component.html
+++ b/apps/prs/angular/src/routes/bugs/4198/bug4198.component.html
@@ -1,0 +1,184 @@
+<goab-block direction="column" gap="l">
+  <goab-text tag="h1" size="heading-l" mb="m">Dropdown text colour hierarchy</goab-text>
+
+  <goab-text size="body-m">
+    The dropdown used to paint its placeholder in the darkest text colour and its menu options
+    in a lighter grey, so an empty field looked filled and the options looked like the least
+    important thing on screen. That weighting is now the other way round.
+  </goab-text>
+
+  <goab-text size="body-m">
+    Two things to look for. The placeholder should look lighter than the text around it, and it
+    should match the placeholder in the text input and text area. Open a menu and the options
+    should be the darkest text in the component, darker than the placeholder sitting above them.
+  </goab-text>
+
+  <goab-text size="body-s">
+    Use the theme toggle in the nav to check both light and dark mode. The fix carries into dark
+    mode on its own, because these tokens resolve through greyscale values that the dark theme
+    already overrides.
+  </goab-text>
+
+  <goab-divider></goab-divider>
+
+  <goab-text tag="h2" size="heading-m" mt="xl" mb="m">Empty fields, all four together</goab-text>
+  <goab-text size="body-m" mb="m">
+    Every placeholder here should look the same weight. Before the fix the two dropdowns were
+    noticeably darker than the text input and text area.
+  </goab-text>
+
+  <div style="max-width: 20rem">
+    <goab-block direction="column" gap="m">
+      <goab-form-item label="Dropdown">
+        <goab-dropdown
+          name="emptyCity"
+          [value]="emptyCity"
+          (onChange)="onEmptyCityChange($event)"
+        >
+          <goab-dropdown-item value="edmonton" label="Edmonton"></goab-dropdown-item>
+          <goab-dropdown-item value="calgary" label="Calgary"></goab-dropdown-item>
+          <goab-dropdown-item value="red-deer" label="Red Deer"></goab-dropdown-item>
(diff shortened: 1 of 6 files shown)


ArakTaiRoth (commented):
This PR essentially isn't doing anything. I'm just looking at your created PR files in Angular.
1. Dropdown and Dropdown Multiselect still have dark text for placeholder text
2. Dropdown and Dropdown Multiselect still have greyed out text for selectable options
3. Dropdown Multiselect also doesn't look as dark as it should?
4. Disabled Dropdown is using the correct placeholder grey value, Dropdown error and Multiselect disabled aren't

I've attached pictures of the issues I raised above:

Image

Image

Image

Image

twjeffery:
@ArakTaiRoth Closing this one. This is now already fixed in GovAlta/design-tokens#171

PR #4202, opened 2026-08-06
fix(#4201): Updated base angular component to allow for setting disabled

### LINKED ISSUE #4201: Radio Group `[disabled]` doesn't work with `formControlName`
# Info

If you use `formControlName` on a `GoabRadioGroup`, then using the property `[disabled]` has no effect. Only in Angular.

Code for replication:
```html
<goab-form-item [formGroup]="interpreterGuardianForm">
  <goab-radio-group
    [disabled]="true"
    formControlName="isGuardianAppearingRadioGroupControl"
    orientation="vertical"
    (onChange)="isGuardianAppearingRadioOnChange($event)"
  >
    <goab-radio-item value="1" label="No guardian" />
    <goab-radio-item
      value="2"
      label="Guardian to appear in person at Law Enforcement Agency"
    />
    <goab-radio-item value="3" label="Guardian to appear by phone" />
    <goab-radio-item value="4" label="Not appearing" />
  </goab-radio-group>
</goab-form-item>
```

## Acceptance Criteria:

1. Should be able to use `formControlName` alongside `[disabled]` and expect both to continue working

ArakTaiRoth:
# Before (the change)

Using `[disabled]` on an input component that was also using FormControl, essentially did nothing. It only worked for Template driven component usage. This is mostly fine because if you're using FormControl, you should be setting the `disabled` property via the Form Control declaration, not via the `[disabled]` property. But this doesn't remove the fact that some people do things this way anyways.

# After (the change)

Using `[disabled]` should now work for both Template drive and Form Control based inputs. This wasn't working initially because both options were working with the same `isDisabled` value, so if you set it via the component, the Form Control would overwrite that. Now, it's being set independently and if either option is true, the component is disabled.

This also involved modifying the `extract-api.ts` script, because it only got properties based off of base variables, not getters and setters.

## Steps needed to test

You can test in the PR bug 4201 for Angular. This is not represented in React, as the problem was Angular specific.

### DIFF
--- a/apps/prs/angular/src/routes/bugs/fix4201/fix4201.component.html
+++ b/apps/prs/angular/src/routes/bugs/fix4201/fix4201.component.html
@@ -1,0 +1,65 @@
+<goab-block direction="column" gap="l">
+  <goab-text tag="h1" size="heading-l" mb="m">
+    Bug #4201: Disabled radio groups with Angular forms
+  </goab-text>
+
+  <goab-text tag="h2" size="heading-m" mb="m">
+    1. Template-driven form using [disabled] property
+  </goab-text>
+  <goab-text size="body-m">
+    The component input supplies the disabled state while ngModel supplies the value.
+  </goab-text>
+  <goab-radio-group
+    name="templateDrivenDisabled"
+    [(ngModel)]="templateDrivenValue"
+    [disabled]="true"
+  >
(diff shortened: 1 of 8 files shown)


twjeffery (approved):
@ArakTaiRoth Looks good to me

PR #4213, opened 2026-08-10
fix(#2456): Updated Details margin-top to use proper token

### LINKED ISSUE #2456: Details: Need a new token for `--goa-details-margin-top`
# Info

In this PR, https://github.com/GovAlta/ui-components/pull/2446, a change was made to the CSS of the Details component. The token `--goa-details-margin-bottom` was assigned to the `margin-top` property. This token is also assigned to the `margin-bottom` property. This could lead to unexpected consequences if people try to change this token rightfully thinking that they're just changing the margin-bottom.

## Acceptance Criteria:

1. A new token needs to be created, `--goa-details-margin-top` and it needs to be associated with the margin-top of the Details component, where the above PR changed it.

ArakTaiRoth:
# Before (the change)

The Details component was using `--goa-details-margin-bottom` for it's `margin-top` property. This wasn't great, as people might change the margin-bottom and not expect the margin-top to also change.

# After (the change)

The details component now properly uses a new token `--goa-details-margin-top` for it's `margin-top` property.

## Steps needed to test

Test using the Details docs PR for both Angular and React. There should be no difference for the actually calculated margin-top, but if you inspect it, it should be using `--goa-details-margin-top` instead.

### DIFF
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -489,23 +489,6 @@
         "node": ">=18"
       }
     },
-    "node_modules/@types/react": {
-      "version": "19.2.17",
-      "resolved": "https://registry.npmjs.org/@types/react/-/react-19.2.17.tgz",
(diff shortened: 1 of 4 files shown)


twjeffery (approved):
Looks good to me. I noticed that the spacing value needed to be adjusted to match design, updated in new tiny PR https://github.com/GovAlta/design-tokens/pull/173

PR #4215, opened 2026-08-10
fix(#2773): Removed the double tab from Tooltip

### LINKED ISSUE #2773: Tooltip always has double focus (should not)
Here is an example of tooltips on icon buttons. The user has to tab twice through each icon button to get through the tooltip and then the button itself. 
https://design.alberta.ca/components/tooltip#tab-1#show-a-label-on-an-icon-only-button

## Acceptance Criteria:
- When a tooltip is on a button or other interactive object, it should not have two focuses. The tooltip should show only once in the tab order, at the same time as the interactive ui (e.g. button) is focuses.

ArakTaiRoth:
# Before (the change)

When using the Tooltip component with an interactive component, you had to tab once to the component to activate the tooltip and then tab a second time to actually activate the interactive component.

# After (the change)

Now with one tab it should active the Tooltip (show it) and allow you to interact with the interactive component.

## Steps needed to test

Use the Tooltip docs PR for both Angular and React

### DIFF
--- a/libs/react-components/specs/tooltip.browser.spec.tsx
+++ b/libs/react-components/specs/tooltip.browser.spec.tsx
@@ -1,11 +1,77 @@
 import { render } from "vitest-browser-react";
-import { page } from "@vitest/browser/context";
+import { page, userEvent } from "@vitest/browser/context";
 import { useState } from "react";
 
 import { GoabTooltip, GoabButton } from "../src";
 import { expect, describe, it, vi } from "vitest";
 
 describe("Tooltip Browser Tests", () => {
+  it("focuses an interactive trigger and shows its tooltip with one tab", async () => {
+    const onClick = vi.fn();
+    const Component = () => (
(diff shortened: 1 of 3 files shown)


twjeffery (commented):
@ArakTaiRoth The double tab fix works, I've wanted that for a while.

One thing I noticed, previously with VoiceOver there were two stops and the first one reads out the tooltip text. On this branch there's now one stop (good) but the tooltip text doesn't get read at all by the screen reader

ArakTaiRoth:
> @ArakTaiRoth The double tab fix works, I've wanted that for a while.
> 
> One thing I noticed, previously with VoiceOver there were two stops and the first one reads out the tooltip text. On this branch there's now one stop (good) but the tooltip text doesn't get read at all by the screen reader

@twjeffery This issue should be resolved

twjeffery (approved):

PR #4216, opened 2026-08-10
fix(#2314): Button width property respected on mobile

### LINKED ISSUE #2314: Button: Mobile Width
# Info

Currently with this story, #1412, the width of the button can be controlled on Desktop experiences. However on mobile experiences, the design system's 100% width on buttons overrides any width currently set to the Button.

## Acceptance Criteria:

1. The Button's `width` property should also be respected on mobile, default to 100% as per usual, but use the `width` property if given.
2. `auto` for the `width` property should work on both Desktop and Mobile, setting the width of the button to the text inside

ArakTaiRoth:
# Before (the change)

The `width` property for `GoabButton` wasn't respected on mobile, instead just defaulting to 100% no matter what was supplied.

# After (the change)

The `width` property on mobile is now respected and used, only defaulting to 100% width when no `width` property is given.

## Steps needed to test

Use the PR bug2314 for both Angular and React

@twjeffery I'm not sure if this might be considered a breaking change. If people have been supplying width to their buttons, but still wanting and expecting them to be 100% on mobile, this would affect that.

### DIFF
--- a/apps/prs/angular/src/routes/bugs/2314/bug2314.component.html
+++ b/apps/prs/angular/src/routes/bugs/2314/bug2314.component.html
@@ -1,0 +1,23 @@
+<goab-text tag="h1" size="heading-l" mb="m">Button: Mobile width</goab-text>
+
+<goab-text tag="p">
+  The width of the Button component was controllable on desktop (ie. the width property
+  was respected). But on mobile, it defaulted to and only used width="100%", despite any
+  width being set.
+</goab-text>
+
+<goab-text tag="p">
+  This fix should see the width property respected on mobile as well as desktop. Default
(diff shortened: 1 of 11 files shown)


twjeffery (commented):
@ArakTaiRoth One thing we should add: width="fit-content" works now too, which is a nice way to stop a button going full width when on mobile. Can we add that to the width property description to show on the props table?

Image

ArakTaiRoth:
> @ArakTaiRoth One thing we should add: width="fit-content" works now too, which is a nice way to stop a button going full width when on mobile. Can we add that to the width property description to show on the props table?
> 
> Image

@twjeffery Documentation should be updated now

twjeffery (approved):

PR #4217, opened 2026-08-10
fix(#2819): headingContent can be null

### LINKED ISSUE #2819: Accordion heading content is output even when nothing exists
# Info

Reported by Daryl C

In our Accordion heading we are only checking if `secondarytext` exists and then outputting it. We always output a `<span class="heading heading-{headingsize}>` even if a slot is provided and no text is provided for `heading`. And we always output `<div class="heading-content">` even if no slot is provided and instead `heading` text is given.

## Acceptance Criteria:

1. Check if elements exist before outputting them for Accordion (`heading`, `heading slot`)

ArakTaiRoth:
# Before (the change)

The div for heading content always existed and took up space regardless if `headingContent` was ever supplied or not

# After (the change)

Now the div for heading content only shows up and takes up space if `headingContent` is supplied.

## Steps needed to test

Use the PR pages for bug2819 to review. You'll need to inspect the each container to very that the associated div for headingContent is shown when it's supplied and not shown when it's not supplied.

### DIFF
--- a/apps/prs/angular/src/routes/bugs/2819/bug2819.component.html
+++ b/apps/prs/angular/src/routes/bugs/2819/bug2819.component.html
@@ -1,0 +1,24 @@
+<goab-block direction="column" gap="l">
+  <goab-text tag="h1">Bug #2819: Accordion heading content container</goab-text>
+  <goab-text tag="p">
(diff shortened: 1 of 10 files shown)


twjeffery (approved):
looks good to me

Issue #2697, opened 2025-05-06
Footer navigation links incorrect spacing when more than one column


twjeffery:
Footer navigation links have incorrect spacing when there is more than one column set (see image)

<img width="1464" alt="Image" src="https://github.com/user-attachments/assets/8e58c3c9-aa81-4bf3-b43e-7cb5dcc1da6c" />

## Acceptance Criteria:
- footer navigation links should maintain goa-space-m (16px) spacing vertically between navigation links

<img width="1309" alt="Image" src="https://github.com/user-attachments/assets/22828cdf-056b-4985-af76-9d5a04fcae1b" />

PR #4219, opened 2026-08-12
feat(#2812): Updated footer to accept slotted content

### LINKED ISSUE #2812: Allow app-footer meta/nav sections to accept goa-link components
Issue reported by: brendan.barr@gov.ab.ca@gov.ab.ca
Web component version: 1.34.1-alpha.5
Angular/React version: 4.3.0-alpha.4

## Expected Behaviour

I can add a trailing icon to a hyperlink that has `target` and `rel` properties. The trailing icon hints that it will open in a new browsing tab.

## Actual Behaviour

The entire footer "meta" section renders as empty if anything elements other than `a` tags are used

## Replication Steps

```html
<goa-app-footer maxContentWidth="100%">
  <goa-app-footer-meta-section slot="meta">
    <goab-link trailingicon="open">
      <a href="https://www.alberta.ca/system/files/custom_downloaded_images/mep-info-referrals-resources.pdf" target="_blank" rel="noopener noreferrer">
        Assistance Resources</a>
    </goab-link>
  </goa-app-footer-meta-section>
</goa-app-footer>
```

## Additional Info

- getting same result in footer slots, "meta" or "nav"

ArakTaiRoth:
# Before (the change)

Previously the only elements allowed in the Nav and/or Meta section of the Footer, were `<a>`. The story called to also allow `<goab-link>`.

# After (the change)

Now, anything is allowed, as the elements for both Nav and Meta are full on slots, just like GoabContainer or GoabAccordion. However `<a>` and `<goab-link>` tags are given a bit more with some CSS help to help them look proper in the Footer with nothing needed on the users' end.

In addition, this shouldn't affect anyone's current work, as all they're work should continue to look and act the exact same.

## Steps needed to test

Use the docs PR for Footer for both Angular and React. Also, I would highly recommend testing this out in alternative situations, or putting together your own footers, to test that yes all current use of the Footer should still work as intended, and that the slots also continue to work as intended.

### DIFF
--- a/apps/prs/angular/src/routes/docs/footer/footer.component.html
+++ b/apps/prs/angular/src/routes/docs/footer/footer.component.html
@@ -12,6 +12,20 @@
   </goab-app-footer-meta-section>
 </goab-app-footer>
 
+<h3>With links</h3>
+<goab-app-footer>
+  <goab-app-footer-nav-section slot="nav" heading="Services">
+    <a href="/apply">Apply online</a>
+    <goab-link trailingIcon="open">
+      <a href="https://www.alberta.ca/services">All services</a>
+    </goab-link>
+  </goab-app-footer-nav-section>
+  <goab-app-footer-meta-section slot="meta">
+    <a href="/privacy">Privacy</a>
+    <goab-link><a href="/accessibility">Accessibility</a></goab-link>
+  </goab-app-footer-meta-section>
+</goab-app-footer>
+
 <h3>With navigation</h3>
 <goab-app-footer>
   <goab-app-footer-nav-section slot="nav" heading="Services">
--- a/apps/prs/angular/src/routes/docs/footer/footer.component.ts
+++ b/apps/prs/angular/src/routes/docs/footer/footer.component.ts
@@ -3,12 +3,13 @@ import {
   GoabAppFooter,
   GoabAppFooterMetaSection,
   GoabAppFooterNavSection,
+  GoabLink,
 } from "@abgov/angular-components";
 
 @Component({
   standalone: true,
   selector: "abgov-docs-footer",
   templateUrl: "./footer.component.html",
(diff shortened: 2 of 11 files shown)


twjeffery (commented):
@ArakTaiRoth I did a full pass on this, looks pretty good, just a couple things:

- The footer nav link spacing comes out at 12px, and #2697 asks for 16px. It's the new slotted rule in FooterNavSection. We just need to make sure that goes back to 16px before this goes in.
- I think this one is breaking. The component used to give back a ul with list items and now it gives back whatever gets slotted in. So a test looking for a footer li stops matching, and screen readers won't announce them as a list anymore. Let's block the story for now.

ArakTaiRoth:
> @ArakTaiRoth I did a full pass on this, looks pretty good, just a couple things:
> 
> * The footer nav link spacing comes out at 12px, and #2697 asks for 16px. It's the new slotted rule in FooterNavSection. We just need to make sure that goes back to 16px before this goes in.
> * I think this one is breaking. The component used to give back a ul with list items and now it gives back whatever gets slotted in. So a test looking for a footer li stops matching, and screen readers won't announce them as a list anymore. Let's block the story for now.

@twjeffery I fixed the spacing issue if you can confirm? Regarding setting it to blocking, I don't think the screen reader thing is an issue. They will still be read out as links, and can still be navigated as links, I don't believe having them read out like a list is a problem. But yes, you are correct that any tests that target `<ul>` and `<li>` will now break, though I think that should be few and far between, as they never had control over those elements, they should've added things like ID's and classes to the `<a>` tags and that will still be targetable as normal.

twjeffery (approved):
@ArakTaiRoth Spacing looks good, thanks. Fair enough on the list, I think you're right that it isn't a problem. And agreed the ul and li were never really theirs to target. Let's keep it as a standard release.

PR #4220, opened 2026-08-12
fix(#2849): Removed Home and End key handling in Dropdown

### LINKED ISSUE #2849: [Bug]: using keyboard keys in the filterable dropdown
### Your email

alana.dawes@gov.ab.ca

### Web components version

TBD

### React or Angular components version

TBD

### Expected Behaviour

1. Using keys shift + end to select the text from the cursor point to the end of the text should work
2. Using keys shift + home to select the text from the cursor point to the beginning of the text should work
3. Typing part of an option (part of a word, a word, or a letter), and then using the arrow keys to find a selection, and then pressing enter on the selection should then select the option in the dropdown.

### Actual Behaviour

1. This is moving the cursor to the end of the text but not selecting the text
2. This is moving the cursor to the begining of the text but not selecting the text
3. When clicking enter, a different option than intended is selected.

### Replication Steps

For # 1:
1. Open a filterable dropdown
2. Put your cursor at the beginning of the editable dropdown selection 
3. Press shift + end

For # 2:
1. Open a filterable dropdown
2. Put your cursor at the end of the editable dropdown selection 
3. Press shift + home

For # 3:
1. In an empty state of the filterable dropdown, type in a letter, part of a word, or a word
2. Use the down arrow key to land on an option
3. Click enter

### Stackblitz URL

_No response_

### Jam.dev URL

_No response_

### Additional context

_No response_

ArakTaiRoth:
# Before (the change)

Pushing either the Home and End key in a Filterable Dropdown would move the cursor to the front or back of the text respectively. This was coded in. I don't know why this was coded in, as that's what those keys do normally anyways.

This had an effect on people using the Shift key alongside these keys. What it should do is select all the letters between the cursor and it's destination, but instead this was overwritten by the handling that was coded in.

# After (the change)

I removed the code that handled Home and End keys. Again, I don't know why it was there, I don't think it needed to be, so it's gone now and this works as expected.

## Steps needed to test

Use the Dropdown Docs PR for either Angular or React, specifically the Filterable Dropdown example

### DIFF
--- a/libs/react-components/specs/dropdown.browser.spec.tsx
+++ b/libs/react-components/specs/dropdown.browser.spec.tsx
@@ -703,6 +703,32 @@ describe("Dropdown", () => {
       })
     });
 
+    it("preserves text selection when Shift+Home and Shift+End are pressed", async () => {
+      const Component = () => (
+        <GoabDropdown name="favcolor" onChange={noop} filterable={true}>
+          <GoabDropdownItem label="Red" value="red" />
+          <GoabDropdownItem label="Blue" value="blue" />
+          <GoabDropdownItem label="Green" value="green" />
+        </GoabDropdown>
+      );
+
+      const result = render(<Component />);
+      const filter = result.getByTestId("input");
+
+      await userEvent.type(filter, "Green");
+      const input = filter.element() as HTMLInputElement;
+
+      input.setSelectionRange(2, 2);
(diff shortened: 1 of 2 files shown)


twjeffery (approved):
Looks good to me. With the cursor mid-word, Shift+Home and Shift+End both extend the selection and it stays put now. Plain Home and End still don't open the menu

PR #4228, opened 2026-08-13
chore(#2899): Updated Input example documentation

### LINKED ISSUE #2899: Update Input component slot documentation
# Info

Our examples for the Input component are still using `<div slot="leadingContent">` for leadingContent examples. All of our other documentation has started to use `ng-template`, and this should follow as well. I would also recommend updating the examples to show how to use `ng-template` as well as just a string value, as both are options.

## Acceptance Criteria:

1. Examples in Input component are updated to use `ng-template` instead of `<div slot>`
2. At least one location in the examples is updated to use a string value for `leading` or `trailingContent`.

ArakTaiRoth:
# Before (the change)

The `leadingContent` and `trailingContent` example for the Input component only uses strings

# After (the change)

The `leadingContent` and `trailingContent` example has been updated to also show how to use them as slots.

## Steps needed to test

Check the Input documentation and check the specific example "With leading or trailing content"

### DIFF
--- a/docs/src/data/configurations/input.ts
+++ b/docs/src/data/configurations/input.ts
@@ -365,7 +365,7 @@ const [sin, setSin] = useState<string>("");`,
     {
       id: "leading-trailing-content",
       name: "With leading or trailing content",
-      description: "Inputs with text content before or after the input field",
+      description: "Inputs with text or slotted content before or after the input field",
       code: {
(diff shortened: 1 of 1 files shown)


twjeffery (approved):
@ArakTaiRoth Looks good to me

Issue #3231, opened 2025-12-01
Restore focus state for v2 Input


bdfranck:
Our v2 changes to the read-only state of an Input in https://github.com/GovAlta/ui-components/pull/3082 assumed that a user doesn't interact with a read-only input. But we discovered in https://github.com/GovAlta/ui-components/pull/3226 that we use an interactive read-only Input in our Date Picker.

[MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/readonly#attribute_interactions) on readonly writes the following:

> The difference between [disabled](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/disabled) and readonly is that read-only controls can still function and are still focusable, whereas disabled controls can not receive focus and are not submitted with the form and generally do not function as controls until they are enabled.
So it appears that the Date Picker implements readonly correctly. This PR reverts our read-only changes so that the Date Picker looks interactive again.

### Acceptance Criteria:
- When an input is read-only...
  - It has focus state
  - It has a cursor pointer on hover

PR #4229, opened 2026-08-14
fix(#2981): Updated read only styling for Input and DatePicker

### LINKED ISSUE #2981: No border on hover
# Info

There should be no border when hovering over a disabled Input:

Image

## Acceptance Criteria:

1. No border on hover of a disabled input

ArakTaiRoth:
# Before (the change)

Input read-only state was incorrectly providing a border on hover and on focus. This was happening because it was assumed it had to happen to allow Date Picker to continue to have a border (it uses input read-only state to function).

# After (the change)

The CSS targeting `box-shadow` for Input's read-only stated was corrected, which wasn't written correctly first off and incorrectly didn't target hover and focus modes. The CSS for Date Picker was expanded to target focus and hover modes specifically instead.

## Steps needed to test

You can test using the docs PRs for both Date Picker and Input, there's a Read Only state for Input in there.

### DIFF
--- a/libs/web-components/src/components/date-picker/DatePicker.svelte
+++ b/libs/web-components/src/components/date-picker/DatePicker.svelte
@@ -11,7 +11,12 @@
   import { onMount, tick } from "svelte";
   import type { Spacing } from "../../common/styling";
   import { toBoolean } from "../../common/utils";
-  import { receive, dispatch, relay, watchFocusWithin } from "../../common/utils";
+  import {
+    receive,
+    dispatch,
+    relay,
+    watchFocusWithin,
+  } from "../../common/utils";
   import { isValidDimension } from "../../common/validators";
   import {
     FieldsetSetValueMsg,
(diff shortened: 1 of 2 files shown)


twjeffery (commented):
@ArakTaiRoth

The selector fix is right, read-only fields look read-only again now.

One thing I noticed, tabbing into a read-only field doesn't show the focus ring anymore. The read-only border wins over the focus styling, and since the input has no outline of its own, that ring was the only thing showing you'd landed in the field.

That's what issue #3231 was about, we put the focus state back last December for the same reason. The DatePicker borders you added here are the same thing showing up, it just applies to every read-only input and not only the ones inside a DatePicker.

We just need to make sure :not(:focus-visible) stays so the ring comes back. Dropping :not(:hover) looks right.

ArakTaiRoth:
> @ArakTaiRoth
> 
> The selector fix is right, read-only fields look read-only again now.
> 
> One thing I noticed, tabbing into a read-only field doesn't show the focus ring anymore. The read-only border wins over the focus styling, and since the input has no outline of its own, that ring was the only thing showing you'd landed in the field.
> 
> That's what issue #3231 was about, we put the focus state back last December for the same reason. The DatePicker borders you added here are the same thing showing up, it just applies to every read-only input and not only the ones inside a DatePicker.
> 
> We just need to make sure :not(:focus-visible) stays so the ring comes back. Dropping :not(:hover) looks right.

@twjeffery This has been fixed, when you tab in now to a read-only input, it gives proper focus

twjeffery (approved):
@ArakTaiRoth Looks good to me now, the focus ring is back and the read-only field still sits flat at rest.

PR #4232, opened 2026-08-17
chore(#3002): Updated error detection for FormItem

### LINKED ISSUE #3002: Fix how the formitem detects errors
Currently all input-like components dispatch an `error::change` event, but the way it is done now includes non-required complexity and can be removed from all the input-like elements.

In the current state the "input" elements watch on the component's `error` property and when it changes dispatches the event which is received by the formitem. Things can be greatly simplified since the FormItem can simply watch its own `error` property.

## Requirements
- [x] Remove all the `$:` svelte code that is watching the `error` property within RadioGroup, Input, Dropdown, Checkbox and TextArea
- [x] In the FormItem remove the addition of the `error::change` listener and add a `$:` on it's own `error` prop.

ArakTaiRoth:
This change should essentially make no difference to users of our components. This is just a simplification of error handling for our various input components and FormItem.

Previously when an error was given to an input component, it registered that error with an event that FormItem would see, and then update it's `aria-describedby` property. That could result in a FormItem updating it's `aria-describedby` property without actually providing an error.

Now, input's control their error property independently of FormItem. When FormItem's error property is updated, it updates it's `aria-describedby` property properly.

## Steps needed to test

You can manually test using PRs bug3002 for both Angular and React. There should be no difference for how the components are actually used.

### DIFF
--- a/apps/prs/angular/src/routes/bugs/bug3002/bug3002.component.html
+++ b/apps/prs/angular/src/routes/bugs/bug3002/bug3002.component.html
@@ -1,0 +1,133 @@
+<goab-text tag="h1" mt="m">Bug #3002: Fix how FormItem detects errors</goab-text>
+
+<goab-link trailingIcon="open">
+  <a
+    href="https://github.com/GovAlta/ui-components/issues/3002"
+    target="_blank"
+    rel="noreferrer"
+  >
+    View on GitHub
+  </a>
+</goab-link>
+
+<goab-text mt="m">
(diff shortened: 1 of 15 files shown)


ArakTaiRoth:
@twjeffery Just adding here, the CodeQL run did fail, it only failed to actually upload the results because Github's server didn't respond. I can't seem to re-run this, but it shouldn't actually block anything so there shouldn't be an actual problem here.

twjeffery (approved):
@ArakTaiRoth It all looks good to me

PR #4233, opened 2026-08-17
chore(#3056): Removed type property from AppHeaderMenu

### LINKED ISSUE #3056: Investigate the usage of `type` in AppHeaderMenu
# Info

There's a property `type` in our `AppHeaderMenu.svelte`. This property isn't exposed in either of the Angular or React wrappers. It's not used internally for anything (at least that I could find), so it will always be set to primary (the default). This `type` property is applied to a regular HTML button component inside the AppHeaderMenu as the `class`. But the only styling is for `button.secondary` which as I said above, is never a thing (from what I can find). So by default, it effectively is doing or adding nothing.

## Acceptance Criteria:

1. Investigate the usage of the `type` property in `AppHeaderMenu.svelte`, determine if it has a purpose
    * If it doesn't remove it and all references to it
    * If it does, don't touch it, do nothing but document your findings (how it's used) in this story.

ArakTaiRoth:
Previously we had a `type` property included in `GoabAppHeaderMenu`. This property was only available to the default web component (so essentially only Vue users). The only effect it had was if you supplied the value `secondary`, it changed the font-weight of buttons to regular from bold. It only had this effect for v1 components, v2 components overwrote that property regardless so it didn't matter.

Technically this could be considered a breaking change for any Vue users that had implemented it and are still using v1 components. But the # of those teams doing that is likely very very small if it even exists.

## Steps needed to test

There are no tests for this, there's no way to test that something doesn't exist anymore, when it couldn't be implemented anyways. The only thing to determine is if we think this is too much of a breaking change and should only be included as a major version update.

### DIFF
--- a/docs/generated/component-apis/app-header-menu.json
+++ b/docs/generated/component-apis/app-header-menu.json
@@ -81,17 +81,6 @@
           "description": "Sets a data-testid attribute for automated testing."
         },
         {
-          "name": "type",
-          "type": "\"primary\" | \"secondary\"",
-          "values": [
-            "primary",
-            "secondary"
-          ],
-          "required": false,
-          "default": "primary",
-          "description": "The menu style variant. Primary uses bold text, secondary uses regular weight."
-        },
-        {
           "name": "version",
           "type": "\"1\" | \"2\"",
           "values": [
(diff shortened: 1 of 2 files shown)


twjeffery (approved):
@ArakTaiRoth Looks good to me. The property table drops type and React and Angular are unchanged.

PR #4234, opened 2026-08-17
fix(#3103): Modified Dropdown to not resize inside containers

### LINKED ISSUE #3103: Pagination dropdown small inside Accordion
Issue reported by: andrzej.zadorozny@gov.ab.ca@gov.ab.ca
Web component version: 1.38.0
Angular/React version: Used in Vue - Tested in Angular

## Expected Behaviour

The page number dropdown should properly display the current page number inside goa-pagination.

## Actual Behaviour

When I place a table and goa-pagination inside a goab-accordion, the page number dropdown is too narrow to properly display the page number.

Works fine if it’s outside the accordion, but for some reason it gets squished inside the accordion.
Same thing happens when I put a goab-dropdown inside the accordion.
For the goa-dropdown I can at least control the width, but not as easily in the goab-pagination, since that goab-dropdown is deeply buried inside goab-pagination.

If a goab-pagination component is placed inside goa-accordion, then the page number dropdown is too narrow to properly display the page number.

Image

## Replication Steps

```html
<goab-accordion heading="Pagination inside Accordion" mb="2xl">
  <goab-table width="100%" mb="xl">
    <thead>
      <tr>
        <th>First name</th>
        <th>Last name</th>
        <th>Age</th>
      </tr>
    </thead>
    <tbody>
      <tr *ngFor="let user of pageUsers; index as i">
        <td>{{ user.firstName }}</td>
        <td>{{ user.lastName }}</td>
        <td>{{ user.age }}</td>
      </tr>
    </tbody>
  </goab-table>
  <goab-pagination
    [itemCount]="users.length"
    [perPageCount]="10"
    [pageNumber]="page"
    (onChange)="handlePageChange($event)"
  />
</goab-accordion>
```

```javascript
interface User {
  id: string;
  firstName: string;
  lastName: string;
  age: number;
}

users: User[] = [];
pageUsers: User[] = [];
page = 1;
perPage = 10;
handlePageChange(event: GoabPaginationOnChangeDetail) {
  this.page = event.page;
  const offset = (this.page - 1) * this.perPage;
  this.pageUsers = this.users.slice(offset, offset + this.perPage);
}
constructor() {
  for (let i = 0; i < 100; i++) {
    this.users.push({
      id: faker.string.uuid(),
      firstName: faker.person.firstName(),
      lastName: faker.person.lastName(),
      age: faker.number.int({ min: 18, max: 60 }),
    });
  }
  this.pageUsers = this.users.slice(0, this.perPage);
}
```

## Acceptance Criteria

1. Width should scale to largest input for Dropdown inside an Accordion
2. Make sure and verify that width is working properly outside of Accordion as well, as per above

ArakTaiRoth:
# Before (the change)

If you used a Dropdown inside of an Accordion, it would be sized differently to its counterpart outside of the Accordion. This was due to an old `@container` CSS property that was being applied to Dropdown's specifically inside of parent containers, which changed the width to use the supplied width parameter, the problem this caused was if the width parameter used something like `ch` values, because that ch would take the parent containers character size to calculate the width, which might be different.

# After (the change)

Dropdowns used inside of components (like Accordion) should now be sized the same as their equivalent outside of that component.

## Steps needed to test

Use PRs bug3103 for either Angular or React to test the difference.

### DIFF
--- a/apps/prs/angular/src/routes/bugs/bug3103/bug3103.component.html
+++ b/apps/prs/angular/src/routes/bugs/bug3103/bug3103.component.html
@@ -1,0 +1,70 @@
+<goab-text tag="h1" mt="m" mb="s">
+  Bug #3103: Pagination dropdown small inside Accordion
+</goab-text>
+<goab-text>
+  Compare the page-number dropdown outside and inside the Accordion. Both dropdowns should
+  be wide enough to display the current page number.
+</goab-text>
+<p>
+  <a
+    href="https://github.com/GovAlta/ui-components/issues/3103"
+    target="_blank"
+    rel="noopener"
+  >
+    View issue #3103 on GitHub
+  </a>
+</p>
+
+<goab-text tag="h2" mt="l" mb="s">Pagination outside Accordion</goab-text>
+<goab-table width="100%" mb="xl">
+  <thead>
+    <tr>
+      <th>First name</th>
+      <th>Last name</th>
+      <th>Age</th>
+    </tr>
+  </thead>
+  <tbody>
+    @for (user of pageUsers; track user.id) {
+      <tr>
+        <td>{{ user.firstName }}</td>
+        <td>{{ user.lastName }}</td>
+        <td>{{ user.age }}</td>
+      </tr>
+    }
+  </tbody>
+</goab-table>
+<goab-pagination
+  [pageNumber]="page"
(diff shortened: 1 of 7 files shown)


twjeffery (approved):
@ArakTaiRoth Looks good to me. The page number dropdown is the same width inside the accordion as it is outside now.

PR #4236, opened 2026-08-17
feat(#3153): Added ariaLabel property to GoabButton

### LINKED ISSUE #3153: Add aria-label support to button component
## Summary:
Add an aria-label property to the button component so product teams can provide a precise accessible name when the visible label alone isn’t specific enough (e.g., multiple “Save” buttons on one page)

## Acceptance criteria:

- Button accepts aria-label and, if used, becomes the button's accessible name.
- If both visible text and aria-label are provided, the accessible name reflects aria-label
- Docs updated
  - Added to the component playground
  - Added to the property table

_Question_

Should we also consider adding aria-labelledby?

ArakTaiRoth:
# After (the change)

There's a new property `ariaLabel` added to the `GoabButton` component. This property if supplied (it is not required) will replace the text in the button as what's read by a screen reader.

## Steps needed to test

Check the PR "feat3153" for either Angular and/or React. There are two options, one with the property supplied, one not supplied. For the supplied one, it should read out the aria label by a screen reader, for the not supplied button, the text on the button should be read out.

### DIFF
--- a/apps/prs/angular/src/routes/features/feat3153/feat3153.component.html
+++ b/apps/prs/angular/src/routes/features/feat3153/feat3153.component.html
@@ -1,0 +1,32 @@
+<goab-text tag="h1" mt="m">Feature #3153: Button aria label</goab-text>
+
+<goab-link trailingIcon="open">
+  <a
+    href="https://github.com/GovAlta/ui-components/issues/3153"
+    target="_blank"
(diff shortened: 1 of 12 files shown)


twjeffery (approved):
@ArakTaiRoth Looks good to me

PR #4237, opened 2026-08-17
fix(#3161): Dropdown filterable now focuses input

### LINKED ISSUE #3161: Filterable Dropdown not focused after caret click
# Info

Reported by Leon Hui

If you setup a Dropdown component with the `[filterable]` property set to True. When you click the caret down icon to open the Dropdown, the text input area should become focused so the user can start typing. Currently this doesn't happen.

## Acceptance Criteria:

1. When a user clicks the caret down icon on a filterable Dropdown, the text input should become focused so the user can immediately start typing
2. If you click the text input after clicking the caret, the focus disappears, it shouldn't.

ArakTaiRoth:
# Before (the change)

When you clicked the caret on a filterable Dropdown, it would just expand the Dropdown, it wouldn't focus the Input.

# After (the change)

Now when you click the caret on a filterable Dropdown, it focuses the input immediately to allow for typing right away. This is identical to how the HTML datalist component functions, as well as uses the same method our DropdownMultiselect component uses to make this work.

## Steps needed to test

Use the Dropdown docs PR to test

### DIFF
--- a/libs/react-components/specs/dropdown.browser.spec.tsx
+++ b/libs/react-components/specs/dropdown.browser.spec.tsx
@@ -618,6 +618,39 @@ describe("Dropdown", () => {
   })
 
   describe("Filterable Dropdown", () => {
+    it("focuses the input when the caret opens the menu", async () => {
+      const Component = () => (
+        <GoabDropdown name="favcolor" onChange={noop} filterable={true}>
+          <GoabDropdownItem label="Red" value="red" />
+          <GoabDropdownItem label="Blue" value="blue" />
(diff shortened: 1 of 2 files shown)


twjeffery (approved):
@ArakTaiRoth

Looks good to me. Clicking the arrow puts the cursor straight in the filter box now, same as the multiselect.

One thing I noticed while comparing the two, escape works differently between them. On the multiselect it clears the filter text and moves focus to the trigger, on this one the filter text stays. This one is the behaviour we want, so nothing to change here. I've added it to #4244 to bring the multiselect in line.

PR #4239, opened 2026-08-18
fix(#3342): Update DatePicker documentation to only allow for strings

### LINKED ISSUE #3342: Datepicker `value`, `min`, and `max` should document only `string` (to discourage `Date` useage)
## Summary
The Datepicker component docs currently show `Date` (or “date”) as an allowed type for `value`, `min`, and `max`. Update the docs to show only `string` for these properties to encourage teams to avoid `Date`, because `Date` represents a point-in-time and introduces time zone conversion risks for date-only values (e.g., off-by-one-day).

## Acceptance criteria
- Datepicker documentation prop tables show **only `string`** for:
  - `value`
  - `min`
  - `max`
- No Datepicker docs examples use `Date` for `value`, `min`, or `max`.
- (add at your discretion) Add short guidance near these props describing the recommended format (date-only string, e.g., `YYYY-MM-DD`) and why `Date` is discouraged.

## Out of scope (for this issue)
- Removing `Date` support from the component API/TypeScript types (docs-only change). If `Date` remains supported for backward compatibility, it should not be shown as a primary documented type.

ArakTaiRoth:
# Before (the change)

Date Picker documentation was a little mixed between React, Angular, and Web. It also listed Date as an accepted value in places for `value`, `min`, and `max`.

# After (the change)

Date Picker documentation for properties has been aligned between React, Angular, and Web. `string` is now the only accepted property for `value`, `min`, and `max`. In addition each of those properties specifies the arrangement of that string that's needed to work.

## Steps needed to test

Check the Date Picker documentation

### DIFF
--- a/apps/prs/angular/src/routes/docs/container/container.component.ts
+++ b/apps/prs/angular/src/routes/docs/container/container.component.ts
@@ -25,7 +25,7 @@ export class DocsContainerComponent {
   private fb = inject(FormBuilder);
 
   open = false;
-  effectiveDate = new Date();
+  effectiveDate = this.formatDate(new Date());
   form: FormGroup = this.fb.group({
     case: [""],
     reason: [""],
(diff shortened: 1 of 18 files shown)


twjeffery (commented):
@ArakTaiRoth The prop tables look right. One thing I noticed on the examples, the Locked date preview shows January 1 2024 and the Web Components tab matches it with value="2024-01-01", but React and Angular both use new Date("2024-01-01"), which renders as December 31 2023

Do the configurations file need to get updated too?

ArakTaiRoth:
> @ArakTaiRoth The prop tables look right. One thing I noticed on the examples, the Locked date preview shows January 1 2024 and the Web Components tab matches it with value="2024-01-01", but React and Angular both use new Date("2024-01-01"), which renders as December 31 2023
> 
> Do the configurations file need to get updated too?

@twjeffery I updated all documentation and all PR files to use strings instead of Date values

twjeffery (approved):

PR #4241, opened 2026-08-18
fix(#3447): Updated margins to work with none

### LINKED ISSUE #3447: Browser styling for p tags
# Info

When using `GoabText` with `tag="p"`. If no `mt` or `mb` is provided, then it defaults to browser styling for the `p` tag. If you supply `mt="none"` or `mb="none"`, it also defaults to browser styling, which is putting a margin-top and a margin-bottom.

## Acceptance Criteria:

1. See if we can explicitly say `margin-top: 0` and `margin-bottom: 0` if `mt="none"` and/or `mb="none"` is provided and have it overwrite browser styling

ArakTaiRoth:
# After (the change)

You can now use "none" for the properties `mt`, `mr`, `mb`, and `ml`, and it will actually calculate none or 0rem instead of just not outputting a margin. This allows people to actually set no margin, and have that override actual browser styles.

## Steps needed to test

Use PR bug3447 for both Angular and React

### DIFF
--- a/apps/prs/angular/src/routes/bugs/3447/bug3447.component.html
+++ b/apps/prs/angular/src/routes/bugs/3447/bug3447.component.html
@@ -1,0 +1,55 @@
+<goab-block direction="column" gap="l">
+  <goab-text tag="h1">Bug 3447: Browser styling for p tags</goab-text>
+
+  <goab-link trailingIcon="open">
+    <a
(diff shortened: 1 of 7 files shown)


twjeffery (approved):
@ArakTaiRoth Looks good to me. mt none actually gives no margin now.

PR #4248, opened 2026-08-19
fix(#4243): Removed hidden field from generated MCP and Markdown records

### LINKED ISSUE #4243: MCP sends `hidden` to consumers, who read it as "don't use this"
## What happens

24 of the 75 component records carry `hidden: true`. It means "no docs-site nav entry." Coming through the MCP it reads like an instruction, and assistants pass it on as one.

Unprompted, in a session about Scroll Panel:

> There's a separate scrollable component ("Container with overflow scrolling"), but it's flagged `hidden: true` in the catalog, so it isn't for consumption.

Fair inference from the field name, and wrong. `hidden` is set for several unrelated reasons:

| Record | Why it's hidden | Usable? |
|---|---|---|
| `tab` | Sub-component, documented on the Tabs page | Yes |
| `radio-item` | Sub-component of RadioGroup | Yes |
| `spinner` | status `stable`, category `feedback-and-alerts` | Yes |
| `chip` | status is `deprecated` | No, and `status` already says so |

So an assistant will steer people away from stable components.

## The intent is already right

From `content-generators/outputs/md-bundle.ts`:

```ts
// Include everything, like the MCP: hidden is a website-nav concept, not a
// knowledge filter. Status (incl. deprecated) rides along in frontmatter so a
// consumer sees it, same as the MCP exposes it as a filterable field.
```

Including hidden records is deliberate and correct, and `status` is already identified as the field carrying deprecation. The gap is only that this reasoning lives in a source comment, while the field travels to consumers named like a directive with nothing to interpret it by.

## Reproduce

```bash
ls data/components/*.json | xargs -I{} jq -r 'select(.hidden == true) | .id' {} | wc -l
# => 24

jq '{id, hidden, status}' data/components/spinner.json
# => hidden: true, status: "stable"
```

Then ask an MCP-connected assistant about a hidden-but-stable component and see whether it warns you off.

## Recommended fix

**Stop sending `hidden` through the MCP and the MD bundle.** One line in the generator's component and example loaders, which currently do:

```ts
hidden: data.hidden === true ? true : undefined,
```

Three reasons it can just go:

- **The server never reads it.** It is pure passthrough from the data to the consumer, so nothing depends on it.
- **It is not a knowledge concept.** The generator's own comment says so.
- **`status` already carries what a consumer needs.** `deprecated` is the actual "don't use this" signal, and it is exposed as a filterable field.

To be clear about what this does *not* change: hidden records keep shipping. The MD bundle's "include everything" decision is right and stays. Only the flag stops travelling, so nothing is hidden from consumers, they just stop being told a nav setting is an instruction.

If the field turns out to be wanted for some maintainer-facing use, the fallback is renaming it to something unambiguous (`hiddenFromDocsNav`), but dropping it is simpler and loses nothing.

## Why it matters now

Worth resolving before the MCP ships as a versioned npm package (#4149), since published versions are immutable and consuming teams inherit whatever ships.

twjeffery:
Closes #4243.

The content generator was passing a `hidden` flag through to the MCP data and the Markdown bundle. It means "no page in the docs site nav." Arriving in an AI tool under that name with no explanation, assistants read it as "do not use this component" and say so.

Seen in a real session, unprompted, while answering a question about something else:

> There's a separate scrollable component ("Container with overflow scrolling"), but it's flagged `hidden: true` in the catalog, so it isn't for consumption.

For scrollable that conclusion happens to be right. The same inference on dropdown item or tab is wrong, and that is the actual problem: one flag carrying two meanings, with no way for a reader to tell which one it got.

## What hidden was covering

24 files set the flag, and review showed they are not one kind of thing:

| Group | Count | What travels in the AI data after this PR |
|---|---|---|
| Subcomponents people write in their own markup (dropdown item, tab, radio item, the footer and side menu sections) | 15 | `subcomponent: true`, already present and kept |
| Deprecated (chip, link button, form stepper) | 3 | `status: deprecated` |
| Internal, rendered inside a parent component (calendar, focus trap, spinner, scrollable) | 4 | `internal: true`, new in the second commit |
| pages, written directly by people in their own markup | 1 | `subcomponent: true`, was missing, added |
| form, the experimental public form work | 1 | nothing for now |

The first commit stops `hidden` travelling. The second commit, prompted by review, adds the explicit `internal` flag for the four components where "not for direct use" is genuinely true, so removing the ambiguous flag does not silence the one true signal it was accidentally carrying. An assistant that finds nothing invents an answer; one that finds `internal` can point at circular progress instead of spinner.

## What this does not change

Hidden records still ship in both outputs. Only the flags change.

The website is unaffected. The docs site does not read generator output, and the nav and search index read `hidden` from the source frontmatter, which keeps it everywhere it is today. The site does not read the new `internal` field.

## Verified

Both commits regenerated and compared every record against the previous output:

- First commit: 371 records before and after, record sets identical, `hidden` gone from all, `subcomponent` retained on 15, zero other differences
- Second commit: exactly five records change, calendar, focus trap, spinner and scrollable gain `internal: true`, pages gains `subcomponent: true`, zero other differences
- The Markdown bundle index now marks the four as (internal)
- Generator and `extract-api` exit clean, docs tests pass 53 of 53 after each commit

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -150,6 +150,7 @@ relatedComponents: [button-group, icon-button]
 figmaUrl: https://www.figma.com/...
 hidden: true # optional — hide from nav (used for subcomponents, deprecated, internal)
 subcomponent: true # optional — show API on parent component page
+internal: true # optional — rendered inside other components, not for direct use (included in the generated data)
 hideTabs: [examples] # optional, hide tabs on this page (examples, usage, accessibility); Properties always shows
 ---
 ```
@@ -157,6 +158,8 @@ hideTabs: [examples] # optional, hide tabs on this page (examples, usage, access
 The MDX body is usually empty. Any content you add appears at the bottom of the "Usage guidelines" tab.
 
 **Subcomponents:** Some components are children that developers compose with a parent (e.g., Tab inside Tabs, Footer Nav Section inside Footer). These have `hidden: true` (not in nav) and `subcomponent: true` (API shown on parent page). The parent discovers subcomponents via its `relatedComponents` array. See "Add a subcomponent" below for how this works.
+
+**Internal components:** A few components exist only to be rendered inside other components (Calendar inside DatePicker, Focus Trap inside Modal and Drawer, Spinner inside Circular Progress). These have `hidden: true` and `internal: true`. Unlike `hidden`, which is a docs site concern, `internal` travels into the generated MCP and Markdown data so AI tools know these components are not for direct use.
 
 ### Guidance (`src/content/guidance/*.mdx`)
 
--- a/docs/src/content.config.ts
+++ b/docs/src/content.config.ts
@@ -34,6 +34,7 @@ const components = defineCollection({
     // Visibility
     hidden: z.boolean().optional(), // Hide from navigation and public views
     subcomponent: z.boolean().optional(), // Show API on parent component page
+    internal: z.boolean().optional(), // Rendered inside other components; not for direct use
     hideTabs: z
       .array(z.enum(["examples", "usage", "accessibility"]))
       .optional(), // Hide specific tabs on this component's page
--- a/docs/src/content/components/calendar.mdx
+++ b/docs/src/content/components/calendar.mdx
@@ -4,6 +4,7 @@ description: Visual calendar for date selection.
 status: stable
 category: utilities
 hidden: true
+internal: true
 relatedComponents:
   - date-picker
 ---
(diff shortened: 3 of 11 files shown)


ArakTaiRoth (commented):
@twjeffery Just noting we have `calendar.ts` and `spinner.ts`, that give examples on how to use them (for documentation). Should those be removed, considering they are marked as `internal`? They also have exported angular and react wrappers, should those be removed as well?

Also, in `ARCHITECTURE.md`, there's documentation (line 151 and 152) for `hidden` and `subcomponent`, should we add documentation for `internal`?

twjeffery:
@ArakTaiRoth Added internal to the architecture guide, good catch.

For the calendar and spinner configuration files and wrappers, I'd leave them for now. This PR is just about what's included in the generated data. Removing components and their wrappers is a different kind of change, so we can create a new issue for that.

twjeffery:
Created #4253 for the component removal work.

ArakTaiRoth (approved):

PR #4250, opened 2026-08-19
fix(#4249): Updated ScrollPanel to allow for a height attribute

### LINKED ISSUE #4249: Drawer and Modal scrolling is broken after ScrollPanel introduction
# Info

Currently the GoabScrollPanel needs a defined height in order to work properly. The problem with this is that it was introduced into GoabDrawer and GoabModal as the default scrolling solution, but neither of those components provide a default height, so if scrolling needs to be enabled in them (too much content), it doesn't actually function.

## Acceptance Criteria:

1. Scrolling should work when needed for GoabDrawer and GoabModal (both mobile and desktop)

ArakTaiRoth:
# Before (the change)

Previously other components that used GoabScrollPanel had to force their own specific height in order to have scrolling (Scrolling was broken in GoabDrawer and GoabModal).

# After (the change)

Now there's an internal `maxHeight` property for GoabScrollPanel, this can be used by both GoabDrawer and GoabModal in order to set a proper height if scrolling is needed (the component is larger than viewport).

I also updated the Scroll Panel from using `layout: flex` to using `layout: grid` as that made more sense for the actual layout of the component (header, body, footer, always)

## Steps needed to test

Use PR bug4249 for both Angular and React. Make sure you test for both Desktop and Mobile. Feel free to do any other testing you think is good, this hits ScrollPanel, so a lot more vast of a fix than changing GoabDrawer and GoabModal.

### DIFF
--- a/apps/prs/angular/src/routes/bugs/bug4249/bug4249.component.html
+++ b/apps/prs/angular/src/routes/bugs/bug4249/bug4249.component.html
@@ -1,0 +1,239 @@
+<goab-text tag="h1" mt="m" mb="s">
+  Bug 4249 - Drawer, Modal, and ScrollPanel sizing
+</goab-text>
+<goab-link trailingIcon="open">
+  <a
+    href="https://github.com/GovAlta/ui-components/issues/4249"
+    target="_blank"
+    rel="noopener"
+  >
+    View issue 4249 on GitHub
+  </a>
+</goab-link>
+<goab-text tag="p">
+  Verify content-sized, overflowing, and fixed-height layouts. In scrolling examples, the
+  final numbered row must be reachable while the heading and actions remain visible.
+</goab-text>
+
+<goab-divider mt="l" mb="l"></goab-divider>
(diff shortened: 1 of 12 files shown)


twjeffery (approved):
@ArakTaiRoth Approved. Push Drawer and the Work Side Menu look correct

One thing I wondered about, the Drawer only sets the max height when version is 2 but the Modal sets it either way. Was that intentional?

ArakTaiRoth:
> @ArakTaiRoth Approved. Push Drawer and the Work Side Menu look correct
> 
> One thing I wondered about, the Drawer only sets the max height when version is 2 but the Modal sets it either way. Was that intentional?

@twjeffery Yup, it was intentional. Drawer version 1 has a definite height on line 412, whereas version 2 has no height set, it's fluid. So the height only has to be set for version 2. Whereas Modal, both versions share a max-height on line 478 and neither version sets a definite height, so it has to apply to both versions.

PR #4258, opened 2026-08-24
feat(#4149): Moved the MCP server into libs/design-system-mcp

### LINKED ISSUE #4149: Design System MCP: realign with current DS release and publish as versioned npm package
## Description

The Design System [MCP](github.com/GovAlta/dcp-monorepo/apps/design-system-mcp) was previously available as a public hosted endpoint. Going forward, Design System AI tooling should be distributed as an open source, npm-packaged tool that teams install and run locally, versioned so coding agents can verify their installed MCP matches the project's Design System version.

## Expected behavior

- MCP data is updated every push to `dev`, and published every time we publish our main packages
- MCP ships as a versioned npm package, published publicly under the `@abgov` npm scope, and run locally by consuming teams (not a hosted endpoint). `
- `peerDependencies` will have to be updated every release to the latest release of `@abgov/web-components`, `@abgov/angular-components`, and `@abgov/react-components`. Also make sure those `peerDependencies` are marked as optional.
- "How to run" is documented in the MCP repo's README and on the Design System docs site which outlines compatibility guidance.

## Acceptance criteria

- [ ] Have the current MCP generation tool update a new package stored in `libs` for the MCP
- [ ] MCP published as a versioned npm package under the public `@abgov`.
    * You will need a `package.json` setup for the MCP that has `peerDependencies`, `type: module`, `bin` - listing the main executable for the MCP
- [ ] Install documentation added to the MCP repo's README.
- [ ] Install/update/compatibility documentation added to the Design System docs site.

twjeffery:
This moves the source for the design system MCP server out of dcp-monorepo and into this repo, at libs/design-system-mcp.

Why here: the generator that produces the MCP's data already lives in docs/src/scripts/content-generators. Right now the data is generated here and copied to the other repo by hand. Putting the server next to its own data source is what lets the data be generated as part of the release instead.

This is a relocation only. Nothing changes behaviour, and nothing is wired up to build or publish yet.

1. The two tools keep their descriptions, their arguments and their logic exactly as they were, and the five helper functions are untouched, so most of this should read as moved lines rather than edited ones.
2. What did change is all about this no longer being a hosted service. data-loader.ts used CommonJS globals (require.main and __dirname) that do not exist in a module build, so those four spots now use import.meta.url and process.argv[1] with the same lookup order. main.ts drops the rate limit check and the HTTP startup path, which only meant something behind a web server. The two small helpers it still needs, one for shaping error responses and one for logging them, are now local copies instead of imports from the shared hosting library, and the logging one is far smaller than what it replaces: errors to stderr, nothing else, no query history. Inside the tool blocks the only edits are that wrapper swap and the removed rate limit call. inverted-index.ts loses one log line that printed on every search.
3. There is no package.json, no build target and no release config in here yet, which means nothing in CI compiles this folder. It was checked by hand instead, see below. The build setup comes next rather than here, because how this gets built and how it gets published turn out to be the same decision: a bundled build works with the imports as they are, a plain tsc build would need every import in these three files rewritten.

The copy in dcp-monorepo stays where it is until Subbu retires it on his own schedule. The endpoint it was serving is already offline, so nothing is consuming it in the meantime.

If libs/design-system-mcp is not where you would put this, now is the cheap time to say so. Nothing is built on top of it yet, so moving it is a folder change.

Checked before opening: types are clean, and the server was run end to end against freshly generated data over a real client session. It loads 371 records, both tools respond, relationship lookups return what they should, and an unknown id gives a proper error rather than an empty answer.

### DIFF
--- a/libs/design-system-mcp/src/data-loader.ts
+++ b/libs/design-system-mcp/src/data-loader.ts
@@ -1,0 +1,476 @@
+/**
+ * GoA Design System Data Loader
+ *
+ * Loads the generator's flat collection output:
+ *   data/components/
+ *   data/examples/
+ *   data/guidance/
+ *   data/foundations/
+ *   data/get-started/
+ *
+ * The ui-components content-generators pipeline produces this shape from the
+ * docs site. Each JSON file carries an explicit `id` field (canonical, can
+ * contain "/" for nested ids); the filename is a flattened version of that id.
+ */
+
+import { existsSync } from 'fs';
+import { readFile, readdir } from 'fs/promises';
+import { join, dirname } from 'path';
+import { fileURLToPath } from 'url';
+import {
+  InvertedIndex,
+  IndexedItem,
+  createSearchableText,
+  extractTags,
+} from './inverted-index';
+
+/**
+ * Resolve the data directory.
+ *
+ * Probes a handful of locations relative to the running entry point and
+ * returns the first one that contains a `components/` subfolder. Handles
+ * both production layouts (compiled `main.js` sitting next to `data/`) and
+ * dev / script layouts (e.g. `npx tsx` on a script, where the entry point is
+ * the script itself and `data/` is one level up).
+ *
+ * Set `GOA_MCP_DATA_DIR` to override entirely.
+ */
+const moduleDir = dirname(fileURLToPath(import.meta.url));
+
+function resolveDataDir(): string {
+  if (process.env.GOA_MCP_DATA_DIR) {
+    return process.env.GOA_MCP_DATA_DIR;
+  }
+
+  const candidates: string[] = [];
+  const entry = process.argv[1];
+  if (entry) {
+    const mainDir = dirname(entry);
+    // dist/main.js shim sits next to data/.
+    candidates.push(join(mainDir, 'data'));
+    // scripts/<name>.ts is one level under data/'s sibling.
+    candidates.push(join(mainDir, '..', 'data'));
+  }
+  // src/data-loader.ts → ../data when imported directly during dev.
+  candidates.push(join(moduleDir, '..', 'data'));
+
+  for (const candidate of candidates) {
+    if (existsSync(join(candidate, 'components'))) return candidate;
+  }
+  // No probe matched; return the first guess so the caller's load attempt
+  // surfaces a clear ENOENT instead of silently loading nothing.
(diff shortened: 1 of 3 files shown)


libs/design-system-mcp/src/data-loader.ts
```diff
@@ -0,0 +1,476 @@
...
+          // Register aliases for `get` lookups.
+          if (Array.isArray(data.aliases)) {
+            for (const alias of data.aliases) {
+              if (typeof alias === 'string' && alias.length > 0) {
+                this.aliasMap.set(alias.toLowerCase(), id);
+              }
+            }
+          }
+        } catch {
+          // Skip invalid files silently
```
ArakTaiRoth: I feel like we should have some error checking here? If this goes through into our build process and publishing, we could in theory miss reading a file, publish it, and the published data is just straight up missing stuff.  Something like `throw new Error("Unable to load data file '${filePath}': {error.message}");` And put by the `catch` statement above, `catch (error) {`

libs/design-system-mcp/src/data-loader.ts
```diff
@@ -0,0 +1,476 @@
...
+        }
+      }
+    } catch {
+      // Folder doesn't exist - that's okay
```
ArakTaiRoth: Same here, but instead something like `throw new Error("Unable to load required folder `{folderPath}`: {error.message}");`

libs/design-system-mcp/src/data-loader.ts
twjeffery: Good catch. A file that couldn't be read was just being skipped and the server carried on like nothing was wrong. Now it stops and says which file.

I moved the folder read out of the try as well, otherwise the error gets picked up below and disappears again.

libs/design-system-mcp/src/data-loader.ts
twjeffery: Done, with one change. Components, examples and guidance stop the server if they're missing. Foundations, get started and product types log a warning and carry on.

The generator only creates a folder when that collection has something in it, and product types is only two records today. I didn't want that going to zero to stop the server for everyone who's installed it.

twjeffery: Changed my mind on this after talking it through with you. All six are required now, not just the core three.

ArakTaiRoth (approved):

PR #4270, opened 2026-09-04
feat(#4149): Make the MCP server build as an npm package

### LINKED ISSUE #4149: Design System MCP: realign with current DS release and publish as versioned npm package

twjeffery:
The server that moved over in 4258 could not actually be run from this repo. It had no dependencies declared, nothing compiled it, and the build knew nothing about it. This wires all of that up, so the next PR only has to add the release config.

1. libs/design-system-mcp is now a real project, set up the same way as the other libs. Building it runs the docs content generator first and copies the generated records in beside the bundled server, so dist/libs/design-system-mcp is the complete package: the server, its data, the readme and the package.json.
2. The package rebuilds when the content changes. A docs content change or a component wrapper change now marks this project affected, which is what will drive republishing once the release config exists. It also means content PRs build this package on GitHub from now on, so a content change that breaks the generated data fails the PR instead of being found after a release.
3. A check that the built package actually works. After the build, the package gets packed and installed into a temporary folder the way a team would install it, then asked real questions through the same command an MCP client runs. It has to come back with at least 300 records, the example to component connections present, and working search and get answers, or the check fails on GitHub. I also made sure it fails when it should: a corrupt record file, missing data and a too-small data set all stop it. Building it this way also compiles and type checks the server on every PR that touches it, which nothing was doing before, so a change that breaks it fails the PR instead of shipping. If that is more weight on content PRs than you would want, now is the cheap time to say so.
4. Two fixes so the installed package works outside this repo:
   - npm installs the goa-design-system-mcp command as a link to the real file, and the server looked for its data next to the link, where there is nothing. Installed that way it could not start. It now also looks next to the real file.
   - The server reported itself as version 2.0.0 no matter what. The real version is stamped into the package when a release publishes it, after the code is built, so the server now reads its own version when it starts. Run from source it reports 0.0.0, which is honest, since that build was never released.
5. Cleanup that comes with this: the three packages the generator was quietly borrowing from the root install (typescript, yaml and fastest-levenshtein) are now declared in docs like the rest of its dependencies, installing at the root now also installs this package's dependencies (the same way docs already works), and sync-to-dcp.sh is gone, since copying data to the other repo by hand is exactly what this replaces.

To try it: npx nx run design-system-mcp:post does the whole loop, build, pack, install, ask. Or run the server straight from source: cd docs && GOA_MCP_DATA_DIR=generated/mcp npx tsx ../libs/design-system-mcp/src/main.ts

Not here yet: the release config and how versions get numbered. That is the next PR, and it is small now that this one exists.


libs/design-system-mcp/scripts/smoke-test.mjs
```diff
@@ -0,0 +1,213 @@
...
+  fail(`no data at ${join(distDir, "data")} (run the build first)`);
+}
+if (!existsSync(join(distDir, "package.json"))) {
+  fail(`no package.json at ${distDir} (run the build first)`);
+}
+
+work = mkdtempSync(join(tmpdir(), "goa-mcp-smoke-"));
+
+// 1. Pack dist, exactly what a publish uploads.
+const pack = spawnSync("npm", ["pack", "--pack-destination", work, "--loglevel=error"], {
```
ArakTaiRoth: This doesn't work for me on Windows. I also had to make a small change in `loaders/get-started.ts` on line 13. Change `const slug = rel.replace(/\.mdx$/, "");` to `const slug = rel.replace(/\\/g, "/").replace(/\.mdx$/, "");`

But, even after that change, the command `npx nx run design-system-mcp:post` doesn't work on Windows for me.

libs/design-system-mcp/scripts/smoke-test.mjs
```diff
@@ -0,0 +1,213 @@
...
+  if (!(names.includes("search") && names.includes("get"))) {
+    fail(`tools/list returned [${names.join(", ")}], expected search and get`);
+  }
+  ok(`tools: ${names.join(", ")}`);
+
+  const search = await request("tools/call", {
+    name: "search",
+    arguments: { query: "button" },
+  });
+  const searchText = search.result?.content?.[0]?.text ?? "";
```
ArakTaiRoth, line 252: I can't test this as per my above comment, so this is just purely based on my comprehension of what I think the code is doing (so could totally be wrong). The `search` function being tested here actually returns a JSON string in `text`, so below where you check for `searchText.length`, that's just checking the length of the entire JSON object as a string, which doesn't really do much. Also, an empty result could still be returned, where everything below will still pass. I would suggest turning `searchText` into a JSON object to actually use it correctly, and check for `searchText.count` and that `searchText.results` actually returns the results you expect from the search above.

libs/design-system-mcp/scripts/smoke-test.mjs
twjeffery: Good catch. Your get-started change is in, and the generated data on Mac and Linux is unchanged by it. There were two more Windows layers underneath: the frontmatter parser only understood Unix line endings, so on a normal Windows checkout every content file quietly parsed as having no frontmatter at all, and the test script started npm and the server in a way that only works on Mac and Linux. Both are fixed, and the line endings one is very likely what was still failing for you after your change. I can't test on Windows myself, so would you mind giving npx nx run design-system-mcp:post another try?

libs/design-system-mcp/scripts/smoke-test.mjs
twjeffery: The response is the shape you describe, and a length check does pass on an empty one. The test now parses it and requires the button component in the results, with the same change on get, which had the same weakness. A search that comes back empty fails the test now.

ArakTaiRoth (approved):

PR #4300, opened 2026-09-21
feat(#4297): Publish the MCP server as a versioned npm package

### LINKED ISSUE #4297: Publish the design system MCP so teams can install the version that matches their components
The MCP server builds as a package now, but nothing publishes it, so teams still have no way to install it.

When this is done, a team adds the design system MCP to their AI tools in one line, and gets answers about the version of the design system they are running, not the newest one.

1. It publishes on dev and main, the same way the component libraries already do. Nothing new in the workflows.
2. The version number tells a team which design system release the answers come from.
3. A content change is enough on its own. Fixing an example or adding guidance does not need a component change to go with it, since the content changes more often than the components do.
4. A release stops if the data is missing or broken, instead of publishing a server with nothing in it. That check already runs on every pull request that touches the package.

The old hosted MCP always served the newest content to everyone, which is how teams on older components got confident answers that did not apply to them. Matching the version is the fix for that.

twjeffery:
Closes #4297.

The MCP server builds but nothing publishes it, so no team can install it yet. This adds the release for @abgov/design-system-mcp.

## What this does

The version follows @abgov/web-components: the major and minor are the design system release the answers describe, and the patch counts content updates within that release. A team on web components 2.5 pins a 2.5 version and stays on 2.5 when it updates. That's the point of publishing this as a package rather than the old hosted endpoint, which answered for the newest release no matter what a team was running.

It publishes on dev and main in the same release step as the component libraries, after them, so it matches the web components version that run published, or the one already on npm when the run published none. Nothing changes in the workflows.

It isn't on semantic-release, because semantic-release computes a version from commit messages and there's no way to make that produce the web components number. A small script (libs/design-system-mcp/scripts/release.mjs, with tests in the unit test run) works out the version, writes it into the built package, runs the same smoke test that runs on pull requests, publishes with the same npm token, and creates the git tag and GitHub release in the same format as the other packages (design-system-mcp-v2.5.1). Nothing publishes from dev until a stable version exists on npm, so latest is never a prerelease. If you'd rather this stayed on semantic-release and we gave up the version matching, now is the cheap time to say so.

The readme tells teams to pin an exact version. On npm 10, which Node 20 and 22 ship, npx never re-resolves a range, so ~2.5 would freeze at the first version installed anyway. On npm 11 a range works as you'd expect, and the readme says so.

## Two things for you to decide

1. The npm token. The package has never been published, so the first release creates a new name in the abgov scope, and the token the release step uses has to be allowed to do that. Could you check it can? If it can't, the release step fails on every push until the token is fixed or the target is removed, and the docs deploy sits behind it in both workflows. Nothing publishes wrongly either way, but this is the one thing I can't check from here.

2. Where the first version comes from. If the two MCP commits (the build from #4270 and this one) go onto main before the 3.0 release, main publishes 2.5.0 from its own content and the teams on 2.5 get a matching MCP now. If main stays as it is until 3.0, nothing publishes until then and the first version is 3.0.0. I can open a small PR to main with just those commits, or you can cherry-pick them when you're doing release work.

## Worth knowing

- A docs-only push runs nothing in the release step today. After this it runs the four libs' semantic-release checks (each finds no relevant changes and stops, about 30 seconds together), then the smoke test and the script, so the dev docs preview now depends on the npm token being valid on docs-only pushes. If you'd rather the MCP could never delay the docs preview, a separate step after the release step with continue-on-error would isolate it, and I'm happy to add that.
- Every push that could change the package publishes a new version, including dependabot bumps and component changes that leave the content the same, so the dev tag gets identical-content prereleases at about the rate the component packages already do. Main only publishes at releases.
- If a component publish fails after tagging, the MCP still publishes and describes components nobody can install until the next component release. The readme says so.
- One small change to package.json: the bin path drops its leading ./, which npm rewrote and warned about at publish. The installed command is the same.

To try it without publishing anything: build the package, then node libs/design-system-mcp/scripts/release.mjs --dry-run --branch main or --branch dev. It prints every decision and what it would publish.


twjeffery:
Closing this one in favour of #4302.

PR #4301, opened 2026-09-22
fix(#4226,#4227): fix radio accessibility and keyboard navigation

### LINKED ISSUE #4226: Radio Items are being marked as `disabled="false"` incorectly
# Info

Radio Items are being read as "Unavailable" to screen readers. This doesn't typically stop them from working, but when using JAWS it actually does prevent them from being selectable.

This is likely happening because `disabled="false"` is being sent to browsers, `disabled` is a boolean property for HTML, so passing `"false"` still marks it as a true because a value is being passed.

## Acceptance Criteria

1. Radio Buttons are no longer marked as `disabled="false"`, only `disabled="true"` when that property is passed to the component.

### LINKED ISSUE #4227: Radio group has no group semantics: announces "1 of 1", no arrow-key movement, one tab stop per option
## Current behavior

A radio group behaves as a set of unrelated single radios rather than one group. This is a component defect that ships to every product using a radio group: in any citizen-facing form or worker tool built with these components, a screen reader user is never told how many options a question has, and a keyboard user cannot move between options the way radios work everywhere else.

It shows up three separate ways, all on the same control.

**1. Screen reader announcement.** VoiceOver on a Yes/No question, in a working application built with these components:

```
"Yes, dimmed, selected, radio button"
"No, dimmed, radio button, 1 of 1"
```

"1 of 1" should be "1 of 2" and "2 of 2". A screen reader user is never told how many options the question has, so they cannot know whether they have heard them all. ("Dimmed" is a separate defect, filed as #4226.)

**2. Arrow keys do nothing.** Arrow keys do not move between options.

**3. Every option is its own tab stop.** Tab moves through a two option group in two stops.

Compared with a native radio group, and with the W3C ARIA Authoring Practices radio group pattern: one Tab stop for the whole group, arrow keys to move within it. That is the behaviour every keyboard user relies on, and it is also how a user discovers there are other options. The missing group role and position information falls under WCAG 2.2 SC 4.1.2 Name, Role, Value and SC 1.3.1 Info and Relationships.

## Expected behavior

In every product these components are used in: a radio group is one stop in the tab order, arrow keys move between its options, and each option is announced with its position: "Yes, radio button, 1 of 2".

## Reproduces on the live docs site, in all frameworks

Measured at https://design.alberta.ca/components/radio-group/ where there is no React on the page, so this is a component-level defect rather than a wrapper one. The docs site serves here only as a public surface built with the library; the same structure ships in every product:

```
roles found anywhere in the group subtree: []      <- no role="radiogroup"
goa-radio-group attributes: { id, version, name, value }   <- no ARIA

input (email)  name="contact"  tabIndex: 0  root host: goa-radio-item
input (phone)  name="contact"  tabIndex: 0  root host: goa-radio-item
input (mail)   name="contact"  tabIndex: 0  root host: goa-radio-item
distinct shadow roots among the inputs: 4
```

`tabIndex: 0` on every option is the tab-stop symptom directly: a grouped set uses a roving tabindex, where only one option is in the tab order at a time.

## Mechanism (hypothesis, for the team to confirm)

Radio grouping by shared `name` does not cross shadow root boundaries. The inputs correctly share a `name`, but each lives inside its own `goa-radio-item` shadow root, so the browser scopes grouping per root and sees several groups of one. Nothing in the light DOM supplies the grouping either, since `goa-radio-group` renders no `role="radiogroup"`.

## Suggested fix or root cause

The fix direction is the team's call, but the constraint is fixed: as long as each radio is encapsulated in its own shadow root, native grouping cannot apply and the semantics have to be declared explicitly. Two directions:

1. Implement the W3C ARIA radio group pattern at `goa-radio-group`: `role="radiogroup"` with an accessible name (from the form item's label), a roving tabindex with arrow key handling, and `aria-posinset` / `aria-setsize` on the items. This keeps the current structure.
2. Render the inputs in a single shadow root owned by the group, so native grouping and native keyboard behaviour apply and no ARIA is needed.

Option 2 is more work and removes this whole class of problem; option 1 is smaller and keeps the component boundaries as they are.

Worth checking checkbox list for the same structure. Checkboxes do not have the arrow key expectation, but the grouping announcement applies.

## Reproduction

1. Render a `GoabRadioGroup` with two `GoabRadioItem` children
2. Turn on VoiceOver and tab into the group
3. Listen for "1 of 1" on each option
4. Click the first option and press the Down arrow: focus does not move

Or in DevTools on the docs site's radio group page, confirm no element in the group subtree has `role="radiogroup"`, and that each `input` returns a different `getRootNode()`.

ArakTaiRoth:
# Before (the change)

1. Radio items were being read as "unavailable" to screen readers. This was compounded because JAWS then made these options unselectable, which made all radio items completely unusable by anyone using JAWS.
2. Radio items were their own individual tab items, when you pressed the tab key it jumped from radio item to the next radio item.
3. Each individual radio item was read as "1 of 1" to a screen reader, instead of presenting  the total group count.

# After (the change)

1. Radio items when disabled should now be completely unavailable to screen readers and keyboard controls. And when not disabled, should not be read as such by screen readers.
2. A tab now takes you to either the first radio item in the group, or the selected radio item in the group. You then control which radio item is selected with the keyboard arrows (right and down take you forward, left and up take you backwards).
3. Each individual radio item should be read as "x of y" where x is the index of the radio item, and y is the total number of non-disabled options in the group.

## Steps needed to test

1. Use Bug 4226 PR for Angular and/or React


twjeffery (commented):
@ArakTaiRoth I tested on the React and Angular previews with VoiceOver on macOS.

The keyboard part works well, including the dynamic cases: disabling and re-enabling an option, and adding, removing and reordering items.

**Two things I found:** 

**1. Disabled options read out as plain text:** A disabled option comes through with no role and no state, just its label. Where Banana starts selected and disabled, VoiceOver reads Apple and Cherry as radio buttons, and Banana as the word "Banana" on its own.

https://github.com/user-attachments/assets/0dd82779-e1d8-4ffa-83bb-1566f1628179

When every option is disabled, the group announces as a radio group and then just reads the three labels.

The test that changed in this PR shows the same thing, looking for the test id now instead of the radio role.

What's the aria-hidden doing there? 4226 only needed the change that stops sending disabled="false", so I wasn't sure if JAWS needed the extra part.

**2. The count still reads as one of one on macOS:** Every option reads out as "one of one" rather than the group count. That's with three options, none disabled.

https://github.com/user-attachments/assets/b29e696d-5402-4e57-b57c-02bc860a229b

I tried three versions side by side:

- plain elements with role=radio carrying aria-posinset and aria-setsize: "one of three"
- real radio inputs sharing one name, no aria: "one of three"
- real radio inputs each on their own name, with aria-posinset and aria-setsize: "one of one"

twjeffery (approved):

ArakTaiRoth:
Adding a comment here to document what we discussed about the two raised issues:

1. **Disabled options read out as plain text:**
    * We believe this is fine, because this is only done when reading out the full detailed screen. And technically this label would be able to be seen by sighted users, so only reading out the label is fine as it says what's on screen, but it doesn't state it as a selectable radio element.
2. **The count still reads as 1 of 1 in VoiceOver:**
    * There is a bug in [VoiceOver](https://bugs.webkit.org/show_bug.cgi?id=231324) that means it doesn't respect the `aria-posinset` and `aria-setsize` properties. And because each of our Radio Items are in their own Shadow DOM, means they do technically appear as "1 of 1". This is only solvable I believe with a complete rewrite of our Radio Group and Radio Item components, which is not in scope for this.

PR #4302, opened 2026-09-23
fix(#4297): Publish the MCP

ArakTaiRoth:
This should publish the MCP with the same version # as web-components is being published to.


twjeffery (approved):
Looks good, and a lot simpler.

